Issue #73, opened 2022-10-20
feat: implement 0x58 - PC


Eikix:
### Since

Frontier

### Group

Stack, Memory, Storage and Flow Operations

### Description

0x58 - PC

Notes
The program counter (PC) is a byte offset in the deployed [code](https://www.evm.codes/about). It indicates which instruction will be executed next. When an [ADD](https://www.evm.codes/#01) is executed, for example, the PC is incremented by 1, since the instruction is 1 byte. The [PUSH](https://www.evm.codes/#60) instructions are bigger than one byte, and so will increment the counter accordingly.

Stack output
counter: PC of this instruction in the current program.
Example
[See in playground](https://www.evm.codes/playground?unit=Wei&codeType=Mnemonic&code=%27~0x~1xJUMPDESTzesq2x~3xPUSH1%201%20l4x~6%20%7Bpreviouminstructionmtakem2%20bytes%7D%27~PCwwwlz%20%2F%2F%20Offx%5Cnw%20%20qt%20ms%20lzseq%01lmqwxz~_).

Error cases
The state changes done by the current context are [reverted](https://www.evm.codes/#FD) in those cases:

Not enough gas.
Stack overflow.

etashhh:
I can take this one

PR #109, opened 2022-10-21
✨ Introduce 0x58 - PC

etashhh:
## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Issue Number: #73 

## What is the new behavior?

- The program counter (PC) indicates which instruction will be executed next.
- The stack output is the PC of this instruction in the current program


src/kakarot/instructions/memory_operations.cairo
```diff
@@ -53,4 +55,27 @@ namespace MemoryOperations {
...
+    // @custom:since Frontier
+    // @custom:group Stack Memory Storage and Flow operations.
+    // @custom:stack_produced_elements 1
+    // @return Updated execution context.
+    func exec_pc{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr}(
+        ctx: model.ExecutionContext*
+    ) -> model.ExecutionContext* {
+        alloc_locals;
+        %{ print("0x58 - PC") %}
+
```
abdelhamidbakhta, line 70: We should return PC prior to the execution of PC opcode. With this implementation the returned PC is not correct because we already incremented it after reading PC opcode. 
You can check on https://evm.codes on the playground.
Please add also a test in test_basic.Cairo and associated JSON file.
Also please update the readme with the number of the supported opcodes.

abdelhamidbakhta (changes requested):
Left few comments. Thanks for this contribution. Once you address the comments I will approve it.

src/kakarot/instructions/memory_operations.cairo
etashhh: Thank you for the review!

Two questions:

1. I believe I understand the functionality after reviewing [https://evm.codes/](https://evm.codes/) but I'm slightly confused as to where the increment is occurring in the `exec_pc()` function. To me, the function appears to read in the current PC of the execution context -> push the value of the PC to the stack -> update the execution context with the current stack -> increment the gas used. So I would assume that if `pc` is initially set to 0, the returned `ctx` still has `pc` set to 0 as I did not explicitly increment `pc` in the code. I understand that after `exec_pc()` runs, the PC should update but where in the code does that occur?
2. How does the test JSON file convention work? I understand everything aside from the value in the `code` field.

Thank you for the guidance!

abdelhamidbakhta: 1. It occurs here: https://github.com/sayajin-labs/kakarot/blob/ce5c178a4d4b5f8efe8933281743085d2d1b3e99/src/kakarot/instructions.cairo#L76
2. The code field is the EVM bytecode to run for the test. Let me know if this is not clear

etashhh: 1. Oh I believe I get it now, so it's as simple as setting `pc` to `ctx.program_counter - 1`? 
2. Okay understood and how would we generate the bytecode for our implementation? I saw you mentioned it's possible in the evm.codes playground but I'm not 100% sure on how that would work.

abdelhamidbakhta: 1. In normal case yes. But once we have JUMP it won’t work. We will likely need to store the last value of the PC, and update it after each instruction 
2. For simple bytecode you can just create it manually.

abdelhamidbakhta (approved):
lgtm

PR #116, opened 2022-10-21
feat(bitwise): add bitwise everywhere

LucasLvy:
## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Can't use the bitwise builtin

Issue Number: N/A

## What is the new behavior?
Can use the bitwise builtin


0xMentorNotAPseudo (approved):
Well .. tried many things but not this one. There is no other way to get around this?
If it permits us to progress let's go with that and review during the journey if we find a better way to tackle that.

PR #132, opened 2022-10-22
0x5a GAS

danilowhk:
## Pull request type

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes

Issue Number: N/A

## What is the new behavior?

-0x5a GAS opcode
-
-


src/kakarot/execution_context.cairo
```diff
@@ -38,7 +38,7 @@ namespace ExecutionContext {
         let initial_pc = 0;
         let gas_used = 0;
         // TODO: Add support for gas limit
-        let gas_limit = 0;
+        let gas_limit = 21002;
```
abdelhamidbakhta, line 41: Why do you hardcode the gas limit ? It should come from the user. We need to redefine a bit the execution context to add some data from the transaction. Please put on hold this PR for the moment

abdelhamidbakhta (changes requested):
Put on hold because we need to think about how we design gas limit and transaction parameters from user

src/kakarot/execution_context.cairo
danilowhk: Ok! I added the gas limit just to test the (Gas_limit - gas_used -2) , when it was gas_limit = 0 it was giving and overflow error

danilowhk: Should i close it for now then?

abdelhamidbakhta: We can let it open. I will put a special status with a label

abdelhamidbakhta: So that you can resume the work once we design a solution. Thanks for your help

danilowhk: Ok! Will make the changes later then :)

PR #138, opened 2022-10-23
0x5b JUMPDEST

danilowhk:
## Pull request type

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes

Issue Number: N/A

## What is the new behavior?

- Jumpdest opcode implemented 
-
-

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,9 +40,9 @@ production.
 ```mermaid
 %%{init: {'theme': 'forest', 'themeVariables': { 'darkMode': 'false'}}}%%
 
-pie title Kakarot EMV opcodes support (100 / 142)
-    "Supported" : 100
-    "Not supported" : 42
+pie title Kakarot EMV opcodes support (101 / 142)
+    "Supported" : 101
+    "Not supported" : 41
 
     "Partially supported" : 0
 ```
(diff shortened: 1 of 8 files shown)


abdelhamidbakhta (changes requested):
The logic of Jumpdest is not implemented in this PR. We should parse the code upfront before execution and have a mapping of valid Jump destinations.

TotalPizza:
> The logic of Jumpdest is not implemented in this PR. We should parse the code upfront before execution and have a mapping of valid Jump destinations.

Are you sure?
I think I would have done it the same way.
In the implementations I've seen you jump to a provided offset with the JUMP opcode and if the opcode at the new pc location is JUMPDEST, then the jump succeeds.

abdelhamidbakhta:
> > The logic of Jumpdest is not implemented in this PR. We should parse the code upfront before execution and have a mapping of valid Jump destinations.
> 
> Are you sure? I think I would have done it the same way. In the EVM implementations I've seen you jump to a provided offset with the JUMP opcode and if the opcode at the new pc location is JUMPDEST, then the jump succeeds.

Well, yes the opcode itself is ok, what I meant is that the full functionality of handling jump destinations is not complete yet. But we discussed together and it is fine to handle it separately.

danilowhk:
Implemented check_jumpdest function on ExecutionContext, it will take the "pc" as parameter and return "1 or 0" depending if its a valid JUMPDEST location or not

abdelhamidbakhta (approved):
Lgtm

abdelhamidbakhta (approved):
Lgtm

PR #159, opened 2022-10-26
0x32 Origin 

mmsc2:
## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Origin opcode was not implemented

Issue Number: 104

## What is the new behavior?

This PR adds the origin opcode to the kakarot VM, It currently fetches starknet address, It has to be changed in the future to fetch EVM address from the starknet address.


danilowhk:
If you run "make format" should fix the "compilation-tests" issue

src/kakarot/instructions/environmental_information.cairo
```diff
@@ -55,6 +56,39 @@ namespace EnvironmentalInformation {
...
+        
+        %{
+        import logging
+        logging.info("0x32 - ORIGIN")
+        %}
+
+        // Get Starknet address
+        /// TODO: get EVM address from Starknet address
+        let (tx_info) = get_tx_info();
+        let origin_address = Helpers.to_uint256(tx_info.account_contract_address);
```
LucasLvy, line 82: update with the accounts registry 
```suggestion
       tx_info.account_contract_address -> evm address
```

LucasLvy (changes requested):

src/kakarot/instructions/environmental_information.cairo
mmsc2: thank you for the feedback. I have been trying to import the function "get_evm_address" from the account_registry.cairo file but i wasn't able to do so, which would be the right way to use it?.

LucasLvy: Check the balance opcode to see how to map the address

mmsc2: Already fixed, thank you!

LucasLvy (approved):
Lgtm

PR #165, opened 2022-10-26
Execute returns the whole stack instead of peek, tests uses parametrize

ClementWalter:
## Pull request type

- [ ] Bugfix
- [x] Feature
- [x] Build related changes

Currently, `kakarot.execute` only returns the peek of the stack. This is not enough to insure that the execution is correct as the whole stack needs to be tested.

## What is the new behavior?

Return the whole stack instead.
Tests have been updated accordingly and rewritten using `pytest.mark.parametrize`.
The failing tests have been marked `skipped` and should be fixed in another PR.

## Other information

We also need to return `storage` and `return_value` to battle test this EVM, see returned value from [the playground](https://www.evm.codes/playground?fork=grayGlacier)

The `coverage` module has been removed as this final coverage op is carried out in the `starknet` fixture.


LucasLvy (changes requested):
This is a good start but it still remains to check the whole stack for every test

LucasLvy (approved):

tests/test_zk_evm.py
```diff
@@ -0,0 +1,835 @@
...
+    @pytest.mark.parametrize(
+        argnames,
+        params,
+    )
+    async def test_case(self, zk_evm, code, calldata, stack, memory, return_value):
+        Uint256 = zk_evm.struct_manager.get_contract_struct("Uint256")
+        res = await zk_evm.execute(
+            code=[int(b, 16) for b in wrap(code, 2)],
+            calldata=[int(b, 16) for b in wrap(calldata, 2)],
+        ).execute(caller_address=1)
```
LucasLvy, line 830:
```suggestion
        ).call(caller_address=1)
```
Switching to call will probably improve the test speed as it won't imply any state change nor hash computing

ClementWalter: trying this rn

LucasLvy (approved):
lgtm

PR #209, opened 2022-10-29
Feat/return gas used on execute

TotalPizza:
## Pull request type

- [ ] Bugfix
- [X] Feature
- [X] Refactoring (no functional changes, no api changes)
- [ ] Build related changes

Issue Number: N/A

## What is the new behavior?

- Removing old code that is not used and was re-introduced with a pr
- `execute_at_address` returns gas used now
-


TotalPizza:
Would mess with the new flow from danilo

PR #217, opened 2022-11-01
Fix: Memory.store_n and Helpers.slice_data for border cases

thomas-quadratic:
## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Adding test cases to the opcode 0x39 led to 2 failures. The failures have been identified to be linked to corner cases in Memory.store_n and Helpers.slice_data. Such a corner case is for example when taking a slice with offset out of bound.

Issue Number: related to 98

## What is the new behavior?

Tests for the CODECOPY (0x39) opcode have been extended with more cases taking different code-paths.

Memory.store_n and Helpers.slice_data have been reimplemented to cover all cases.

Cases for slice_data:

  1. slice fully overlaps with data:  data_offset + slice_len <= data_len
  2. slice partially overlaps with data: data_offset < data_len < data_offset + slice_len
  3. slice does not overlap with data: data_len <= data_offset

For Memory.store_n, the new memory is composed of 3 parts:

  1. Head: current bytes before offset. Might be zero padded in case len < offset
  2. New elements: always inserted between offset and offset + element_len
  3. Tail: current bytes after the new elements. Might be zero padded in case total memory length is not a multiple of 32

### DIFF
--- a/src/utils/utils.cairo
+++ b/src/utils/utils.cairo
@@ -209,45 +209,21 @@ namespace Helpers {
         data_len: felt, data: felt*, data_offset: felt, slice_len: felt
     ) -> felt* {
         alloc_locals;
-        let (local sliced_data: felt*) = alloc();
-
-        let diff = data_len - data_offset;
-
-        let is_diff_greater_than_element_len: felt = is_le_felt(slice_len, diff);
-
-        if (is_diff_greater_than_element_len == 0) {
-            memcpy(dst=sliced_data, src=data + data_offset, len=diff);
-
-            let pad_n: felt = slice_len - diff;
-
-            fill(arr=sliced_data + diff, value=0, length=pad_n);
-        } else {
-            memcpy(dst=sliced_data, src=data + data_offset, len=slice_len);
-        }
-
-        return (sliced_data);
-    }
-
-    // @notice Returns a zero-padded slice of an array
-    // @param array_len Length of the array
-    // @param array Pointer to the array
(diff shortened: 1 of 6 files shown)


src/utils/utils.cairo
```diff
@@ -207,47 +207,23 @@ namespace Helpers {
...
+        let is_non_empty: felt = is_le_felt(data_offset, data_len);
+        let max_len: felt = (data_len - data_offset) * is_non_empty;
         let is_within_bound: felt = is_le_felt(slice_len, max_len);
         let len = max_len + (slice_len - max_len) * is_within_bound;
-        memcpy(dst=new_array, src=array + offset, len=len);
-        fill(arr=new_array + len, value=0, length=slice_len - len);
-        return new_array;
+
+        memcpy(dst=new_data, src=data + data_offset, len=len);
+        fill(arr=new_data + len, value=0, length=slice_len - len);
```
Eikix, line 225: So for an out of bounds offset, with slice_len = x, this function will return an array of length x, filled with zeros?

Eikix (approved):

src/utils/utils.cairo
thomas-quadratic: Yes! Offset out of bounds means all 0's. I am not sure if there are some opcodes for which the zero-padding rule is not applicable, but on a first look, it seems to be consistent through out the EVM. Arguably, we could separate slicing and padding if ever needed.

PR #233, opened 2022-11-07
ERC20 implementation and tests

danilowhk:
**Implementations**:

 - Contract is deployed when sent to address 0 on execute_at_address 
 - Contract needs to be initiated to run initial constructor code and set initial state(inmate function implemented) 
 - Dummy implementation of callvalue  Fix: 
 -mstore byte_len calculator -return data logic 
 - sub safemath implementation .(The EVM uses underflow to make some calculations faster)
 - Implement byte_len calculator that doesn’t need hints
   
  **Todos**:
  
   - configure new hash_class of account_contract on purest 
    - finalize call value implementatino(being done)
     - Fix gas calculator
-Properly pad caller starknet_address into 20 Bytes evm_address  

**Some Extra improvements:**

 - Implementation of View function(currently all functions are going through execute_at_address which is a write function)
- Optimise arrays_len calculation as with bigger contracts, which have more than 1KBytes processing becomes slow 

## Pull request type

- [X] Bugfix
- [X] Feature
- [ ] Build related changes

Issue Number: N/A

## What is the new behavior?

-
-
-


abdelhamidbakhta (approved):
lgtm

PR #237, opened 2022-11-10
Add timeit and traceit decorators for optimization purposes

ClementWalter:
## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Some `print` spread apart in tests to have resources and time consumed.

## What is the new behavior?

Automatic reporting of any call to `deploy`, `declare` or `zk_evm.<>`.

## Other information

Not sure how we want to store/display the results. Open to discussion.


tests/utils/utils.py
```diff
@@ -0,0 +1,91 @@
...
+logger = logging.getLogger("timer")
+
+
+_time_report = []
+_resources_report = []
+
+
+def timeit(fun):
+    async def timed_fun(*args, **kwargs):
+        start = time()
```
Eikix, line 20: i'm not a pythonist but could we use `time.perf_counter()` here instead of time.time()? I heard it yields better results 

```python
start = perf_counter()
# logic
stop = perf_counter()
```

Eikix (changes requested):
Could you add some comments to help understand the utils logic:)?

.gitignore
```diff
@@ -31,3 +31,4 @@ package-lock.json
 tsconfig.json
 
 lint.sarif
+kakarot.code-workspace
```
Eikix, line 34: why are we ignoring this workspace file:)? could be useful for other vscode users

tests/utils/utils.py
```diff
@@ -0,0 +1,126 @@
...
+
+
+class traceit:
+
+    _context = ""
+
+    @classmethod
+    @property
+    def prefix(cls):
+        return cls._context + ":" if cls._context != "" else ""
```
Eikix, line 43: pardon my python ignorance but what does this do?

.gitignore
ClementWalter: we could maybe, but for now I've put in mine paths like:

```
    "cairols.venvCommand": ". ~/.pyenv/versions/kakarot/bin/activate",
```

so it too specific I guess

tests/utils/utils.py
ClementWalter: you mean what a `@property` is ? high level answer is: define something that looks like an attribute but is computed

.gitignore
Eikix: alright

Eikix (approved):
lgtm

PR #243, opened 2022-11-15
Use pytest xdist to run tests in parallel

ClementWalter:
## Pull request type

- [ ] Bugfix
- [ ] Feature
- [x] Build related changes

Tests are run sequentially and take (in my personal MBP) ~30 minutes to run.

## What is the new behavior?

Use the `pytest-xdist` plugin to run tests in parallel. Default configuration in the `Makefile` uses the detected number of logical cores to run as many sub test suites as (virtual) CPUs.

## Other information

Currently, each sub task runs all the fixtures again, this could be mitigated with https://pytest-xdist.readthedocs.io/en/latest/how-to.html#making-session-scoped-fixtures-execute-only-once but I was not sure that it was worth spending the time to pickle and retrieve the starknet state while it's not the bottleneck today.


Eikix (approved):
lgtm

PR #251, opened 2022-11-16
Create a dedicated deploy function

ClementWalter:
## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Currently, deployment of an EVM bytecode is done using:
- exec_at_address with address=0
- which stores all the bytecode to memory
- then execute the bytecode
- then overwrite the bytecode with the output bytecode

Issue Number: N/A

## What is the new behavior?

The `deploy` function now makes only the required steps and remove the extra bytecode storage.


src/kakarot/library.cairo
```diff
@@ -170,33 +168,35 @@ namespace Kakarot {
...
-        // Prepare constructor data
         assert [calldata] = kakarot_address;
-        assert [calldata + 1] = bytes_len;
-        memcpy(dst=calldata + 2, src=bytes, len=bytes_len);
+        assert [calldata + 1] = 0;
 
-        // Deploy contract account
-        let (contract_address) = deploy(
+        // Deploy contract account with no bytecode
+        let (contract_address) = deploy_syscall(
```
Eikix, line 196: do we no longer emit an event with the contract_address (starknet) and evm_contract_address

src/kakarot/library.cairo
```diff
@@ -206,9 +206,54 @@ namespace Kakarot {
         // TODO: TEMPORARY SOLUTION FOR HACK-LISBON !!!
         let (_, low) = split_felt(contract_address);
         local mock_evm_address = 0xAbdE100700000000000000000000000000000000 + low;
```
Eikix, line 208: let's rather call it evm_contract_address and not "mock", as it refers to tests

src/kakarot/library.cairo
```diff
@@ -206,9 +206,54 @@ namespace Kakarot {
...
+        let ctx = ExecutionContext.compute_intrinsic_gas_cost(ctx);
+
+        // Start execution
+        let ctx = run(instructions, ctx);
+
+        // Update contract bytecode with execution result
+        IEvmContract.write_bytecode(
+            contract_address=contract_address,
+            bytecode_len=ctx.return_data_len,
+            bytecode=ctx.return_data,
```
Eikix, line 254: I'm not sure I fully understand the flow at stake:
We give as input of this function (deploy) `bytecode_len: felt, bytecode: felt*` as params
This bytecode is the bytecode of the constructor of the smart contract?

This bytecode is literally the code of the constructor for that specific smart contract, and its return value is the smart contract's code without the constructor inside it?

src/kakarot/library.cairo
ClementWalter: just forgot to emit, good catch

src/kakarot/library.cairo
ClementWalter: so basically a contract has its own bytecode.
However, when you want to deploy a contract, you send a transaction with a differente bytecode because you need to run the constructor of this contract.

Eventually the deploy transaction is special as it creates a new bytecode that somehow concat the contract bytecode and with an initial bytecode standing for the call to the constructor. This new bytecode is send as calldata to address `0x0`.

For this special transaction, the EVM runs the calldata as is, ie process the calldata as a bytecode. The output of this initial run is the bytecode of the initialized contract. This output is eventually stored as the final contract bytecode.

You will see it clearer in the tests with the 2 PRs merged and the use of `web3py`.

Eikix: Awesome! Thanks

ClementWalter:
Closed in favor of PR #245

PR #267, opened 2022-11-17
Add wrapper for web3 contract to work with kakarot

ClementWalter:
## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

When deploying a compiled solidity contract to kakarot, there is no easy way to use the known abi to test target functions.

## What is the new behavior?

A wrapper of `web3.Contract` to be used after contract deployment that allow for using the contract like `await contract.method_name(*args, **kwargs)` handling both `call` and `execute` automatically.

This wrapper is itself used in a factory fixture called `deploy_solidity_contract` to be used "as is" in test, e.g.:

```
async def test_deploy(
    self,
    deploy_solidity_contract: Callable,
):
    erc_20 = await deploy_solidity_contract("ERC20", "name", "symbol", 18, caller_address=1)
    name = await erc_20.name()
    assert name == "name"
    symbol = await erc_20.symbol()
    assert symbol == "symbol"
    decimals = await erc_20.decimals()
    assert decimals == 18
```

## Other information

There is probably a better way to integrate with  `web3` but I didn't want to spend to much time on a specific integration of one given web3 lib. This is more a first step to ease our testings with the probable increase of solidity input files.

I could spend a bit more time to update the `constructor` attribute, as well as parsing the output with the abi.


tests/integrations/test_zk_evm.py
```diff
@@ -93,19 +82,16 @@ async def test_execute(self, zk_evm: StarknetContract, params: dict, request):
 
     @pytest_asyncio.fixture(scope="module")
     async def erc_20(self, zk_evm: StarknetContract) -> dict:
-        solidity_output_path = Path("tests") / "solidity_files" / "output"
-        abi = json.load(open(solidity_output_path / "ERC20.abi"))
-        bytecode = (solidity_output_path / "ERC20.bin").read_text()
-        w3 = Web3()
-        ERC20 = w3.eth.contract(abi=abi, bytecode=bytecode)
-        deploy_bytecode = self.hex_string_to_bytes_array(
-            ERC20.constructor("name", "symbol", 18).data_in_transaction[2:]
```
matthieuauger, line 102: do you know why we were slicing the array and not anymore?

ClementWalter: I've updated the `hex_string_to_bytes_array` utils to make it handle leading `0x`, see `tests/utils/utils.py`

ClementWalter:
EDIT: couldn't resist the temptation of formatting the output as well

tests/integrations/conftest.py
```diff
@@ -0,0 +1,93 @@
...
+        ],
+    )
+
+
+@pytest_asyncio.fixture(scope="module", autouse=True)
+async def set_account_registry(
+    kakarot: StarknetContract, account_registry: StarknetContract
+):
+    await account_registry.transfer_ownership(kakarot.contract_address).execute(
+        caller_address=1
```
Eikix, line 39: we are not allowed to say `caller_address=kakarot.contract_address`
from the beginning?

ClementWalter: actually the `account_registry` is owned by `1` while here we need to make it owned by kakarot, so we update it when using kakarot and do a cleaning step afterwards (after the yield is the teardown)

see other usage of the fixture in tests/units/test_account_registry.py

tests/integrations/test_zk_evm.py
```diff
@@ -51,15 +13,15 @@ class TestZkEVM:
         "params",
         params_execute,
     )
-    async def test_execute(self, zk_evm: StarknetContract, params: dict, request):
+    async def test_execute(self, kakarot: StarknetContract, params: dict, request):
```
matthieuauger, line 16: 👍 we should rename zk_evm everywhere in the code, in the filename as well

matthieuauger (approved):
lgtm

tests/integrations/test_zk_evm.py
ClementWalter: I hesitated to rename the file actually; but I'm not sure where we want to go with the test organisation so left it like that; like I think that `test_erc20` should be a separate file dedicated to the testing of the `ERC20.sol` but this PR was about tooling and not test reorg.

tests/integrations/test_zk_evm.py
```diff
@@ -148,60 +128,66 @@ async def test_execute_at_address(
...
-                address=evm_contract_address,
-                value=0,
-                calldata=self.hex_string_to_bytes_array(params["mint"]),
-            ).execute(caller_address=2)
+        caller_addresses = list(range(4))
+        addresses = ["0x" + "0" * 39 + str(i) for i in caller_addresses]
+        with traceit.context(request.node.own_markers[0].name):
+
+            await erc_20["contract"].mint(
+                addresses[2], 0x164, caller_address=caller_addresses[1]
```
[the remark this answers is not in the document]
ClementWalter: possibly, I didn't change the test but just the tooling. Leaving this for another PR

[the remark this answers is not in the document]
matthieuauger: https://github.com/sayajin-labs/kakarot/issues/273

PR #281, opened 2022-11-21
feat/add integration test for ADDRESS opcode

### LINKED ISSUE #252: feat: Add integration test for 0x30 - ADDRESS opcode
# Feature Request

As discussed with @Eikix 

> Deploy a dummy smart contract (be it Counter Program, ERC20, or w/e) through Kakarot, store the evm_address created and then run ADDRESS to assert equality

Create integration test for `ADDRESS` opcode

**Describe Alternatives**

**Related Code**

**Additional Context**

Yes / No _(Help can be provided if you need assistance submitting a PR)_

### LINKED ISSUE #276: feat: Compile all .sol files in `tests/solidity_files` in `build-sol`
# Feature Request

The `make build-sol` command only builds targeted files hard coded in Makefile. It would be easier to add example contracts if the command would compile all the found files.

Use `find | grep`, `for` or any Makefile command/script to compile all the .sol files in the given directory.

**Describe Alternatives**

Create a .sh file to script.
Create a .py file to script.
Update the makefile each time one adds a solidity file.

**Related Code**

Makefile

ftupas:
## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Issue Number: Fix #252 and fix #276 

## What is the new behavior?

- Add integration test for ADDRESS opcode
    - Add sol for IntegrationTestContract
    - Create pytest mark
- Modify build-sol to compile all .sol files


tests/solidity_files/IntegrationTestContract.sol
```diff
@@ -0,0 +1,26 @@
...
+                            FUNCTIONS FOR OPCODES
+    //////////////////////////////////////////////////////////////*/
+    function opcodeAddress()
+        public
+        view
+        returns (address selfAddress) 
+    {
+        return (address(this));
+    }
+}
```
ftupas, line 26: Entrypoint here. I created this solidity file to the integration test for the opcodes, do you think this is a good idea? Open to discuss

Makefile
```diff
@@ -65,4 +66,6 @@ format-mac:
 	isort tests/.
 
 build-sol:
-	docker run -v $(solidity_files):/sources ethereum/solc:stable -o /sources/output --abi --bin --overwrite --opcodes /sources/ERC20.sol /sources/ERC721.sol
+	for file in ${solidity_files} ; do \
+		docker run -v ${solidity_folder}:/sources ethereum/solc:stable -o /sources/output --abi --bin --overwrite --opcodes /sources/$$file /sources/$$file; \
+	done
```
ftupas, line 71: This compiles the `.sol` files in the `tests/solidity_files` folder for #276

tests/integrations/conftest.py
```diff
@@ -85,7 +85,7 @@ async def _factory(name, *args, **kwargs):
         setattr(kakarot_contract, "contract_account", contract_account)
         deployed_contracts[name] = kakarot_contract
 
-        return kakarot_contract
+        return kakarot_contract, tx
```
ftupas, line 88: Adding `tx` here to get the resulting `evm_contract_address` and `starknet_contract_address` after contract is deployed thru kakarot

tests/solidity_files/IntegrationTestContract.sol
Eikix: I really like this, open for other views but this helps a lot for:
- separation of concerns -> you're not testing the addresses of erc20 or erc721
- this is very visual

Eikix (approved):
Really nice PR !!!

tests/integrations/conftest.py
ClementWalter: this was my first implementation but really disliked eventually this returned tuple, that's why I added an attribute "contract_account".
You should be able to retrieve it using `kakarot_contract.contract_account.deploy_call_info`

ClementWalter (changes requested):
please don't update the fixture as explained in the corresponding comment

tests/integrations/test_zk_evm.py
```diff
@@ -107,3 +107,20 @@ async def test_erc20(
...
+        state = kakarot.state.copy()
+        caller_addresses = list(range(4))
+        integration_contract, tx = await deploy_solidity_contract(
+            "IntegrationTestContract", caller_address=caller_addresses[1]
+        )
+
+        with traceit.context(request.node.own_markers[0].name):
+
+            evm_contract_address = await integration_contract.opcodeAddress()
+            assert tx.result.evm_contract_address == int(evm_contract_address, 0)
```
ClementWalter, line 124: why do we have to cast `evm_contract_address`? also surprised by the `, 0)`

tests/integrations/conftest.py
ftupas: Ohh thanks @ClementWalter ! I missed this nice attribute from the wrapper you made, awesome 🙂 I will use that instead 👌

tests/integrations/test_zk_evm.py
ftupas: So calling `opcodeAddress` gives us `evm_contract_address="0xAbDe100763b3fb83ae8C5aFE6c635eEBA23E276B"`. I had to cast it to `int(foo, 0)` as it's a string with prefix 0x in order to assert equality with the resulting evm contract address from `deploy_call_info`.

matthieuauger: Can't you do the opposite and cast tx.result.evm_contract_address in hex() ?

ftupas: @Eikix Ah yes that's true, base 0 interprets it exactly as a code literal but you're right it's better to be explicit! Indeed it's good to change this to `int(foo, 16)` instead 🤙

@matthieuauger Good idea, also works with hex but had to format it to uppercase like so `assert f"0x{..result.evm_contract_address:X} == evm_contract_address`. Perhaps using `int` is more readable.. should we use that instead?

Eikix: i guess it'd make sense to compare hexes as hexes are used to represent addresses, but I wonder what equality operator we'd be using -> we'd be asserting over string equality instead of integer equality?

ftupas: Yeah I agree, perhaps casting them both to hex and assert equality like this `hex(..result.evm_contract_address) == hex(int(evm_contract_address), 16)`

Eikix: what do you think is the problem with comparing both hexes?
to be fair, everything in the context of starknet is an int or at least a felt:)!
comparing both data as int lgtm!!!
i'll approve your PR with int() comparison, double casting (int and then hex to not break) feels like overdoing it for me

Eikix (dismissed):
nice pr ! thanks 🔥

tests/integrations/test_zk_evm.py
ClementWalter: int is easier in our context because python can handle BigInt natively

ClementWalter (dismissed):

Eikix:
resolve conflicts and good-to-go:)
as @ClementWalter said, it's a bit annoying to have conflicts on coverage files, we'll prioritise fixing that

Eikix (approved):
✅ 🏆

Issue #268, opened 2022-11-18
dev: Benchmark the difference in performance between generating the entire list of instructions for every opcode or running an (extensive) if/else to carefully choose instructions


Eikix:

abdelhamidbakhta:
Currently we build dynamically an array to store the function offset of each opcode. It is quite elegant because it enables us to add them individually, and have a more readable code. There is a section that handles the dynamic invocation of each opcode related function. However, this code might be greedy in terms of computation, and we should investigate other options to handle the mapping of opcodes and there function.

The possible options are:
1. keep as is, dynamic invocations: elegant, flexible, but slow
2. mega giant if / return pattern: somewhat non readable giant function, but maximal efficiency
3. use `dw` to build opcodes array and use this to do the dynamic `invoke`: something in between, but my intuition is that it terms of performance it will be closer to 1. than to 2.

I would not even try 3. and go directly with 2. and estimate the benefit in terms of computation steps / builtins

Eikix:
Thanks a lot Abdel for this precise and comprehensive description 🔥

abdelhamidbakhta:
Ok, @spapinistarkware the GOAT has spoken:

```text
I assume you are discussing how to do a "switch case" on opcode? Not really up to date with details, but what about a jump table, like:

tempvar offset =2+2*opcode
jmp rel offset
jmp op0
jmp op1
```

Let's try this approach!!!

Eikix:
How does the jmp know what op0 mean, op1 mean etc.? As in, where should one write the location of op0 ?

spapinistarkware:
You can define them as labels in the dame function, or as new functions, if you carefully manage the arguments as to not push again

ClementWalter:
I'm on it

PR #282, opened 2022-11-21
Use a jump tables for opcodes

ClementWalter:
## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

We initialize an `instructions: felt*` array with all the opcodes each time we need to run kakarot.

solve #268 

## What is the new behavior?

I've added a (big) jump table in the `decode_and_execute` function. (this is though not really bigger than the previous `generate_instructions` — compared to the previous solution we need a line per index while before it was possible to leave memory holes.

## Other information

Observed a ~15% perf improvement in terms of n_steps


tests/integrations/test_zk_evm.py
```diff
@@ -61,6 +61,24 @@ async def test_deploy(
         decimals = await erc_20.decimals()
         assert decimals == 18
 
+    async def test_deploy_erc721(
+        self,
+        deploy_solidity_contract: Callable,
+    ):
+        erc_721 = await deploy_solidity_contract(
+            "ERC721", "name", "symbol", caller_address=1
```
Eikix, line 69: are these arguments named? 
we could have a more visual setup:
```python
erc_721 = await deploy_solidity_contract(
            contract="ERC721", name="KakarotNft", symbol="KRTNFT", caller_address=1
)
```

Eikix: same for erc20

src/kakarot/instructions.cairo
```diff
@@ -34,80 +34,579 @@ from kakarot.instructions.sha3 import Sha3
...
-    // @param instructions The instruction set.
+    // @dev The function uses an internal jump table to execute the corresponding opcode
     // @param ctx The pointer to the execution context.
     // @return The pointer to the updated execution context.
-
     func decode_and_execute{
         syscall_ptr: felt*,
         pedersen_ptr: HashBuiltin*,
-        range_check_ptr,
+        range_check_ptr: felt,
```
Eikix, line 43: I've never seen this, what does this do?:)

src/kakarot/instructions.cairo
```diff
@@ -34,80 +34,579 @@ from kakarot.instructions.sha3 import Sha3
...
-        let is_pc_le_code_len = is_le_felt(ctx.call_context.bytecode_len, pc);
+        let is_pc_ge_code_len = is_le_felt(ctx.call_context.bytecode_len, pc);
 
-        if (is_pc_le_code_len == 1) {
+        if (is_pc_ge_code_len == TRUE) {
             assert opcode = 0;
         } else {
             assert opcode = [ctx.call_context.bytecode + pc];
         }
+        tempvar offset = 1 + 4 * opcode;
```
Eikix, line 59: could you add a comment to help readability 🪄 ?

src/kakarot/instructions.cairo
```diff
@@ -34,80 +34,579 @@ from kakarot.instructions.sha3 import Sha3
...
+        // Prepare arguments
+        [ap] = syscall_ptr, ap++;
+        [ap] = pedersen_ptr, ap++;
+        [ap] = range_check_ptr, ap++;
+        [ap] = bitwise_ptr, ap++;
+        [ap] = ctx, ap++;
+
+        // call opcode
+        jmp rel offset;
+        call exec_stop;
```
Eikix: can you add a comment to state this is opcode number 0? 
when is this called? when offset == 0?

src/kakarot/instructions.cairo
```diff
@@ -34,80 +34,579 @@ from kakarot.instructions.sha3 import Sha3
...
-        }
-
         local opcode;
 
-        let is_pc_le_code_len = is_le_felt(ctx.call_context.bytecode_len, pc);
+        let is_pc_ge_code_len = is_le_felt(ctx.call_context.bytecode_len, pc);
 
-        if (is_pc_le_code_len == 1) {
+        if (is_pc_ge_code_len == TRUE) {
             assert opcode = 0;
```
Eikix, line 55: what happens when opcode <- 0? 
offset <- 1
then jump points to exec_stop ?

Eikix (changes requested):
very cool PR!!
~20% improvement is giga based

tests/integrations/test_zk_evm.py
ClementWalter: yes they could be kwargs as well, they are directly given to the `web3.contract().constructor()`.
I can update this, but tbh I'd like to move ERC20 and ERC721 tests into their own files integrating all the tests from solmate repo, so I'm note sure it's worth doing anything here.

But it makes me realized that these utils aren't well documented, so in any case I'm adding some docstring, and consequently why not these args as kwargs

src/kakarot/instructions.cairo
ClementWalter: it's the same, I think that I changed this for an easier copy/pasting, which led to the useless cast at line 590 😇

src/kakarot/instructions.cairo
```diff
@@ -34,80 +34,579 @@ from kakarot.instructions.sha3 import Sha3
...
-
-        // Revert if pc < 0
-        with_attr error_message("Kakarot: InvalidCodeOffset") {
-            assert_nn(pc);
-        }
-
         local opcode;
 
-        let is_pc_le_code_len = is_le_felt(ctx.call_context.bytecode_len, pc);
+        let is_pc_ge_code_len = is_le_felt(ctx.call_context.bytecode_len, pc);
```
spapinistarkware, line 52: is_felt_le is not that cheap. Use the 128bit version if possible.

ClementWalter: good to know, thx

Eikix (approved):
lgtm

PR #285, opened 2022-11-21
Use ret instead of jmp end in instructions

ClementWalter:
## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

I used a combination of `jmp end; end:ret` because of a index shift I was not able to solve (🤯) but thanks to Shara actually got it to work, so updating the code.

This also brings a slight improvements as this makes less steps.

## What is the new behavior?

Call `ret` right after `call opcode`.


Eikix (commented):
Nice PR!

As I understand it, calling ret is equivalent to casting different parameters:
```
let syscall_ptr = cast([ap - 5], felt*);
        let pedersen_ptr = cast([ap - 4], HashBuiltin*);
        let range_check_ptr = [ap - 3];
        let bitwise_ptr = cast([ap - 2], BitwiseBuiltin*);
        let ctx = cast([ap - 1], model.ExecutionContext*);

        return ctx;
```

ClementWalter:
> if you can resolve conflicts, i'll merge :))

your call now

Eikix (approved):

PR #300, opened 2022-11-25
Update check-resources to display all historical data and not only main

ClementWalter:
## Pull request type

- [ ] Bugfix
- [ ] Feature
- [x] Build related changes

The `check_resources.py` script checks local test reports against the latest one from `main`.
The output of the dataframe in the log is truncated.

## What is the new behavior?

The output is not truncated anymore.
Furthermore, the command display stats not only for main but for all the branches with artifacts.
The command can also be used locally to fetch perf, even though no local artifacts exist.

## Other information

There is still an open question regarding this hard limit on every resources used.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -899,6 +899,17 @@ python-versions = "!=3.0.*,!=3.1.*,!=3.2.*,>=2.7"
 six = ">=1.5"
 
 [[package]]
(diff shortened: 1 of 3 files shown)


Eikix (approved):
lgtm

PR #303, opened 2022-11-26
Profiling data

spapinistarkware:
## Pull request type

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes

No profiling data

## What is the new behavior?

Profiling data for each test in pprof format when running tests with the `--trace-run` flag.

## Other information

Generating profiling data is **slow** and consequently disabled by default! When you want to use the tool, just run the target test with `pytest <target test> --trace-run`

### DIFF
--- a/tests/utils/utils.py
+++ b/tests/utils/utils.py
@@ -250,6 +250,7 @@ def dump_reports(path: Union[str, Path]):
     times.to_csv(p / "times.csv", index=False)
     traces.to_csv(p / "resources.csv", index=False)
     for label, runner in _profile_data.items():
+        logger.info(f"Dumping TracerData for runner {label}")
         runner.relocate()
         tracer_data = TracerData(
             program=runner.program,
--- a/tests/utils/utils.py
+++ b/tests/utils/utils.py
@@ -1,35 +1,3 @@
-from starkware.cairo.lang.tracer.profile import profile_from_tracer_data
-from starkware.cairo.lang.tracer.tracer_data import TracerData
-from starkware.starknet.business_logic.execution.execute_entry_point import (
-    ExecuteEntryPoint,
-)
-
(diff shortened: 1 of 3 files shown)


abdelhamidbakhta (approved):
Lgtm

abdelhamidbakhta (approved):
Lgtm

tests/utils/utils.py
```diff
@@ -1,3 +1,35 @@
...
+            program_base=1,
+            debug_info=runner.get_relocated_debug_info(),
+        )
+
+        profile = profile_from_tracer_data(tracer_data)
+        _profile_datas[traceit._context] = profile
+    return runner, syscall_handler
+
+
+ExecuteEntryPoint._run = override_run
```
ClementWalter, line 30: can you move this after the import statements? and don't forget to run `make format` to make sure the ordering is the right one

tests/utils/utils.py
```diff
@@ -1,3 +1,35 @@
...
+from starkware.cairo.lang.tracer.tracer_data import TracerData
+from starkware.starknet.business_logic.execution.execute_entry_point import (
+    ExecuteEntryPoint,
+)
+
+# Override _run() entrypoint for hooking and profiling cairo runner.
+old_run = ExecuteEntryPoint._run
+
+
+def override_run(*args, **kwargs):
```
ClementWalter, line 11: could be better to make it a decorator, like : 

```
def trace_run(run):
    def override_run(*args, **kwargs):
    # use the run param here and not `old_run`
          return

    return override_run

```

and in `conftest.py` we use a fixture with `autouse=True` to enable the behavior during the test suite

```
@pytest.fixutre(autouse=True)
def overide_run():
    ExecuteEntryPoint._run = trace_run(ExecuteEntryPoint._run)
    return
```

Eikix (approved):
lgtm

PR #304, opened 2022-11-27
fix: fix mstore not ceiling to next byte when offset

matthieuauger:
Related to: https://github.com/orgs/sayajin-labs/projects/3/views/6?query=is%3Aopen+sort%3Aupdated-desc

Goal of this ticket is to remove the [SKIPPED test](https://github.com/sayajin-labs/kakarot/pull/304/files#diff-33071fb3d4629eb5c6c61cecfc36090d09f4f9b5c408e92088f1df2488cc97a1L994) about mstore.

Currently when there is an offset we increase the memory with the exact necessary size but normal behaviour should be to expand memory to next byte (see for example [evmjs implementation](https://github.com/ethereumjs/ethereumjs-monorepo/blob/985f2aa43c005573382b98423fcfd3bd368b2c8e/packages/evm/src/memory.ts#L30)

Current PR add a test and remove the skipping of the current one in mstore

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

### DIFF
--- a/src/kakarot/memory.cairo
+++ b/src/kakarot/memory.cairo
@@ -72,7 +72,8 @@ namespace Memory {
         let word_dict = self.word_dict;
 
         // Compute new bytes_len.
-        let new_min_bytes_len = offset + 32;
+        let new_min_bytes_len = Helpers.ceil_bytes_len_to_next_32_bytes_word(offset + 32);
+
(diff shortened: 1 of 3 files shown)


spapinistarkware (approved):

spapinistarkware:
Lgtm

PR #310, opened 2022-11-28
perf: change boolean comparisons to `!=`

### LINKED ISSUE #307: perf: [optimization] Use `a !=0` instead of `a == 1` for boolean comparisons
## Description

In Cairo, it is more efficient to do  `a !=0` instead of `a == 1` for boolean comparisons, there are some places in the code where we use `== 1 (see [example](https://github.com/sayajin-labs/kakarot/blob/a8b99f306ac0bbeb181c96189caa16ea2ffca4bf/src/utils/utils.cairo#L58))

## Acceptance criteria

- All boolean comparisons use `!= 0` instead of `== 1`
- Test pass
- Performance improvement is measured and reported

ftupas:
## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: Fix #307 

## What is the new behavior?

- Use `!=` instead of `==` for boolean comparisons
- Use `TRUE` or `FALSE` instead of `0` or `1` for readability

### DIFF
--- a/src/kakarot/instructions.cairo
+++ b/src/kakarot/instructions.cairo
@@ -9,7 +9,7 @@ from starkware.cairo.common.invoke import invoke
 from starkware.cairo.common.math import assert_nn
 from starkware.cairo.common.math_cmp import is_le
 from starkware.cairo.common.memcpy import memcpy
-from starkware.cairo.common.bool import TRUE
+from starkware.cairo.common.bool import FALSE
 from starkware.cairo.common.registers import get_ap
 from starkware.cairo.common.registers import get_label_location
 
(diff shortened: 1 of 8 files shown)


ClementWalter (approved):

src/kakarot/instructions.cairo
```diff
@@ -51,7 +51,7 @@ namespace EVMInstructions {
 
         let is_pc_ge_code_len = is_le(ctx.call_context.bytecode_len, pc);
 
-        if (is_pc_ge_code_len == TRUE) {
+        if (is_pc_ge_code_len != FALSE) {
```
ClementWalter, line 54: according to @spapinistarkware you just need to use `TRUE` as much as possible

ClementWalter (changes requested):

src/kakarot/instructions.cairo
spapinistarkware: No no, the other way around. Avoid TRUE(1), use only FALSE(1)

Issue #239, opened 2022-11-13
Create a bytepacking library for Kakarot: optimizing storage of bytes array


Eikix:
Currently, Kakarot is developed using a naive implementation.

As such, bytes (8-bit words in memory) are stored in a single felt. This means that a felt (up to 251-bit of information) is used to store only 8 bits.

The aim of this issue is to pave the way for byte packing.

Best outcome: 31 bytes per felt -> maximum efficiency
Mild outcome: 16 bytes per felt -> ok efficiency
Bad outcome: 1 byte per felt -> current efficiency

Eikix:
@danilowhk is currently assigned

danilowhk:
I can do this one

Eikix:
I believe @ClementWalter and @danilowhk will tackle this issue in the coming days:)

danilowhk:
Yes haha :)

ClementWalter:
Closing this as bytes16 packing was merged 3 days ago. Should we want to investigate further with, e.g. bytes31 packing, we'll create another issue, but this one is AKAIF resolved.

PR #312, opened 2022-11-28
Use internal bytes16 storage, no external API change

ClementWalter:
## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

The Account contract stores the bytecode of the contract as an array of felt, one byte per felt.

Issue Number: #239 

## What is the new behavior?

I have used bytes16 internal felt packing.

## Other information

Is 16 better than 31? I've used this because of the `unsigned_div_rem` but maybe better to use bytes31 and Uint256 div.
May also be better to accept the packed bytecode in the constructor to avoid the cost of doing it in the contract.
Maybe to return packed felt as well and update the other functions accordingly.
This was a fast (and furious 🏁) quick look at the question to have a first estimate of potential gains.

### DIFF
--- a/src/kakarot/accounts/contract/library.cairo
+++ b/src/kakarot/accounts/contract/library.cairo
@@ -34,6 +34,8 @@ func storage_(key: Uint256) -> (value: Uint256) {
 func is_initialized_() -> (res: felt) {
 }
 
+// Define the number of bytes per felt. Above 16, the following code won't work as it uses unsigned_div_rem
+// which is bounded by RC_BOUND = 2 ** 128 ~ uint128 ~ bytes16
 const byte_size = 16;
 
 namespace ContractAccount {
@@ -158,6 +160,9 @@ namespace ContractAccount {
 }
 
 namespace internal {
+    // Use a precomputed 2 ** n array to save on resources usage.
+    // Array starts with a 0 to be shifted and have pow[i] = bit shift for byte i with
+    // i as a counter, ie i \in (0, byte_size]
     pow_:
     dw 0;
     dw 1;
@@ -178,7 +183,7 @@ namespace internal {
     dw 2 ** 120;
 
     // @notice Store the bytecode of the contract.
-    // @param index: The index in the bytecode_stored.
+    // @param index: The current free index in the bytecode_ storage.
     // @param bytecode_len: The length of the bytecode.
     // @param bytecode: The bytecode of the contract.
     func write_bytecode{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr}(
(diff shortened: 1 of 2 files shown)


spapinistarkware:
Re the discussion (maybe for future PRs), I think we want 31 and get and return in packed representation.
Wdyt?

spapinistarkware:
Another note: currently the memory is also packed in 128b chunks, and I think we are deserializing to bytes and serializing back to 128b.
If the chunks are aligned, maybe we could avoid this altogether.

And another note: the packing can be a bit more optimized low level, I think, by doing aligned chunks at once (kind of loop unrolling). Checkout the memory implementation.

ClementWalter:
> Re the discussion (maybe for future PRs), I think we want 31 and get in packed representation. When loading in preparation for execution, it makes sense to keep it in bytes though Wdyt?

indeed, we could store and return the packed representation, and update the call context usage to handle a packed representation. Seems reasonable

ClementWalter:
> Another note: currently the memory is also packed in 128b chunks, and I think we are deserializing to bytes and serializing back to 128b. If the chunks are aligned, maybe we could avoid this altogether.
> 
> And another note: the packing can be a bit more optimized low level, I think, by doing aligned chunks at once (kind of loop unrolling). Checkout the memory implementation.

this would help the CODECOPY opcode only right?

you mean that writing 16 lines like `packed_felt = bytecode[i] * 2**(120 - 8*i) + ...` is more efficient?

ClementWalter:
codecov  fails because cairo-coverage doesn't recognize that the array:

```
    pow_:
    dw 0;
    dw 1;
    dw 2 ** 8;
    dw 2 ** 16;
    dw 2 ** 24;
    dw 2 ** 32;
    dw 2 ** 40;
    dw 2 ** 48;
    dw 2 ** 56;
    dw 2 ** 64;
    dw 2 ** 72;
    dw 2 ** 80;
    dw 2 ** 88;
    dw 2 ** 96;
    dw 2 ** 104;
    dw 2 ** 112;
    dw 2 ** 120;
```

is used

Eikix (approved):

LucasLvy:
> codecov fails because cairo-coverage doesn't recognize that the array:
> 
> ```
>     pow_:
>     dw 0;
>     dw 1;
>     dw 2 ** 8;
>     dw 2 ** 16;
>     dw 2 ** 24;
>     dw 2 ** 32;
>     dw 2 ** 40;
>     dw 2 ** 48;
>     dw 2 ** 56;
>     dw 2 ** 64;
>     dw 2 ** 72;
>     dw 2 ** 80;
>     dw 2 ** 88;
>     dw 2 ** 96;
>     dw 2 ** 104;
>     dw 2 ** 112;
>     dw 2 ** 120;
> ```
> 
> is used

@spapinistarkware Are `dw` transformed at compile time so they are not instructions at runtime ?

PR #315, opened 2022-11-29
dev: change formatting check in ci

### LINKED ISSUE #296: dev: change formatting check in CI from blocking ❌ a PR to running format command and committing the diff to the PR branch

ftupas:
## Pull request type

- [ ] Bugfix
- [ ] Feature
- [x] Build related changes

Issue Number: Fix #296 

## What is the new behavior?

- change formatting check in CI from blocking ❌ a PR
- Format all cairo files and commit diff to the pr branch

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -5,8 +5,13 @@ on: [push, pull_request]
 jobs:
   compilation-tests:
     runs-on: ubuntu-latest
+    env:
+      CI_COMMIT_MESSAGE: CI Formatting Auto Commit
+      CI_COMMIT_AUTHOR: ${{ github.event.repository.name }} CI
     steps:
       - uses: actions/checkout@v3
+
+      # Python setup
       - name: Set up Python 3.9
         uses: actions/setup-python@v4
         with:
@@ -33,6 +38,20 @@ jobs:
       - name: Install dependencies
         if: steps.cached-poetry-dependencies.outputs.cache-hit != 'true'
         run: make setup
+
+      # Build artifacts
+      - name: Event name
+        run: echo "github.event_name=${{ github.event_name }}"
+      - name: Format all the cairo files
+        run: make format
+      - name: Commit changes
+        # Run on push events and commit only when there are changes, otherwise it continues
+        if: github.event_name == 'push'
+        run: |
+          git config --global user.name "${{ env.CI_COMMIT_AUTHOR }}"
+          git config --global user.email "sayajin-labs@users.noreply.github.com"
+          git add -A
+          git diff-index --quiet HEAD || (git commit -m "${{ env.CI_COMMIT_MESSAGE }}" && git push)
       - name: Compile all the cairo files
         run: |
           make format-check
--- a/src/kakarot/library.cairo
+++ b/src/kakarot/library.cairo
@@ -48,8 +48,7 @@ namespace Kakarot {
     ) {
         Ownable.initializer(owner);
         native_token_address.write(native_token_address_);
-        evm_contract_class_hash.write(
-        evm_contract_class_hash_);
+        evm_contract_class_hash.write(evm_contract_class_hash_);
         return ();
     }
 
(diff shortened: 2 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -5,38 +5,76 @@ on: [push, pull_request]
...
+      - name: Set environment variable "is-auto-commit"
+        if: env.commit-message == env.CI_COMMIT_MESSAGE && env.commit-author == env.CI_COMMIT_AUTHOR
+        run: echo "is-auto-commit=true" >> $GITHUB_ENV
+      - name: Display environment variable "is-auto-commit"
+        run: echo "is-auto-commit=${{ env.is-auto-commit }}"
+
+      # Python setup
       - name: Set up Python 3.9
         uses: actions/setup-python@v4
+        if: env.is-auto-commit == false
```
ClementWalter, line 36: why can't we use the cache when the commit is the auto-commit?

If I understand correctly, when the `pull_request` event trigger this CI, then `is-auto-commit` will be false, which will in turn create another commit (in "Format all the cairo files and commit" step"). This new commit will also trigger the workflow, but will be detected as auto-commit and consequently skip all the tasks to go directly to the tests. Still it needs to have a working venv to perform those tests, right?

ftupas: Ah what I had in mind was to skip them when the auto-commit is pushed i.e. `is-auto-commit` is true so we don't need to test and build again, but I think it would be better to still do it and indeed we need a working venv to perform those steps ✅ let me play modify that

ClementWalter: auto format should work but I think that it's always a safe check to run test on the final code version

.github/workflows/test.yml
```diff
@@ -33,6 +38,20 @@ jobs:
...
         run: make setup
+
+      # Build artifacts
+      - name: Event name
+        run: echo "github.event_name=${{ github.event_name }}"
+      - name: Format all the cairo files
+        run: make format
+      - name: Commit changes
+        # Don't run on pull request events. Commit only when there are changes, otherwise it continues
+        if: github.event_name != 'pull_request'
```
ClementWalter, line 49: can't we add the commit to the PR instead of adding it to main?

.github/workflows/test.yml
ftupas: Finally I took out the `is-auto-commit` check and just made it simpler. So on `push`, it runs `make format` and commit the diff, if any. In the pr, this commit is then added and could be reviewed as well

.github/workflows/test.yml
ftupas: So on `push` event, this step adds the [commit](https://github.com/sayajin-labs/kakarot/pull/315/commits/cebd1b1e31253a4b40d5bf8b7c6428af47d69dce) to the pr

ClementWalter: ok got it sorry, not an expert in GH action, I had "push to main" in mind

ftupas: Ah no worries, my wording was also a bit vague 😅 Upon merging, it does a `push` to main but there won't be files modified from `make format` (since it's been committed already). Then it `git commit && git push` would be skipped and just proceeds to the compile step

ClementWalter (dismissed):

ftupas:
Thanks for the approval @ClementWalter , I also wanted to test the happy path i.e. no formatting issues, to make sure there's no auto-commit added

ClementWalter (dismissed):

.github/workflows/test.yml
Eikix, line 49: I think this doesn't help the reader know the other events.

If there are like 5-6 `github.event_name`, then I guess this is fine you wouldn't write `if: github.event_name == 'push' || ....` -> then I guess we can probably list them in comment

If there's only 1 or 2 we're targeting, could we prefer the `==` notation?

ftupas: That's true we only have `push` and `pull_request` event triggers we're targeting and it will be more explicit to say we check and commit the formatting every `push` event 👍  I've updated this to `== 'push'` notation, thanks @Eikix

Eikix (approved):
🔥🔥🔥🔥🔥

Eikix:
Congratulations 🔥

PR #319, opened 2022-11-29
Stack 128bits - Dict implementation

danilowhk:
Implemented a 128bits Dict for Stack based on the 128bits Dict model from memory.

## Pull request type

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes

Issue Number: N/A

## What is the new behavior?

-
-
-


src/kakarot/stack.cairo
```diff
@@ -77,30 +88,101 @@ namespace Stack {
...
-        return (new_stack=new_stack, elements=self.elements + new_len);
+        // Read and Copy the elements on an array
+
+        // Generate an array of Uint256* to return
+        let (stack_word_dict) = dict_copy(
+            stack_word_dict=stack_word_dict, stack_len=position_zero, n=n * 2, output=new_elements
+        );
+
+        // Return Stack with updated Len
+        let reduce = 2 * n;
```
Eikix, line 106: not so explicit name -> could be `popped_len` or `popped_elements`, idk wdyt

src/kakarot/stack.cairo
```diff
@@ -12,60 +12,71 @@ from starkware.cairo.common.memcpy import memcpy
...
 // @notice This file contains functions related to the stack.
 // @author @abdelhamidbakhta
 // @custom:namespace Stack
 // @custom:model model.Stack
 namespace Stack {
-    const element_size = Uint256.SIZE;
+    // New Code
+    // Summary of stack. Created upon finalization of the stack.
+    struct Summary {
+        stack_16bytes_len: felt,
```
Eikix, line 30: since this is inside Stack namespace, you don't have to prefix everything by stack imo

Eikix, line 30: `16bytes_len` what does this naming refer to? it's not so clear

danilowhk: They are the len of 16bytes felts dict(for each len, there is a 16byte felt)

src/kakarot/stack.cairo
```diff
@@ -109,41 +191,41 @@ namespace Stack {
...
     // @notice Check stack overflow.
     // @param self - The pointer to the stack.
     // @custom:revert if stack overflow.
-    func check_overflow{range_check_ptr}(self: model.Stack*) {
-        let stack_len = Stack.len(self);
+    func check_overflow{range_check_ptr}(stack_len: felt) {
         // Revert if stack overflow
         with_attr error_message("Kakarot: StackOverflow") {
-            assert_lt_felt(stack_len, Constants.STACK_MAX_DEPTH);
+            assert_lt_felt(stack_len, Constants.STACK_MAX_DEPTH * 2);
```
spapinistarkware, line 228: Use assert_le, it's less expensive

src/kakarot/stack.cairo
```diff
@@ -152,23 +234,12 @@ namespace Stack {
...
     // @param stack_index - The index of the element.
     // @custom:revert if stack underflow.
-    func check_underflow{range_check_ptr}(self: model.Stack*, stack_index: felt) {
+    func check_underflow{range_check_ptr}(stack_len: felt, pop_len: felt) {
         alloc_locals;
-        let stack_len = Stack.len(self);
         // Revert if stack underflow
         with_attr error_message("Kakarot: StackUnderflow") {
-            assert_lt_felt(stack_index, stack_len);
+            assert_lt_felt(pop_len * 2, stack_len + 1);
```
spapinistarkware, line 241: Same

src/kakarot/stack.cairo
```diff
@@ -109,41 +191,41 @@ namespace Stack {
...
 
     // @notice Check stack overflow.
     // @param self - The pointer to the stack.
     // @custom:revert if stack overflow.
-    func check_overflow{range_check_ptr}(self: model.Stack*) {
-        let stack_len = Stack.len(self);
+    func check_overflow{range_check_ptr}(stack_len: felt) {
         // Revert if stack overflow
         with_attr error_message("Kakarot: StackOverflow") {
-            assert_lt_felt(stack_len, Constants.STACK_MAX_DEPTH);
```
spapinistarkware, line 146: I think it makes sense to inline these checks, it's just one function call

spapinistarkware (changes requested):

src/kakarot/stack.cairo
```diff
@@ -77,30 +88,101 @@ namespace Stack {
...
-    func peek{range_check_ptr}(self: model.Stack*, stack_index: felt) -> Uint256 {
-        alloc_locals;
-        Stack.check_underflow(self=self, stack_index=stack_index);
-        let array_index = Stack.get_array_index(self=self, stack_index=stack_index);
-        return self.elements[array_index];
+    func peek{range_check_ptr}(self: model.Stack*, stack_index: felt) -> (
+        self: model.Stack*, value: Uint256
+    ) {
+        let stack_word_dict = self.stack_word_dict;
+        let position_zero = self.stack_16bytes_len;
```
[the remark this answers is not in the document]
danilowhk: Its just the name, basically is the "newest" stack value or Stack[0] or Stack[1](evm.codes dont have stack[0])

src/kakarot/stack.cairo
spapinistarkware: Also, I think here it is enough to check for equality, since only push one at a time.

src/kakarot/stack.cairo
```diff
@@ -77,30 +90,107 @@ namespace Stack {
...
+    // @notice Pop an element from the stack.
+    // @param self - The pointer to the stack.
+    // @return The new pointer to the stack.
+    // @return The popped element.
+    func pop{range_check_ptr}(self: model.Stack*) -> (new_stack: model.Stack*, element: Uint256) {
+        let stack_word_dict = self.stack_word_dict;
+        let position_zero = self.stack_16bytes_len;
+        // Check if stack will underflow
+        with_attr error_message("Kakarot: StackUnderflow") {
+            assert_le(2, position_zero);
```
spapinistarkware, line 150: I actually think both assert_le for pip and push can be replaced with equality checks after every uint256 check.

spapinistarkware (changes requested):

src/kakarot/stack.cairo
danilowhk: Got a bit confused on this one, which uint256 check would it be?

spapinistarkware: For underflow, you need to check that self.stack_16bytes_len != 0.
For overflow, you need to check that self.stack_16bytes_len != MAX.
felt check, not uint256 check.

danilowhk: Oh Ok! I see, for pop_n should we keep the assert_le , or do a "minus operation" and then check if its != 0?

spapinistarkware: You can keep it assert_le at pop_n

spapinistarkware (approved):

src/kakarot/stack.cairo
```diff
@@ -4,68 +4,82 @@
...
-from starkware.cairo.common.math import assert_lt_felt
+from starkware.cairo.common.math import assert_le
 from starkware.cairo.common.uint256 import Uint256
 from starkware.cairo.common.memcpy import memcpy
 
 // Internal dependencies
 from kakarot.constants import Constants
 from kakarot.model import model
 
+// New Stack dependencies
```
ClementWalter, line 15: why splitting "new" and "other"?

src/kakarot/stack.cairo
```diff
@@ -4,68 +4,82 @@
...
+from utils.utils import Helpers
+
 // @title Stack related functions.
 // @notice This file contains functions related to the stack.
 // @author @abdelhamidbakhta
 // @custom:namespace Stack
 // @custom:model model.Stack
 namespace Stack {
-    const element_size = Uint256.SIZE;
+    // New Code
```
ClementWalter, line 27: don't need this

src/kakarot/stack.cairo
```diff
@@ -4,68 +4,82 @@
...
-        // Get length of new stack copy
-        let new_len = self.raw_len - element_size;
-        // Copy stack without last element
-        memcpy(dst=new_elements, src=self.elements, len=new_len);
-        // Create new stack
-        local new_stack: model.Stack* = new model.Stack(elements=new_elements, raw_len=new_len);
-        return (new_stack=new_stack, element=element);
+        if (position_zero == Constants.STACK_MAX_DEPTH * 2 + 2) {
+            with_attr error_message("Kakarot: StackOverflow") {
+                assert 1 = 0;
```
ClementWalter, line 69: why not writing the actual check in the `with_attr`, smth like

```
with_attr error_message(...)
    assert_le(position_zero, Constants.STACK_MAX_DEPTH * 2)
```

src/kakarot/stack.cairo
```diff
@@ -4,68 +4,82 @@
...
+    // @param offset - The offset to store the element at.
     // @return The new pointer to the stack.
     func push{range_check_ptr}(self: model.Stack*, element: Uint256) -> model.Stack* {
-        alloc_locals;
-        Stack.check_overflow(self=self);
-        assert [self.elements + self.raw_len] = element;
-        return new model.Stack(elements=self.elements, raw_len=self.raw_len + element_size);
-    }
+        let stack_word_dict = self.stack_word_dict;
+        let position_zero = self.stack_16bytes_len;
```
ClementWalter, line 65: is `position_zero` `index` or `current_index` ?

src/kakarot/stack.cairo
```diff
@@ -77,30 +91,110 @@ namespace Stack {
         new_stack: model.Stack*, elements: Uint256*
     ) {
         alloc_locals;
```
ClementWalter, line 93: don't need `locals` do we?

src/kakarot/stack.cairo
```diff
@@ -77,30 +91,110 @@ namespace Stack {
...
-        // Get length of new stack copy
-        let new_len = self.raw_len - (element_size * n);
-        // Copy stack without last N elements
-        memcpy(dst=new_elements, src=self.elements, len=new_len);
-
-        // Create new stack
-        local new_stack: model.Stack* = new model.Stack(elements=new_elements, raw_len=new_len);
-        // Return new stack & pointer to first popped element
-        return (new_stack=new_stack, elements=self.elements + new_len);
+        // Read and Copy the elements on an array
```
ClementWalter, line 103: what does this comment refer to?

src/kakarot/stack.cairo
```diff
@@ -77,30 +91,110 @@ namespace Stack {
...
+            new model.Stack(
+            stack_word_dict_start=self.stack_word_dict_start,
+            stack_word_dict=stack_word_dict,
+            stack_16bytes_len=self.stack_16bytes_len - popped_len,
+            ),
+            new_elements,
+        );
+    }
+
+    func dict_copy{range_check_ptr}(
```
ClementWalter, line 122: is this a generic `dict` helper or smth only relevant for the stack context? naming is confusing

ClementWalter: should add docstring

src/kakarot/stack.cairo
```diff
@@ -77,30 +91,110 @@ namespace Stack {
...
+        stack_word_dict: DictAccess*, stack_len: felt, n: felt, output: Uint256*
+    ) -> (stack_word_dict: DictAccess*) {
```
ClementWalter, line 124: why do you return? it seems like this function doesn't return anything but fill the input `output: Uint256*` array

src/kakarot/stack.cairo
ClementWalter: so if I'm right, this function is more like `stack_to_uint256`?

src/kakarot/stack.cairo
```diff
@@ -77,30 +91,110 @@ namespace Stack {
...
+    func pop{range_check_ptr}(self: model.Stack*) -> (new_stack: model.Stack*, element: Uint256) {
+        let stack_word_dict = self.stack_word_dict;
+        let position_zero = self.stack_16bytes_len;
+        // Check if stack will underflow
+
+        if (position_zero == 0) {
+            with_attr error_message("Kakarot: StackUnderflow") {
+                assert 1 = 0;
+            }
+        }
```
ClementWalter, line 155: same as before, why not asserting what we actually want to check?

src/kakarot/stack.cairo
```diff
@@ -77,30 +91,110 @@ namespace Stack {
...
+        return dict_copy(
+            stack_word_dict=stack_word_dict, stack_len=stack_len, n=n - 2, output=output
+        );
+    }
+
+    // @notice Pop an element from the stack.
+    // @param self - The pointer to the stack.
+    // @return The new pointer to the stack.
+    // @return The popped element.
+    func pop{range_check_ptr}(self: model.Stack*) -> (new_stack: model.Stack*, element: Uint256) {
```
ClementWalter, line 146: is this really more efficient than using `pop_n(n=1)` ?

src/kakarot/stack.cairo
```diff
@@ -77,30 +91,110 @@ namespace Stack {
...
     // @param self - The pointer to the stack.
     // @param stack_index - The index of the element to return.
     // @return The element at the given index.
-    func peek{range_check_ptr}(self: model.Stack*, stack_index: felt) -> Uint256 {
-        alloc_locals;
-        Stack.check_underflow(self=self, stack_index=stack_index);
-        let array_index = Stack.get_array_index(self=self, stack_index=stack_index);
-        return self.elements[array_index];
+    func peek{range_check_ptr}(self: model.Stack*, stack_index: felt) -> (
+        self: model.Stack*, value: Uint256
```
ClementWalter, line 178: I don't understand why we need to return the stack as well as it is unchanged in peel

tests/integrations/test_cases/execute.py
```diff
@@ -1353,10 +1353,10 @@
             "events": [
                 [
                     [
-                        0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF,
-                        0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF,
                         0xFF,
                         0x00,
+                        0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF,
+                        0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF,
```
ClementWalter, line 1359: why do we have changes in test while the stack is, though refactor, meant to behave the same?

ClementWalter (changes requested):

src/kakarot/stack.cairo
spapinistarkware: dict_read()s change the dict. That's how they work in Cairo.

src/kakarot/stack.cairo
spapinistarkware: dict_read()s change the dict.

src/kakarot/stack.cairo
spapinistarkware: The assert already happens. if it's 0, we want to raise an error -> revert.
Though, we can use assert_non_zero() here instead.

spapinistarkware: Although, inlining it is better, especially for an op that repeats a lot like stack stuff

src/kakarot/stack.cairo
spapinistarkware: No need for assert_le, it's less efficient.
This check is efficient, see my answer below.

tests/integrations/test_cases/execute.py
danilowhk: Before pop[n] was returning, the "highest" stack index in popped[0] but now its the "lowest index". Basically it have changed the order of the array that we get back on popped, so all I did was to "reverse" the order of the test, in case for logging

src/kakarot/stack.cairo
danilowhk: If we consider the assert_le difference yes, but not sure overall as pop_n does initialize an array and also have to fall amother function, while pop just returns the Unit256

src/kakarot/instructions/arithmetic_operations.cairo
```diff
@@ -61,8 +61,8 @@ namespace ArithmeticOperations {
         // 1 - b: second integer value to add.
         let stack = ctx.stack;
         let (stack, popped) = Stack.pop_n(self=stack, n=2);
-        let a = popped[1];
-        let b = popped[0];
+        let a = popped[0];
+        let b = popped[1];
```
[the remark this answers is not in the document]
danilowhk: Forgot to change the description, basically now the order of values within an index is reversed

src/kakarot/stack.cairo
danilowhk: It was for my own organization, just forgot to change

src/kakarot/stack.cairo
danilowhk: Will check

src/kakarot/stack.cairo
danilowhk: Yes, is more specific to Stack as it is considering packs of 128bits

src/kakarot/stack.cairo
ClementWalter: not sure to understand what you mean by inlining, I was talking about your proposed solution of doing:

```
with_attr error_message("Kakarot: StackUnderflow") {
      assert_non_zero(position_zero);
}
```

src/kakarot/stack.cairo
ClementWalter: @Eikix may be a subject of readability VS optim; and I'd like to see the numbers if possible, otherwise I'd plea for `pop = pop_n(n=1)`

tests/integrations/test_cases/execute.py
ClementWalter: I've realized that pop was reversed, but AFAIU the LOG should not be impacted by implementation of the stack

danilowhk: It is, because it pops "log_n + 2" using pop_n, before was popped[n] and popped[n-1] was the arguments and the rest log info, now the arguments are in popped[0] and popped[1] and the log info are from popped[2] -> popped[n]

src/kakarot/stack.cairo
spapinistarkware: Yeah, it's a bit less efficient, since you have a call and ret instructions for the good flow, instead of 1 instruction in the if

Eikix:
What remains to be "done" before I can merge your PR? 
Since other PRs use the Stack and its signature changed, it's quite important to merge this PR asap (granted it is OK to go)

src/kakarot/model.cairo
```diff
@@ -8,8 +8,9 @@ from starkware.cairo.common.dict import DictAccess
 
 namespace model {
     struct Stack {
-        elements: Uint256*,
-        raw_len: felt,
+        stack_word_dict_start: DictAccess*,
```
ClementWalter: `stack_` prefix is redundant here

src/kakarot/stack.cairo
ClementWalter: so do we change this or nor?

ClementWalter:
> What remains to be "done" before I can merge your PR? Since other PRs use the Stack and its signature changed, it's quite important to merge this PR asap (granted it is OK to go)

I think that there are still some unresolved comments, though not critical, apart from the LOG maybe. I still don't understand why there is a change in these tests

Eikix:
> 

Okay cool:)! @danilowhk if you need anything, don't hesitate to reach out

ClementWalter (approved):

Eikix (approved):
gg 🔥

Issue #327, opened 2022-12-02
bug: RETURNDATASIZE and RETURNDATACOPY


ClementWalter:
# Bug Report

**Kakarot version:**

main branch

**Current behavior:**

These opcodes used the current context return data while they should use the child context one's.

**Expected behavior:**

Use latest sub context return data.

**Related code:**

See `src/kakarot/instructions/environmental_information.cairo` opcodes implementations.

**Other information:**

There is most probably a need for another field in the `ExecutionContext` called `child_context` based on the model of the `parent_context`. Then, these opcodes would just need to use this `ctx.child_context....` instead of `ctx....`.

Issue #260, opened 2022-11-17
feat: disable some opcodes for STATICCALL


matthieuauger:
### Description

0x -

Eikix:
First implementation is proposed by @ClementWalter in #335, will need a few tweaks to be fully operational:)

ClementWalter:
Apart from the gas usage issue common with the 4 CALLs opcodes, STATICCALL needs also to disable few opcodes: 

> The disallowed instructions are [CREATE](https://www.evm.codes/#F0), [CREATE2](https://www.evm.codes/#F5), [LOG0](https://www.evm.codes/#A0), [LOG1](https://www.evm.codes/#A1), [LOG2](https://www.evm.codes/#A2), [LOG3](https://www.evm.codes/#A3), [LOG4](https://www.evm.codes/#A4), [SSTORE](https://www.evm.codes/#55), [SELFDESTRUCT](https://www.evm.codes/#FF) and [CALL](https://www.evm.codes/#F1) if the [value](https://www.evm.codes/#34) sent is not 0.

Issue #258, opened 2022-11-17
feat: implement 0xF4 - DELEGATECALL


matthieuauger:
### Since

Byzantium

### Group

System operations

### Description

0xF4 - DELEGATECALL

see https://www.evm.codes/#f4?fork=merge

Eikix:
First implementation is proposed by @ClementWalter in https://github.com/sayajin-labs/kakarot/pull/335, will need a few tweaks to be fully operational:)

ClementWalter:
See comment for issue #256

Issue #256, opened 2022-11-17
(wip) feat: implement 0x - CALLCODE


matthieuauger:
### Description

0x -

Eikix:
First implementation is proposed by @ClementWalter in https://github.com/sayajin-labs/kakarot/pull/335, will need a few tweaks to be fully operational:)

PR #335, opened 2022-12-03
Add CALLCODE, STATICCALL and DELEGATECALL opcodes

ClementWalter:
## Pull request type

- [x] Bugfix
- [x] Feature
- [ ] Build related changes

The `return_data` of a given context is assigned a new segment with the RETURN opcode.
The RETURNDATASIZE and RETURNDATACOPY use the calling context data instead of the child context one.

Fix: #327 
Implement: #256 #258 #260 

## What is the new behavior?

The `return_data` segment is filled but no new segment is created.
The RETURNDATASIZE and RETURNDATACOPY use the child_context data.
The STATICCALL, DELEGATECALL and CALLCODE are added.

## Other information

It's sort of a hack that I have used to be able to store the ret_offset in memory because we cannot assign a segment in memory.


tests/conftest.py
```diff
@@ -110,7 +110,7 @@ async def kakarot(
     return await starknet.deploy(
         source="./src/kakarot/kakarot.cairo",
         cairo_path=["src"],
-        disable_hint_validation=False,
+        disable_hint_validation=True,
```
matthieuauger, line 113: Why do you remove the hint_validation?

matthieuauger (dismissed):

tests/conftest.py
ClementWalter: because for testing when you want to `print` or `breakpoint` you need to write hints, and each time one needs to disable then enable again. So for testing I think that it's easier not to validate the hint at all. In any case we don't write hint and this wouldn't pass the compile CI step

ClementWalter:
Resources-tests fails because this PR adds one field to the execution context

src/kakarot/instructions/system_operations.cairo
```diff
@@ -152,37 +137,182 @@ namespace SystemOperations {
...
+
+    // @notice STATICCALL operation.
+    // @dev
+    // @custom:since Homestead
+    // @custom:group System Operations
+    // @custom:gas 0 + dynamic gas
+    // @custom:stack_consumed_elements 6
+    // @custom:stack_produced_elements 1
+    // @return The pointer to the sub context.
+    func exec_staticcall{
```
Eikix, line 164: How do we achieve staticcall spec?

> This instructions is equivalent to [CALL](https://www.evm.codes/#F1), except that it does not allow any state modifying instructions or sending ETH in the sub [context](https://www.evm.codes/about). The disallowed instructions are [CREATE](https://www.evm.codes/#F0), [CREATE2](https://www.evm.codes/#F5), [LOG0](https://www.evm.codes/#A0), [LOG1](https://www.evm.codes/#A1), [LOG2](https://www.evm.codes/#A2), [LOG3](https://www.evm.codes/#A3), [LOG4](https://www.evm.codes/#A4), [SSTORE](https://www.evm.codes/#55), [SELFDESTRUCT](https://www.evm.codes/#FF) and [CALL](https://www.evm.codes/#F1) if the [value](https://www.evm.codes/#34) sent is not 0.

I get that we force with_value = 0 to ensure ETH sent with value is 0, but what about the rest? Is this for future implems / diffs?

src/kakarot/instructions/system_operations.cairo
```diff
@@ -152,37 +137,182 @@ namespace SystemOperations {
...
+    // @custom:stack_produced_elements 1
+    // @return The pointer to the sub context.
+    func exec_delegatecall{
+        syscall_ptr: felt*,
+        pedersen_ptr: HashBuiltin*,
+        range_check_ptr,
+        bitwise_ptr: BitwiseBuiltin*,
+    }(ctx: model.ExecutionContext*) -> model.ExecutionContext* {
+        let sub_ctx = exec_staticcall(ctx);
+        let sub_ctx = ExecutionContext.update_addresses(
```
Eikix, line 223: nice workaround:)

src/kakarot/instructions/system_operations.cairo
```diff
@@ -152,37 +137,182 @@ namespace SystemOperations {
...
 
-        return sub_ctx;
+        let ctx = ExecutionContext.update_memory(ctx, memory);
+
+        return (ctx, call_args);
+    }
+
+    // @notice The teardown of CALLs is made in the run loop of the EVMInstructions.
+    // @return The pointer to the context
+    func teardown_call{
```
Eikix, line 294: Could this function name/functionality be more explicit?
I found: 
> In the context of Python programming and the pytest framework, a teardown is a specific type of code that is run after a test has been completed. This code is used to clean up any resources that were allocated or modified during the test, such as closing open files or connections, or rolling back changes to a database. Teardown code is typically placed in a tearDown() method in a test class, and it is automatically run by the pytest framework after each test is completed. This ensures that the tests are self-contained and do not affect each other or the overall state of the system.

Is this what you meant?

src/kakarot/instructions/system_operations.cairo
ClementWalter: good point. I was too focus single CALL/STATICCALL usage and had in mind that value=0 would break any attempt to something that cost, which is wrong.

A possible implementation that comes to me would be to add a flag "static" in the ExecutionContext and update these opcodes to check if the ExecutionContext is allowed to call them. What do you think?

src/kakarot/instructions/system_operations.cairo
ClementWalter: well I named this influenced by the pytest fixture actually but I don't really know how to name this. Open to suggestion; `finalize_call`?

src/kakarot/instructions/system_operations.cairo
Eikix: If you want we can split these into two PRs. Maybe the scope of the PR became too wide?

src/kakarot/instructions/system_operations.cairo
Eikix: I'm not sure, what does the function do? Or what is it's consequence?

ClementWalter: Well, it updates the calling context's stack (with success/failing) and memory (as the memory is immutable, the sub context cannot update it directly).

All the CALLs are almost the same, this is the factorization of the end of these ops (as `prepare_args` is the factorization of the beginning of these ops).

PR #338, opened 2022-12-04
Add tests for bytes_i_to_uint256 method and fixes it for 16 bytes sequences

### LINKED ISSUE #333: bug: Conversion from bytes to uint256 fails for sequences = 16 bytes
# Bug Report

The [bytes_i_to_uint256](https://github.com/sayajin-labs/kakarot/blob/main/src/utils/utils.cairo#L46) does not seem to work for sequences of bytes = 16 bytes following the migration from [hints to pure cairo](https://github.com/sayajin-labs/kakarot/pull/226/files#diff-7125bfc4851cb1b3a82161b692339e996956180ec6a010a5f5dd7504ee97cef9R31).

**Kakarot version:**
v0.1.0.
**Current behavior:**
image

**Steps to reproduce:**

You can checkout the `increase_coverage` branch and run `poetry run pytest -k tests/units/instructions/test_push_operations.py`

matthieuauger:
The bytes_i_to_uint256 was bugged with sequences of exactly 16 bytes.

Issue was caused by a [lower_or_equal]([`main...matthieuauger:kakarot:fix-bytes_to_uint256?expand=1`#diff-7125bfc485](https://github.com/sayajin-labs/kakarot/compare/main...matthieuauger:kakarot:fix-bytes_to_uint256?expand=1#diff-7125bfc4851cb1b3a82161b692339e996956180ec6a010a5f5dd7504ee97cef9)L59) used instead of a strict lower.

- First commit of the PR consists of adding tests for the function.
- Second commit is a refactoring of the function, simplying its implementation and improving naming
- Third commit is the actual bugfix
- Fourth commit is the re-enablement of the exec_push16 test that was blocked because of this bug

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Other (please describe): tests

Fixed #333

### DIFF
--- a/src/utils/utils.cairo
+++ b/src/utils/utils.cairo
@@ -51,17 +51,17 @@ namespace Helpers {
             assert is_sequence_32_bytes_or_less = 1;
         }
 
-        let is_sequence_15_bytes_or_less = is_le(i, 15);
+        let is_sequence_16_bytes_or_less = is_le(i, 16);
 
-        // 1 - 15 bytes
-        if (is_sequence_15_bytes_or_less == TRUE) {
+        // 1 - 16 bytes
+        if (is_sequence_16_bytes_or_less == TRUE) {
             let (low) = compute_half_uint256(val=val, i=i, res=0);
             let res = Uint256(low=low, high=0);
 
             return res;
         }
 
-        // 16 - 32 bytes
+        // 17 - 32 bytes
         let (low) = compute_half_uint256(val=val + i - 16, i=16, res=0);
         let (high) = compute_half_uint256(val=val, i=i - 16, res=0);
         let res = Uint256(low=low, high=high);
--- a/src/utils/utils.cairo
+++ b/src/utils/utils.cairo
@@ -16,7 +16,7 @@ from starkware.cairo.common.memcpy import memcpy
 from starkware.cairo.common.pow import pow
 from starkware.cairo.common.uint256 import Uint256
 from starkware.cairo.common.registers import get_label_location
-from starkware.cairo.common.bool import TRUE, FALSE
+from starkware.cairo.common.bool import FALSE
 
 // @title Helper Functions
 // @notice This file contains a selection of helper function that simplify tasks such as type conversion and bit manipulation
@@ -54,7 +54,7 @@ namespace Helpers {
         let is_sequence_16_bytes_or_less = is_le(i, 16);
 
         // 1 - 16 bytes
-        if (is_sequence_16_bytes_or_less == TRUE) {
+        if (is_sequence_16_bytes_or_less != FALSE) {
             let (low) = compute_half_uint256(val=val, i=i, res=0);
             let res = Uint256(low=low, high=0);
 
--- a/src/utils/utils.cairo
+++ b/src/utils/utils.cairo
@@ -16,7 +16,7 @@ from starkware.cairo.common.memcpy import memcpy
 from starkware.cairo.common.pow import pow
 from starkware.cairo.common.uint256 import Uint256
 from starkware.cairo.common.registers import get_label_location
-from starkware.cairo.common.bool import FALSE
+from starkware.cairo.common.bool import TRUE, FALSE
 
 // @title Helper Functions
 // @notice This file contains a selection of helper function that simplify tasks such as type conversion and bit manipulation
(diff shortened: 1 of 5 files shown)


src/utils/utils.cairo
```diff
@@ -41,44 +41,31 @@ namespace Helpers {
...
 
-        // Check if i is lower to 32
-        let is_le32 = is_le(i, 32);
+        let is_sequence_32_bytes_or_less = is_le(i, 32);
         with_attr error_message("number must be shorter than 32 bytes") {
-            assert is_le32 = 1;
+            assert is_sequence_32_bytes_or_less = 1;
         }
 
-        let is_16_le_i = is_le(16, i);
```
matthieuauger, line 59: bug was caused by this line. is_le(16, i) was taking sequences of 16 to 32 bytes but intention was to take 17 to 32 bytes

src/utils/utils.cairo
```diff
@@ -41,44 +41,31 @@ namespace Helpers {
...
-            new_val = val;
-            new_i = i;
-            high = 0;
-            tempvar range_check_ptr = range_check_ptr;
-        }
+        let is_sequence_16_bytes_or_less = is_le(i, 16);
 
-        let is_i_le_16 = is_le(new_i, 16);
+        // 1 - 16 bytes
+        if (is_sequence_16_bytes_or_less == TRUE) {
```
matthieuauger, line 57: I inversed the conditional to start with short strings then large strings

src/utils/utils.cairo
```diff
@@ -41,44 +41,31 @@ namespace Helpers {
...
+            assert is_sequence_32_bytes_or_less = 1;
         }
 
-        let is_16_le_i = is_le(16, i);
-        if (is_16_le_i != FALSE) {
-            assert new_val = val + i - 16;
-            new_i = 16;
-            let (high_temp) = compute_half_uint256(val=val, i=i - 16, res=0);
-            high = high_temp;
-            tempvar range_check_ptr = range_check_ptr;
```
matthieuauger, line 65: there was a lot of temporary variables and instructions that I factorized to keep function simpler

src/utils/utils.cairo
```diff
@@ -41,44 +41,31 @@ namespace Helpers {
...
-            new_val = val;
-            new_i = i;
-            high = 0;
-            tempvar range_check_ptr = range_check_ptr;
-        }
+        let is_sequence_15_bytes_or_less = is_le(i, 15);
 
-        let is_i_le_16 = is_le(new_i, 16);
+        // 1 - 15 bytes
+        if (is_sequence_15_bytes_or_less == TRUE) {
```
Eikix: I think that, as @spapinistarkware mentioned, we need to always use `FALSE` when checking conditions. It is because checking 0 condition is apparently more optimised than checking against 1

`if (is_sequence_15_bytes_or_less != FALSE)`

src/utils/utils.cairo
```diff
@@ -41,44 +41,31 @@ namespace Helpers {
...
+            let (low) = compute_half_uint256(val=val, i=i, res=0);
```
Eikix, line 58: I think `compute_half_uint256` is not so explicit in its argument or what it does.

Could we add a commit before this one to slightly make `compute_half_uint256` more explicit? (There is no doc comment on the util, no comments inside, and unclear parameter names, unclear flow).

src/utils/utils.cairo
```diff
@@ -51,17 +51,17 @@ namespace Helpers {
             assert is_sequence_32_bytes_or_less = 1;
         }
 
-        let is_sequence_15_bytes_or_less = is_le(i, 15);
+        let is_sequence_16_bytes_or_less = is_le(i, 16);
 
-        // 1 - 15 bytes
-        if (is_sequence_15_bytes_or_less == TRUE) {
+        // 1 - 16 bytes
+        if (is_sequence_16_bytes_or_less == TRUE) {
```
Eikix, line 57: I'm still unsure of how big are the benefits of using `FALSE` always, but this is inline with our recent optimisations:
`if (is_sequence_16_bytes_or_less != FALSE) {`

Eikix (commented):
🔥🔥🔥🔥🔥

src/utils/utils.cairo
matthieuauger: Yeah I agree, I can push the refactoring deeper

src/utils/utils.cairo
ClementWalter: note: @spapinistarkware told us that it's more optimized to compare against FALSE than TRUE, so should prefer `!= FALSE`

src/utils/utils.cairo
ClementWalter: we miss a simple benchmark I do agree

Eikix (approved):
lgtm

PR #343, opened 2022-12-06
refactor: kakarot file architecture

matthieuauger:
See https://github.com/sayajin-labs/kakarot/discussions/337

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes


Eikix (commented):
I wonder whether we need tests/units/src/** we could simply have tests/units/**

Otherwise this is awesome:D 🔥

matthieuauger:
Fair point, I checked some frameworks and indeed they skip the src/ part in the tests folder. I don't think we need it

matthieuauger:
PR is not easy to read, main changes:
- rename units to unit and integrations to integration
- split fixtures (starknet, kakarot, account_registry) in separate folder (fixtures) and files
- move solidity_files in integration folder
- mimic src architecture in tests/unit
- split the big tests/utils in different helpers located in unit or integration + split in multiple folders (reporting.py, wrap_kakarot.py)

tests/conftest.py
```diff
@@ -41,117 +23,8 @@ def event_loop():
...
-        registry_address_=account_registry.contract_address
-    ).execute(caller_address=1)
-    yield
-    await account_registry.transfer_ownership(1).execute(
-        caller_address=kakarot.contract_address
-    )
+pytest_plugins = [
+    "tests.fixtures.1_starknet",
+    "tests.fixtures.2_kakarot",
+    "tests.fixtures.3_account_registry",
```
ClementWalter, line 29: why adding numbering as prefix?

matthieuauger: I added this to help navigate in the fixtures when reading the file tree, the fixtures order matters because kakarot relies on starknet and account registry on kakarot.

It's a tradeoff for evolution though, especially if we need to add fixtures we may need to rename files, I propose to let this and if there is any issue in evolutivity I remove it

matthieuauger:
codecov is decreasing because we had the helpers.cairo() count in codecov but it shouldn't as this is a test util

PR #347, opened 2022-12-07
Stark Exchange contracts deployment tests

danilowhk:
Added tests for deployment of 1 Main contract and 6 Sub Contracts for StarkExchange

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [X] Other (please describe): Tests

Issue Number: N/A

## What is the new behavior?

-
-
-


tests/integration/test_starkExchange.py
```diff
@@ -0,0 +1,73 @@
+from typing import Callable
```
ClementWalter, line 1: don't mix casing in file name: use instead `test_starkexchange.py` or `test_stark_exchange.py`

tests/integration/test_starkExchange.py
```diff
@@ -0,0 +1,73 @@
...
+
+@pytest.mark.asyncio
+@pytest.mark.starkExchange
+class TestStarkExchange:
+    class TestDeploy:
+        async def test_stark_exchange_proxy_utils_contract_deployment(
+            self,
+            deploy_solidity_contract: Callable,
+        ):
+            # 1 ProxyUtils
```
ClementWalter, line 15: comments are redundants

ClementWalter (changes requested):
you need to add the .sol files as well, and it's good!

tests/integration/test_starkExchange.py
danilowhk: Ok!

tests/integration/test_starkExchange.py
danilowhk: Ok!

ClementWalter (approved):

ClementWalter:
So currently the `make build-sol` won't compile these files since the command uses:
```
solidity_folder = $(shell pwd)/tests/integration/solidity_files
solidity_files  = $(shell ls ${solidity_folder} | grep .sol)
```

how did you compile the files in `output`?

danilowhk:
I compiled them in foundry and then copies the Abis and Bins

ClementWalter:
> I compiled them in foundry and then copies the Abis and Bins

you should not use in tests anything that is not fully reproducible, so you'd need to add the compile step somewhere in this codebase

Eikix (changes requested):
Just waiting for docker solidity compile step rather than manual foundry compile

Eikix:
One rebase + 
> Just waiting for docker solidity compile step rather than manual foundry compile as @ClementWalter suggested

And we're good to go imo
wdyt @ClementWalter

danilowhk:
Will wait confirmation to rebase

danilowhk:
Or should I just rebase now?

danilowhk:
Ok! Will work on the compilers and do the rebase

danilowhk:
I was looking into creating a compilation script, because the dir is big it would need very specific script. This would be the case for any other complex contracts, couldnt we have a separate Foundry Repo for complex contracts and we just compare the Bins from those 2?

ClementWalter:
> I was looking into creating a compilation script, because the dir is big it would need very specific script. This would be the case for any other complex contracts, couldnt we have a separate Foundry Repo for complex contracts and we just compare the Bins from those 2?

looking at these tests again, we were just like testing that it was possible to deploy the contracts, not even that they work. Consequently I think that it would make sense not to add these files in kakarot but to create another repo / another package in a monorepo setting were we actually only test solidity contracts using regular soidity dev tools plugged using kakarot as a backend

Eikix:
Why would they belong in another repo?
These are solidity files that we want to deploy through Kakarot
We could add functional tests to this PR or subsequent PR

ClementWalter:
> @ClementWalter why would the solidity files / functional tests belong in another repo? These are solidity files that we want to deploy through Kakarot We could add functional tests to this PR or subsequent PR

I'm ok to merge this to clear this stock PR but, following our discussion in tg main channel, I think that it should be the last one before we tackle seriously the integration of sol written web3 projects with kakarot

Issue #253, opened 2022-11-17
feat: implement 0xF0 - CREATE


TotalPizza:
### Since

Frontier

### Group

System Operations

### Description

0xF0 - CREATE

**TL;DR**
Allows for deployed contract code to trigger the creation of a new contract.

**More specifically:**

Creates a new contract. Enters a new sub [context](https://www.evm.codes/about) of the calculated destination address and executes the provided initialisation code, then resumes the current context.

Should deployment succeed, the new account's [code](https://www.evm.codes/about) is set to the [return data](https://www.evm.codes/about) resulting from executing the initialisation code.

The destination address is calculated as the rightmost 20 bytes (160 bits) of the Keccak-256 hash of the rlp encoding of the sender address followed by its nonce.

## Stack Input
1.) value: value in [wei](https://www.investopedia.com/terms/w/wei.asp) to send to the new account.
2.) offset: byte offset in the [memory](https://www.evm.codes/about) in bytes, the initialisation code for the new account.
3.) size: byte size to copy (size of the initialisation code).

## Stack Output

1.) address: the address of the deployed contract, 0 if the deployment failed.

## Error Cases
- Not enough gas.
- Not enough values on the stack.
- The current execution context is from a [STATICCALL](https://www.evm.codes/#FA) (since Byzantium fork).

TotalPizza:
I'd like to start concepting the implementation on this one.

TotalPizza:
Current blocker is the need for an [rlp](https://ethereum.org/en/developers/docs/data-structures-and-encoding/rlp/) encoding library in order to generate eth addresses in the exact same way as on evm ethereum.

liamzebedee:
@TotalPizza you might be encouraged to find @maxgillet has written some RLP decoding utils here - https://github.com/maxgillett/stark-attestations/blob/master/programs/lib/storage_verification/extract_from_rlp.cairo.

TotalPizza:
Implementation of CREATE can be further discussed in the draft PR:
https://github.com/sayajin-labs/kakarot/pull/293

TotalPizza:
> @TotalPizza you might be encouraged to find @MaxGillet has written some RLP decoding utils here - https://github.com/maxgillett/stark-attestations/blob/master/programs/lib/storage_verification/extract_from_rlp.cairo.

WOAAA thanx dude.
Haven't looked into it yet, but hopefully it'll mean I don't have to write it from scratch!

TotalPizza:
[...]

Issue #259, opened 2022-11-17
feat: implement 0xF5 - CREATE2


matthieuauger:
### Since

Constantinople

### Group

System operations

### Description

0xF5 - CREATE2

see https://www.evm.codes/#f5?fork=merge

PR #351, opened 2022-12-08
feat: CREATE and CREATE2 opcodes

ClementWalter:
## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Resolves: #253 #259 

## Other information

I am wondering if we should push further the factorization of the `kakarot.deploy` and `exec_create` methods.

### DIFF
--- a/src/kakarot/accounts/contract/library.cairo
+++ b/src/kakarot/accounts/contract/library.cairo
@@ -5,6 +5,7 @@
 // Starkware dependencies
 from openzeppelin.access.ownable.library import Ownable
 from starkware.cairo.common.alloc import alloc
+from starkware.cairo.common.bool import FALSE
 from starkware.cairo.common.cairo_builtins import HashBuiltin, BitwiseBuiltin
 from starkware.cairo.common.math import unsigned_div_rem, split_felt
 from starkware.cairo.common.registers import get_label_location
@@ -90,7 +91,7 @@ namespace ContractAccount {
             contract_address_salt=salt,
             constructor_calldata_size=2,
             constructor_calldata=calldata,
-            deploy_from_zero=0,
+            deploy_from_zero=FALSE,
         );
 
         // Generate EVM_contract address from the new cairo contract
--- a/src/kakarot/accounts/contract/library.cairo
+++ b/src/kakarot/accounts/contract/library.cairo
@@ -13,7 +13,7 @@ from starkware.starknet.common.syscalls import deploy as deploy_syscall
 from starkware.starknet.common.syscalls import get_contract_address
 
 from kakarot.constants import native_token_address, registry_address, evm_contract_class_hash
-from kakarot.accounts.registry.library import AccountRegistry
+from kakarot.interfaces.interfaces import IRegistry
 
 // @title SmartContractAccount main library file.
 // @notice This file contains the EVM smart contract account representation logic.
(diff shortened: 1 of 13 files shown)


abdelhamidbakhta (approved):
lgtm

src/kakarot/instructions.cairo
```diff
@@ -612,8 +613,16 @@ namespace EVMInstructions {
             if (is_parent_root != FALSE) {
                 return ctx;
             } else {
-                let ctx = CallHelper.finalize_calling_context(ctx);
-                return run(ctx=ctx);
+                let (bytecode_len) = IEvmContract.bytecode_len(
+                    contract_address=ctx.starknet_contract_address
+                );
+                if (bytecode_len == 0) {
+                    let ctx = CreateHelper.finalize_calling_context(ctx);
```
Eikix, line 620: why do we need this if/else?

src/kakarot/instructions/system_operations.cairo
```diff
@@ -317,3 +390,81 @@ namespace CallHelper {
         return ctx;
     }
 }
+
+namespace CreateHelper {
+    // @notice Deploy a new Contract account and initialize a sub context at these addresses
+    //         with bytecode from calling context memory.
+    // @return The pointer to the updated calling context.
+    func create_sub_context{
```
Eikix, line 398: this naming sounds like you're creating a sub_context rather than it being a sub_context for the CREATE opcode
maybe `sub_context_create` could help? or even `sub_context_create_opcode`

src/kakarot/instructions.cairo
ClementWalter: the CALLs & the CREATEs opcodes both require a finalize when they RETURN, but they don't require the same: CALLs push `success` while `CREATEs push `address`.

src/kakarot/instructions/system_operations.cairo
ClementWalter: !! the CREATE kw is confusing indeed !!
however here I intended to actually say create like initialize, not create like CREATE. We're actually creating a sub context for CREATE (🤔)
Bottom line: `initialize_sub_context` may be better and somehow symmetrical with `finalize_calling_context` in terms of prefix/suffix

src/kakarot/accounts/contract/library.cairo
```diff
@@ -56,6 +67,53 @@ namespace ContractAccount {
...
+        assert [calldata + 1] = 0;
+
+        // Deploy contract account with no bytecode
+        let (class_hash) = evm_contract_class_hash.read();
+        let (starknet_contract_address) = deploy_syscall(
+            class_hash=class_hash,
+            contract_address_salt=salt,
+            constructor_calldata_size=2,
+            constructor_calldata=calldata,
+            deploy_from_zero=0,
```
Eikix, line 93: does that mean we're deploying with `msg.sender = addres(0)`

src/kakarot/instructions.cairo
Eikix: but here the code is identical, no?

ClementWalter: it's `CreateHelper.{}` and `CallHelper.{}`

src/kakarot/accounts/contract/library.cairo
ClementWalter: Well, this was like that already, it's 0 for FALSE, I've put FALSE again

Issue #349, opened 2022-12-08
feat: use gas in CALLs opcodes


ClementWalter:
# Feature Request

The gas args is not used when creating subcontexts.

**Describe Preferred Solution**

Fix `// TODO: Add support for gas limit` in `ExecutionContext.init_at_address`.

**Related Code**

`execution_context.cairo:init_at_address` and opcodes `CALL`, `CALLCODE`,  `STATICCALL` & `DELEGATECALL`.

**Additional Context**

Gas accounting in kakarot is currently erratic.

Eikix:
I think this is a completed issue. What do you think @ClementWalter?

ClementWalter:
> I think this is a completed issue. What do you think @ClementWalter?

it is, see https://github.com/sayajin-labs/kakarot/pull/373

PR #373, opened 2022-12-13
feat: use gas in CALLs opcodes #349

seshanthS:
## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

`gas_limit` is passed as `0` to the sub contexts.

Resolves #[349](https://github.com/sayajin-labs/kakarot/issues/349)

## What is the new behavior?

- Adds `gas_limit` param to `execution_context.init_at_address()`, `kakarot.execute_at_address()`
- Limits the gas passed to sub-contexts to `min(remaining_gas, remaining_gas/64)`

### DIFF
--- a/src/kakarot/instructions/system_operations.cairo
+++ b/src/kakarot/instructions/system_operations.cairo
@@ -212,11 +212,14 @@ namespace SystemOperations {
         range_check_ptr,
         bitwise_ptr: BitwiseBuiltin*,
     }(ctx: model.ExecutionContext*) -> model.ExecutionContext* {
+        alloc_locals;
         let (ctx, call_args) = CallHelper.prepare_args(ctx=ctx, with_value=1);
-
-        // TODO: use gas_limit when init_at_address is updated
+        let remaining_gas = ctx.gas_limit - ctx.gas_used;
+        let (max_allowed_gas, _) = Helpers.div_rem(remaining_gas, 64);
+        let gas_limit = Helpers.min(call_args.gas, max_allowed_gas);
         let sub_ctx = ExecutionContext.init_at_address(
             address=call_args.address,
+            gas_limit=gas_limit,
             calldata_len=call_args.args_size,
             calldata=call_args.calldata,
             value=call_args.value,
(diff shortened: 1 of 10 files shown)


src/kakarot/instructions/system_operations.cairo
```diff
@@ -212,11 +212,14 @@ namespace SystemOperations {
         range_check_ptr,
         bitwise_ptr: BitwiseBuiltin*,
     }(ctx: model.ExecutionContext*) -> model.ExecutionContext* {
+        alloc_locals;
         let (ctx, call_args) = CallHelper.prepare_args(ctx=ctx, with_value=1);
-
-        // TODO: use gas_limit when init_at_address is updated
+        let remaining_gas = ctx.gas_limit - ctx.gas_used;
+        let (max_allowed_gas, _) = Helpers.div_rem(remaining_gas, 64);
+        let gas_limit = Helpers.min(call_args.gas, max_allowed_gas);
```
ClementWalter, line 219: can't we include this in `prepare_args` as it's the same op for all the CALLs codes?

src/utils/utils.cairo
```diff
@@ -478,4 +478,13 @@ namespace Helpers {
         let (q, _) = unsigned_div_rem(bytes_len + 31, 32);
         return q * 32;
     }
+
+    // @notice Returns the min value between a and b
+    func min{range_check_ptr}(a:felt, b: felt) -> felt {
+        if(is_le(a, b) == 1) {
```
ClementWalter, line 484: @spapinistarkware told us one day that it's better (save some resources) to compare against 0 instead of 1 (FALSE instead of TRUE), so better write this the other way around

tests/integration/helpers/wrap_kakarot.py
```diff
@@ -30,6 +30,7 @@ async def _wrapped(contract, *args, **kwargs):
                 call = kakarot.execute_at_address(
                     address=evm_contract_address,
                     value=0,
+                    gas_limit=1000000,
```
ClementWalter, line 33: could we instead do something like:

```
                if "value" in kwargs:
                    value = kwargs["value"]
                    del kwargs["value"]
                else:
                    value = 0
```

for `gas_limit` so that we're able to update it easily in tests? (to be put before the `if`)

tests/unit/src/kakarot/instructions/test_system_operations.cairo
```diff
@@ -102,7 +102,8 @@ func test__exec_call__should_return_a_new_context_based_on_calling_ctx_stack{
     assert sub_ctx.return_data_len = ret_size.low;
     assert [sub_ctx.return_data] = ret_offset.low;
     assert sub_ctx.gas_used = 0;
-    assert sub_ctx.gas_limit = 0;
+    let (gas_felt, _) = Helpers.div_rem(Constants.TRANSACTION_GAS_LIMIT, 64);
+    assert_le_felt(sub_ctx.gas_limit, gas_felt);
```
ClementWalter, line 106: not meaningful here but fyi when it's possible it's better (more optimized) to use `assert_le` instead of `assert_le_felt`

ClementWalter (commented):
looks good! only minor comment and go for the merge

ClementWalter:
resources-check fails because of the addition of the arg in `gas_limit`, not relevant, force merge is ok
image

ClementWalter (approved):

PR #376, opened 2022-12-14
Opcode SELFDESTRUCT implementation

hurrikaanig:
## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

No SELFDESTRUCT opcode

Resolves #<261>

## What is the new behavior?

Implementing SELFDESTRUCT following "possible implementation" described here -> https://github.com/sayajin-labs/kakarot/issues/261#issue-1453122908

-
-
-


src/kakarot/execution_context.cairo
```diff
@@ -67,6 +67,7 @@ namespace ExecutionContext {
     func init(call_context: model.CallContext*) -> model.ExecutionContext* {
         alloc_locals;
         let (empty_return_data: felt*) = alloc();
+        let (empty_destroy_contract: felt*) = alloc();
```
matthieuauger, line 70: Can we have `empty_destroy_contracts` with an s at the end to be consistent?

src/kakarot/instructions.cairo
```diff
@@ -611,6 +611,7 @@ namespace EVMInstructions {
         // Terminate execution
         if (stopped != FALSE) {
             if (is_parent_root != FALSE) {
+                let ctx = SelfDestructHelper.selfdestruct(ctx);
```
matthieuauger, line 614: I think we should call selfdestruct only if the destroy_contracts_len is not 0, it feels risky when reading the whole codeflow that the selfdestruct() is called everytime

src/kakarot/instructions/system_operations.cairo
```diff
@@ -301,6 +301,60 @@ namespace SystemOperations {
...
+        let (stack, address_uint256) = Stack.pop(stack);
+        let address_felt = Helpers.uint256_to_felt(address_uint256);
+        let ctx = ExecutionContext.update_stack(ctx, stack);
+        // Get the number of native tokens owned by the given starknet
+        // account and transfer them to receiver
+        let (native_token_address_) = native_token_address.read();
+        let (balance: Uint256) = IEth.balanceOf(
+            contract_address=native_token_address_, account=ctx.starknet_contract_address
+        );
+        let (success) = IEth.transfer(contract_address=native_token_address_, recipient=address_felt, amount=balance);
```
matthieuauger, line 334: Do you know how errors are managed on these instructions? For example if there is an issue with the selfdestruct later in the code, would that revert the transaction at a whole? I think we should transfer the token after the contract has been destructed instead of before to not credit the client with contract balance if there are issues with selfdestruction

matthieuauger (commented):
Really cool PR 🔥🔥 I made a few comments. @ClementWalter could you have a look as well?

src/kakarot/instructions/system_operations.cairo
hurrikaanig: From my understanding balance is transferred immediately when selfdestruct is called but contract destroyed only at the end of the tx. I don't have any example coming to my mind but transferring balance only at the end of the tx  may bring unwanted behaviors. Also if tx fails everything is reverted (no contract destruction and no balance transfer). So if anything fail in selfdestruct, transfer would also be reverted. Or am I missing something ?

src/kakarot/execution_context.cairo
```diff
@@ -457,6 +480,38 @@ namespace ExecutionContext {
...
             sub_context=self.sub_context,
+            destroy_contracts_len=self.destroy_contracts_len,
+            destroy_contracts=self.destroy_contracts,
+            );
+    }
+
+    // @notice Update the array of contracts to destroy.
+    // @param self The pointer to the execution context.
+    // @param destroy_contracts_len Array new length.
+    // @param destroy_contracts The pointer to the new array of contracts.
```
ClementWalter, line 491: Because no item in this array is supposed to be removed we may prefer having an update method that only pushes items.

src/kakarot/instructions/system_operations.cairo
```diff
@@ -301,6 +301,60 @@ namespace SystemOperations {
...
+            input_arr=ctx.destroy_contracts,
+            output_arr=destroy_contracts,
+        );
+        assert [destroy_contracts + ctx.destroy_contracts_len] = ctx.starknet_contract_address;
+        
+        let ctx = ExecutionContext.update_destroy_contracts(
+            self=ctx,
+            destroy_contracts_len= ctx.destroy_contracts_len + 1,
+            destroy_contracts=destroy_contracts,
+            stop=TRUE,
```
ClementWalter, line 353: Do we really stop the execution context in a SELFDESTRUCT ?

src/kakarot/instructions/system_operations.cairo
```diff
@@ -301,6 +301,60 @@ namespace SystemOperations {
...
+        let (balance: Uint256) = IEth.balanceOf(
+            contract_address=native_token_address_, account=ctx.starknet_contract_address
+        );
+        let (success) = IEth.transfer(contract_address=native_token_address_, recipient=address_felt, amount=balance);
+        with_attr error_message("Kakarot: Transfer failed") {
+            assert success = TRUE;
+        }
+
+        // Save contract to be destroyed at the end of the transaction in a NEW array
+        // Directly filling ctx.destroy_contract produces error "Expected memory address to be relocatable value. Found: 0."
```
ClementWalter, line 340: If we always create a new array, how can we destruct several contracts in the same tx? (My understanding is that it’s possible that several sub context call SELFDESTRUCT and at the end of the tx there are consequently several contracts to be destroyed)

hurrikaanig: Yes you're right. At first I was updating the current array but it was raising this error _"Expected memory address to be relocatable value. Found: 0."_ and I don't really know why but creating a new array solved this issue. Also I'm filling the newly created array with the old one data with Helpers.fill_array() so contract to destroy collected from other contexts are not lost.

src/kakarot/instructions/system_operations.cairo
hurrikaanig: I think so as the opcode description is "**Halt execution** and register account for later deletion", which is similar to STOP opcode "**Halts execution**"

src/kakarot/instructions/system_operations.cairo
ClementWalter: I am quite surprised because for `return_data` for example we don't create a new array but fill with `Memory.n_load` the originally empty array. AFAIK in cairo an "array" is just an initialized segment, so this error may indicate that the pointers were not used correctly. Can you double check?

src/kakarot/instructions/system_operations.cairo
```diff
@@ -467,10 +538,74 @@ namespace CreateHelper {
...
     }
 }
+
+namespace SelfDestructHelper {
+
+    // @notice helper for SELFDESTRUCT operation.
+    // @notice overwrite contract account bytecode with 0s
+    // @notice remove contract from registry
+    // @return The pointer to the updated execution_context.
+    func selfdestruct{
```
ClementWalter, line 569: could we name this this `finalize`? to be similar to the other helpers, and because the `selfdestruct` is already the name of the helper?

src/kakarot/instructions/system_operations.cairo
```diff
@@ -467,10 +538,74 @@ namespace CreateHelper {
...
+
+        // Remove contract from registry
+        let (registry_address_) = registry_address.read();
+        IRegistry.set_account_entry(
+            contract_address=registry_address_,
+            starknet_contract_address=starknet_contract_address,
+            evm_contract_address=0
+        );
+
+        let ctx = ExecutionContext.update_destroy_contracts(
```
ClementWalter, line 602: you should maybe create a loop with a dedicated internal function to avoid creating new ExecutionContext each time, and consequently save some steps.

That being said, it's smooth like that and I don't expect selfdestruct to be called dozens of time so it's probably fine like that

ClementWalter (approved):

src/kakarot/instructions/system_operations.cairo
hurrikaanig: My bad, I had not initialized destroy_contract array in my test

src/kakarot/instructions/system_operations.cairo
hurrikaanig: Good idea, yes maybe not necessary here but I will keep it in mind for further implementations

hurrikaanig: Added it finally

src/kakarot/execution_context.cairo
```diff
@@ -487,18 +487,22 @@ namespace ExecutionContext {
 
     // @notice Update the array of contracts to destroy.
     // @param self The pointer to the execution context.
-    // @param destroy_contracts_len Array new length.
-    // @param destroy_contracts The pointer to the new array of contracts.
-    func update_destroy_contracts(
+    // @param destroy_contracts_len Array length of destroy_contracts to add.
+    // @param destroy_contracts The pointer to the new array of contracts to destroy.
+    func add_n_destroy_contracts(
```
ClementWalter, line 479: `append_to_destroy_contracts`? `push_to_destroy_contracts` ? non blocking though

src/kakarot/execution_context.cairo
```diff
@@ -487,18 +487,22 @@ namespace ExecutionContext {
...
+        Helpers.fill_array(
+            fill_len=destroy_contracts_len,
+            input_arr=destroy_contracts,
+            output_arr=self.destroy_contracts + self.destroy_contracts_len,
+        );
         return new model.ExecutionContext(
             call_context=self.call_context,
             program_counter=self.program_counter,
-            stopped=stop,
+            stopped=self.stopped,
```
ClementWalter, line 492: `stop=TRUE` as per your previous comments?

ClementWalter (dismissed):

src/kakarot/execution_context.cairo
hurrikaanig: Since this one is called to append sub_ctx.destroy_contracts to the calling context I think it shouldn't modify stopped. I am only setting it to true with the single item push_to_destroy_contract function called in exec_selfdestruct

src/kakarot/execution_context.cairo
hurrikaanig: Let's go for push !

src/kakarot/execution_context.cairo
ClementWalter: but as soon as there's something in this array it should be stopped right? Consequently I would enforce the stopped=TRUE otherwise there's a risk that it's actually never stopped (if only using `add_n`, even with n=1)

Also, it's bug prone to have two functions add_1/add_n not behaving the same (imagine someone makes a refacto one day to use it in exec_selfdestruct)

then go to go!

hurrikaanig: I understood that current context should be stopped but not the calling one and it's the calling context that is being updated here. We only want to save contract to destroy from sub_ctx so they can be destroyed at the end of the whole tx. Maybe I can put `stopped` as a parameter to both function and add comment on the wanted behavior so it gets less confusing

ClementWalter: ok let's keep it like that for now

PR #386, opened 2022-12-20
Creating a github action to automatically generate the contract's call graph + readme + fix makefile

gaetbout:
This PR contains:
 - A new github action
 - Updating the readme with a small  decription

- [ ] Bugfix
- [X] Feature
- [X] Code style update (formatting, renaming)
- [ ] Build related changes
- [X] Documentation content changes

## Other information

Time spent on this PR:  

0.5 - 1 day

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -103,6 +103,8 @@ markers = [
     "SolmateERC20",
     "SolmateERC721",
     "IntegrationTestContract",
+    "Precompiles",
+    "EC_RECOVER:     Precompile Value 0x01 - Elliptic curve digital signature algorithm (ECDSA) public key recovery function",
 ]
 
 [tool.isort]
--- a/src/kakarot/instructions.cairo
+++ b/src/kakarot/instructions.cairo
@@ -15,27 +15,28 @@ from starkware.cairo.common.registers import get_label_location
 from starkware.cairo.common.uint256 import Uint256
 
 // Internal dependencies
-from kakarot.model import model
-from kakarot.memory import Memory
 from kakarot.execution_context import ExecutionContext
-from kakarot.stack import Stack
-from kakarot.instructions.push_operations import PushOperations
-from kakarot.instructions.stop_and_arithmetic_operations import StopAndArithmeticOperations
+from kakarot.instructions.block_information import BlockInformation
 from kakarot.instructions.comparison_operations import ComparisonOperations
 from kakarot.instructions.duplication_operations import DuplicationOperations
+from kakarot.instructions.environmental_information import EnvironmentalInformation
 from kakarot.instructions.exchange_operations import ExchangeOperations
 from kakarot.instructions.logging_operations import LoggingOperations
 from kakarot.instructions.memory_operations import MemoryOperations
-from kakarot.instructions.environmental_information import EnvironmentalInformation
-from kakarot.instructions.block_information import BlockInformation
+from kakarot.instructions.push_operations import PushOperations
+from kakarot.instructions.sha3 import Sha3
+from kakarot.instructions.stop_and_arithmetic_operations import StopAndArithmeticOperations
 from kakarot.instructions.system_operations import (
     SystemOperations,
     CallHelper,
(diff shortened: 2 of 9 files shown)


abdelhamidbakhta (changes requested):
The call graph directory and generated files should be ignored. Otherwise looks good to me. Thanks for this useful contribution. I will accept the PR once you ignore the files

gaetbout:
What do you mean by ignored?
If you mean within the .gitignore, they are, otherwise I'm not sure to understand what you mean/ask ^^'

gaetbout:
The thing is that if I push the action will be triggered and those files will be committed because in the action there is the flag to force push.
Also in the action takes care to delete the callgraphs folder to make sure it is always clean :)

Unless there is something I'm missing?

.github/workflows/thoth.yml
```diff
@@ -0,0 +1,69 @@
...
+
+jobs:
+  thoth:
+    runs-on: ubuntu-latest
+    env:
+      CI_COMMIT_MESSAGE: CI Generating contract's call graphs
+      CI_COMMIT_AUTHOR: ${{ github.event.repository.name }} CI
+    steps:
+      # Python setup
+      - name: Set up Python 3.10
```
ClementWalter, line 13: we use python 3.9 in the project, is this on purpose?

README.md
```diff
@@ -208,6 +209,16 @@ Kakarot is provided **"as is"** without any **warranty**. Use at your own risk.
 _For more information and to report security issues, please refer to our
 [security documentation](docs/SECURITY.md)._
 
+## Callgraphs
+
+[The callgrap folder](callgraphs) contains all the contract's call graph. Those
```
ClementWalter, line 214: typo "callgrap"

ClementWalter: contracts call graphes?

src/kakarot/execution_context.cairo
```diff
@@ -503,15 +501,14 @@ namespace ExecutionContext {
             sub_context=self.sub_context,
             destroy_contracts_len=self.destroy_contracts_len + destroy_contracts_len,
             destroy_contracts=self.destroy_contracts,
-        ); 
+            );
```
ClementWalter, line 504: are these changes from `make format`?

ClementWalter:
> The thing is that if I push the action will be triggered and those files will be committed because in the action there is the flag to force push. Also in the action takes care to delete the callgraphs folder to make sure it is always clean :)
> 
> Unless there is something I'm missing?

it seems to me that the purpose of this PR @abdelhamidbakhta is to have these files automatically generated and committed.

To me it's somehow like the resources files, it's painful to have them committed because it's automatically generated files that are an eternal source of conflicts in PRs. I'd suggest @gaetbout not to commit them in the action but:

1. use the artifacts actions like for the resources
2. if you feel so, add a command in makefile to generate them locally

.github/workflows/thoth.yml
gaetbout: while trying to create the action it failed with 3.9.
As I faced many issue to do this action trying python 3.10 was something I tried, and I forgot to revert it, will change it back.

gaetbout:
I understand the conflict issue it can create but as it is generic documentation I don't think "hiding" them in an artifact is a good idea. If you are not aware those are available I wouldn't go and look there.
Maybe making a "documentation" folder that has a callgraph folder is more relevant to regroup all documentation related stuff?
As for generating them locally it would push people to install thoth locally before being able to access them (unless the makefile takes care of doing it, but then it would require admin right (sudo)).
That's just my view on it, but if everyone agrees on doing it your way, I'll do it :)

ClementWalter (commented):
so if I understand correctly, two non conflicting changes in *.cairo files can lead to a conflict in these graph.

Looking at these graphs, I don't feel like they're likely to be embedded somewhere (some docs/markdown files) as their readability would be quite poor (needed to zoom in and scroll a lot)

Consequently I would still prefer having them as artifacts for now.

PR #389, opened 2022-12-21
Documentation model.cairo

hurrikaanig:
## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Documentation content changes

no doc in model

Resolves #<379>

## What is the new behavior?

doc in model.cairo

-
-
-

## Other information

Time spent on this PR:  0.2 days

### DIFF
--- a/src/kakarot/model.cairo
+++ b/src/kakarot/model.cairo
@@ -62,7 +62,7 @@ namespace model {
     // @param starknet_contract_address - starknet address of the contract interacted with
     // @param evm_contract_address - evm address of the contract interacted with
     // @param calling_context - parent context of the current execution context, can be empty when context
-    //                          is root context                      
+    //                          is root context | see execution_context.is_root(ctx)                   
     // @param sub_context - child context of the current execution context, can be empty
     // @param destroy_contracts_len - destroy_contract length
     // @param destroy_contracts - array of contracts to destroy at the end of the transaction
--- a/src/kakarot/model.cairo
+++ b/src/kakarot/model.cairo
@@ -8,10 +8,11 @@ from starkware.cairo.common.dict import DictAccess
 
 namespace model {
 
-    // @dev Stack with a 1024 items size. Each item is a 256 bits word. The stack is used by most 
-    // @dev opcodes to consume their parameters from.
-    // @dev Each element is represented by two 128bits (16bytes) chunks.
-    // @param word_dict_start - pointer to a DictAccess array used to simulate a read/write memory
+    // @notice info: https://www.evm.codes/about#stack
+    // @notice Stack with a 1024 items maximum size. Each item is a 256 bits word. The stack is used by most 
+    // @notice opcodes to consume their parameters from.
+    // @dev Each word is represented by two 128bits (16bytes) chunks.
+    // @param word_dict_start - pointer to a DictAccess array used to store the stack's value at a given index
     // @param word_dict - pointer to the end of the DictAccess array
     // @param len_16_bytes - length of the DictAccess array
     struct Stack {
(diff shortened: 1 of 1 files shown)


hurrikaanig:
Not sure about everything I wrote when documenting it, I'll gladly accept your corrections :)

src/kakarot/model.cairo
```diff
@@ -7,18 +7,36 @@ from starkware.cairo.common.uint256 import Uint256
 from starkware.cairo.common.dict import DictAccess
 
 namespace model {
+
+    // @dev Stack with a 1024 items size. Each item is a 256 bits word. The stack is used by most 
+    // @dev opcodes to consume their parameters from.
+    // @dev Each element is represented by two 128bits (16bytes) chunks.
+    // @param word_dict_start - pointer to a DictAccess array used to simulate a read/write memory
```
ClementWalter, line 14: "memory" is confusing here as there's also the Memory down below.

You can say "DictAccess is used to store the stack's value at a given index" or something like that

src/kakarot/model.cairo
```diff
@@ -7,18 +7,36 @@ from starkware.cairo.common.uint256 import Uint256
...
+    // @param word_dict - pointer to the end of the DictAccess array
+    // @param len_16_bytes - length of the DictAccess array
     struct Stack {
         word_dict_start: DictAccess*,
         word_dict: DictAccess*,
         len_16bytes: felt,
     }
 
+    // @dev Transient memory maintened by the EVM during an execution which doesn't persist
```
ClementWalter, line 23: save comment as above for @notice and https://www.evm.codes/about#memory

also, you should install a spell check in your IDE as you have ! (maintened=>maintained)

src/kakarot/model.cairo
```diff
@@ -7,18 +7,36 @@ from starkware.cairo.common.uint256 import Uint256
...
     struct Memory {
         word_dict_start: DictAccess*,
         word_dict: DictAccess*,
         bytes_len: felt,
     }
 
+    // @dev Struct storing data related to a call
+    // @param bytecode - smart contract bytecode
+    // @param bytecode_len - length of bytecode
+    // @param calldata - byte space where the data parameter of a transaction or call is held
```
ClementWalter, line 37: see and reference https://www.evm.codes/about#calldata

src/kakarot/model.cairo
```diff
@@ -27,6 +45,23 @@ namespace model {
...
+    // @param return_data_len - return_data length
+    // @param stack - current execution context stack
+    // @param memory - current execution context memory
+    // @param gas_used - gas consumed by the current state of the execution
+    // @param gas_limit - maximum amount of gas for the execution
+    // @param gas_price - the amount to pay per unit of gas
+    // @param starknet_contract_address - starknet address of the contract interacted with
+    // @param evm_contract_address - evm address of the contract interacted with
+    // @param calling_context - parent context of the current execution context
+    // @param sub_context - child context of the current execution context
```
ClementWalter, line 62: these might not exist, I mean, there are always in the struct but they can be empty

ClementWalter (approved):

src/kakarot/model.cairo
hurrikaanig: `bytecode` is the bytecode of the smart contract making the call ?

src/kakarot/model.cairo
hurrikaanig: I added `(optional)` at the end, not sure if it's the most accurate term

ClementWalter: you right it's not accurate because there are no optional args in cairo. Better saying smth like "can be empty when context is root context, see `ExecutionContext.is_root`" etc.

src/kakarot/model.cairo
ClementWalter: not especially of the smart contract making the call but the executed bytecode for sure. Your formulation was vague enough to be ok for me but you could say instead smth like "the executed bytecode"

ClementWalter (approved):

PR #398, opened 2022-12-26
Feat staticcall disallows subcontext statechanges (#260)

### LINKED ISSUE #260: feat: disable some opcodes for STATICCALL

jobez:
## Pull request type

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes

Opcodes invoked in staticcall calling context can change state.

Resolves #260 

## What is the new behavior?

Stateful opcodes check against a read_only field in the calling context. When in read only mode, it returns a context with an updated stack and bypasses state changing behavior.

-
-
-

## Other information

Time spent on this PR:  .75 days


src/kakarot/instructions/logging_operations.cairo
```diff
@@ -39,6 +39,13 @@ namespace LoggingOperations {
 
         // Pop offset + size.
         let (stack, popped) = Stack.pop_n(stack, topics_len + 2);
+        let ctx = ExecutionContext.update_stack(ctx, stack);
+
+        // This instruction is disallowed when called from a `staticcall` context, which we demark by a read_only attribute
+        if (ctx.read_only == 1) {
```
ClementWalter, line 45: Should we raise here? Not sure

src/kakarot/instructions/memory_operations.cairo
```diff
@@ -365,6 +365,14 @@ namespace MemoryOperations {
         // 0 - key: key of memory.
         // 1 - value: value for given key.
         let (stack, popped) = Stack.pop_n(self=stack, n=2);
+        // Update context stack.
+        let ctx = ExecutionContext.update_stack(ctx, stack);
+
+        // This instruction is disallowed when called from a `staticcall` context, which we demark by a read_only attribute
+        if (ctx.read_only == 1) {
+            return ctx;
```
ClementWalter, line 373: Same here

src/kakarot/precompiles/precompiles.cairo
```diff
@@ -66,6 +66,7 @@ namespace Precompiles {
             sub_context=cast(0, model.ExecutionContext*),
             destroy_contracts_len=0,
             destroy_contracts=cast(0, felt*),
+            read_only=calling_context.read_only,
```
ClementWalter, line 69: This does not really make sense as this execution context is stopped at construction and is actually only a “placeholder” for the precompile returned data. I’d put false/0 instead

ClementWalter (commented):
Very nice PR. Just left minor comments and we’re good to plain 100% compatibility!

src/kakarot/precompiles/precompiles.cairo
jobez: Great point at being stopped at construction. On it!

src/kakarot/instructions/logging_operations.cairo
jobez: Thinking about it, not raising means that the success/failure result of the callops can't be depended on. The state won't be updated for the calls, but the return of the call would always be one. 

I think it makes sense to raise.

jobez:
Is the call opcode doing transfers based on its value param currently? I can add it if not, while also making sure we check if it isn't being called from a `read_only` context.

I see there's an example of eth transfer in `exec_selfdestruct`, I can follow that example and add it before `ExecutionContext.init_at_address` is called.

ClementWalter:
> Is the call opcode doing transfers based on its value param currently? I can add it if not, while also making sure we check if it isn't being called from a `read_only` context.
> 
> I see there's an example of eth transfer in `exec_selfdestruct`, I can follow that example and add it before `ExecutionContext.init_at_address` is called.

I’m not sure to understand the question. The value args is the value of the execution context, ie the value in `msg.value`.

I agree that this value should be transferred to the called contract but I’m not sure it’s especially related to CALLs opcodes.

That being said, I can’t see any place where we actually transfer this value, sub context or not, @Eikix @abdelhamidbakhta do you know smth about this?

Eikix:
> 

Maybe @jobez was talking about the STATICCALL and the likes:

> Creates a new sub [context](https://www.evm.codes/about) and execute the [code](https://www.evm.codes/about) of the given account, then resumes the current one. Note that an account with no code will return success as true (1).

This instructions is equivalent to [CALL](https://www.evm.codes/#F1), except that it does not allow any state modifying instructions or sending ETH in the sub [context](https://www.evm.codes/about). The disallowed instructions are [CREATE](https://www.evm.codes/#F0), [CREATE2](https://www.evm.codes/#F5), [LOG0](https://www.evm.codes/#A0), [LOG1](https://www.evm.codes/#A1), [LOG2](https://www.evm.codes/#A2), [LOG3](https://www.evm.codes/#A3), [LOG4](https://www.evm.codes/#A4), [SSTORE](https://www.evm.codes/#55), [SELFDESTRUCT](https://www.evm.codes/#FF) and [CALL](https://www.evm.codes/#F1) if the [value](https://www.evm.codes/#34) sent is not 0.

src/kakarot/instructions/system_operations.cairo
```diff
@@ -245,6 +255,7 @@ namespace SystemOperations {
             calling_context=ctx,
             return_data_len=call_args.ret_size,
             return_data=call_args.return_data,
+            read_only=ctx.read_only,
```
ClementWalter, line 258: we should check here that

```
with_attr error_message("Kakarot: StateModificationError") {
            assert ctx.read_only * call_args.value = 0;
```
right? (CALL is allowed if `not read_only or value == 0`)

ClementWalter (commented):
lgtm, apart from the check in CALL. Regarding your comment for value transfer, I don't know how it's implemented in other clients but it could be done at the end of `run` right after `if (stopped != FALSE) {` (https://github.com/sayajin-labs/kakarot/blob/b456851f17f68857820177d96dcb1b831d7fb2f3/src/kakarot/instructions.cairo#L617)

src/kakarot/instructions/system_operations.cairo
jobez: Yep, will add the assertion before the subcontext construction 👍

ClementWalter (changes requested):
Re-reading the whole CALLs things, there's a required change (my bad): once STATICCALL is fixed, the `delegate_call` becomes buggy because it's only doing:

```
let sub_ctx = exec_staticcall(ctx);
let sub_ctx = ExecutionContext.update_addresses(
            sub_ctx, ctx.starknet_contract_address, ctx.evm_contract_address
);
return sub_ctx;
```

which was ok when `static` was not `static` (!) (Note: I wrote this code smell)

`exec_delegatecall` should be update with a. `ExecutionContext.update_read_only(false)` or similar to `exec_staticcall` but with `read_only=FALSE` (note: for these CALLs opcodes, the `msg.sender` should also be updated for some, but I don't think it's supported rn in kakarot — the CALLER opcode returning starknet's `get_caller_address`)

ClementWalter (approved):

ClementWalter:
1.6% more steps on average, quite significant but I don't know how to easily reduce this. This PR adds a field to the `ExecutionContext` and few `if`s.

Go for merge for now as it's non bloquant for `ERCs` contracts; we will go back to optim later on (cairo 1.0 maybe).

PR #401, opened 2022-12-27
feat: integrate warm/cold account concepts in EXTCODEXXX opcodes (#340)

### LINKED ISSUE #340: feat: integrate warm/cold account concepts in EXTCODEXXX opcodes
# Feature Request

As mentioned in the [EVM spec](https://www.evm.codes/about#accesssets): 
> Access sets have been introduced in the hardfork Berlin. They are kept per transaction (and not per call context). Two of them exist: the touched contracts addresses, and the touched contract storage slots. If an address or slot is present in the set, it is called 'warm', otherwise it is 'cold'. The dynamic cost of some opcodes depends on whether the address or slot is warm or cold.
Addresses: a set of contract addresses that have been touched in the current transaction. It is initialized with the sender and receiver (or the new contract address in case of a creation) of the transaction, as well as all the [precompiled contracts](https://www.evm.codes/precompiled). When an opcode accesses an address that is not present in the set, it adds it in it. The relevant opcodes are [EXTCODESIZE](https://www.evm.codes/#3B), [EXTCODECOPY](https://www.evm.codes/#3C), [EXTCODEHASH](https://www.evm.codes/#3F), [BALANCE](https://www.evm.codes/#31), [CALL](https://www.evm.codes/#F1), [CALLCODE](https://www.evm.codes/#F2), [DELEGATECALL](https://www.evm.codes/#F4), [STATICCALL](https://www.evm.codes/#FA), [CREATE](https://www.evm.codes/#F0), [CREATE2](https://www.evm.codes/#F5) and [SELFDESTRUCT](https://www.evm.codes/#FF).
Slots: a set of contract address and their storage slot keys that have been accessed. It is initialized to empty. When an opcode accesses a slot that is not present in the set, it adds it to it. The relevant opcodes are [SLOAD](https://www.evm.codes/#54) and [SSTORE](https://www.evm.codes/#55)
When a context reverts, the sets are also reverted to the state they had before that context.

For now, Kakarot will treat any account as cold and won't cache in memory "warm" accounts.

To do: make apparent these concepts and hard code the fact that accesses are cold. Gas consumed should be parametrised and therefore gas consumed should be in line with cold accounts.

**Describe Alternatives**

**Related Code**

**Additional Context**

Yes / No _(Help can be provided if you need assistance submitting a PR)_

jobez:
## Pull request type

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes

`extcodesize` implementation did not mention the warm/code accounts and how it influences its dynamic gas costs

Resolves #340 

## What is the new behavior?

`extcodesize` implementation now mentions the warm/code accounts and how it influences its dynamic gas costs, and increments gas accordingly.

-
-
-

## Other information

Time spent on this PR:  .25 days

### DIFF
--- a/src/kakarot/instructions/environmental_information.cairo
+++ b/src/kakarot/instructions/environmental_information.cairo
@@ -36,6 +36,7 @@ namespace EnvironmentalInformation {
     const GAS_COST_CODESIZE = 2;
     const GAS_COST_CODECOPY = 3;
     const GAS_COST_GASPRICE = 2;
+    const GAS_COST_EXTCODESIZE = 2600;
     const GAS_COST_EXTCODECOPY = 2600;
     const GAS_COST_RETURNDATASIZE = 2;
     const GAS_COST_RETURNDATACOPY = 3;
@@ -436,7 +437,7 @@ namespace EnvironmentalInformation {
     // @dev Get size of an account’s code
     // @custom:since Frontier
     // @custom:group Environmental Information
-    // @custom:gas 100
+    // @custom:gas 100 || 2600
     // @custom:stack_consumed_elements 1
     // @custom:stack_produced_elements 1
     // @param ctx The pointer to the execution context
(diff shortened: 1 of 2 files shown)


src/kakarot/instructions/environmental_information.cairo
```diff
@@ -461,20 +462,27 @@ namespace EnvironmentalInformation {
         let (starknet_contract_address) = IRegistry.get_starknet_contract_address(
             contract_address=registry_address_, evm_contract_address=address_felt
         );
-        if (starknet_contract_address == 0) {
+        if (starknet_contract_address != 0) {
```
Eikix, line 465: I'm surprised that this works. It's surprising to me that the `ctx` variable assignment inside the scope of an if{} is persisted outside of it.

I would advise against this way of writing out the two possibilities for extcodesize, as it would not work in most languages

src/kakarot/instructions/environmental_information.cairo
```diff
@@ -461,20 +462,27 @@ namespace EnvironmentalInformation {
...
-            contract_address=starknet_contract_address
-        );
-        // bytecode_len cannot be greater than 24k in the EVM
-        let stack = Stack.push(stack, Uint256(low=bytecode_len, high=0));
+            tempvar syscall_ptr = syscall_ptr;
+        }
 
-        // Update context stack.
-        let ctx = ExecutionContext.update_stack(self=ctx, new_stack=stack);
+        // TODO:distinction between warm and cold addresses determines dynamic cost
```
Eikix, line 483: can you add the link to explanation?
https://www.evm.codes/about#accesssets

src/kakarot/instructions/environmental_information.cairo
jobez: I can change it back. The rationale was it allows gas to be updated in once place, instead of twice. Thought this would help avoid bugs.

Eikix (approved):
lgtm

PR #403, opened 2022-12-28
fix: fix basefee test

### LINKED ISSUE #308: dev: unskip BASEFEE opcode test

Eikix:
## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Resolves #308 

## What is the new behavior?

-
-
-

## Other information

Time spent on this PR:  0.1day

### DIFF
--- a/src/kakarot/instructions/block_information.cairo
+++ b/src/kakarot/instructions/block_information.cairo
@@ -309,7 +309,7 @@ namespace BlockInformation {
         bitwise_ptr: BitwiseBuiltin*,
     }(ctx: model.ExecutionContext*) -> model.ExecutionContext* {
         // Get the base fee.
-        let basefee = Helpers.to_uint256(val=0);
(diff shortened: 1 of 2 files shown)


ClementWalter (approved):
So the test skip message was wrong ? Ie the stack was neither the expected nor the output value? (Note: I wrote this skip mark, that’s why I’m asking)

Eikix:
> So the test skip message was wrong ? Ie the stack was neither the expected nor the output value? (Note: I wrote this skip mark, that’s why I’m asking)

I think the person who wrote the test copy pasted the outputed stack from evm.codes with the same bytecode, which is:
```
PUSH1 0xa
BASEFEE
STOP
```

But in our case BASEFEE returns 0, not 10. (For some reason on evm.codes it returns 10)

I modified the test to get rid of PUSH1 0xa (why would we push this?)
And kept only BASEFEE. And in our case it returns always 0.

PR #404, opened 2022-12-28
Feat/unsupported precompile event

### LINKED ISSUE #372: Throw custom exceptions (event) when some Opcode or Precompile is not implemented

jobez:
In order to allow for granular knowledge of what precompile users need most, we include the address of the unimplemented precompile in the `NotImplementedPrecompile` error
 
 ## Pull request type
 
 * [ ]  Bugfix
 * [x]  Feature
 * [ ]  Code style update (formatting, renaming)
 * [ ]  Refactoring (no functional changes, no api changes)
 * [ ]  Build related changes
 * [ ]  Documentation content changes
 * [ ]  Other (please describe): 

Resolves #372
 
## Other information
Time spent on this PR: 0.1d

### DIFF
--- a/src/kakarot/precompiles/precompiles.cairo
+++ b/src/kakarot/precompiles/precompiles.cairo
@@ -139,8 +139,6 @@ namespace Precompiles {
         range_check_ptr,
         bitwise_ptr: BitwiseBuiltin*,
     }(address: felt, _input_len: felt, _input: felt*) {
-        precompile_not_supported.emit(precompile_address=address);
-
         with_attr error_message("Kakarot: NotImplementedPrecompile {address}") {
             assert 0 = 1;
         }
--- a/src/kakarot/precompiles/precompiles.cairo
+++ b/src/kakarot/precompiles/precompiles.cairo
@@ -93,7 +93,7 @@ namespace Precompiles {
         output_len: felt, output: felt*, gas_used: felt
     ) {
         // Compute the corresponding offset in the jump table:
-        // count 1 for "next line" and 3 steps per precompile address: call, precompile, ret
+        // count 1 for "next line" and 4 steps per precompile address: call, precompile, ret
         tempvar offset = 1 + 3 * address;
 
         // Prepare arguments
@@ -101,6 +101,7 @@ namespace Precompiles {
         [ap] = pedersen_ptr, ap++;
         [ap] = range_check_ptr, ap++;
         [ap] = bitwise_ptr, ap++;
+        [ap] = address, ap++;
         [ap] = input_len, ap++;
         [ap] = input, ap++;
 
@@ -137,8 +138,10 @@ namespace Precompiles {
         pedersen_ptr: HashBuiltin*,
         range_check_ptr,
         bitwise_ptr: BitwiseBuiltin*,
-    }(ctx_ptr: model.ExecutionContext*) {
-        with_attr error_message("Kakarot: NotImplementedPrecompile") {
+    }(address: felt, _input_len: felt, _input: felt*) {
+        precompile_not_supported.emit(precompile_address=address);
+
+        with_attr error_message("Kakarot: NotImplementedPrecompile {address}") {
             assert 0 = 1;
         }
         return ();
--- a/src/kakarot/precompiles/precompiles.cairo
+++ b/src/kakarot/precompiles/precompiles.cairo
@@ -93,7 +93,7 @@ namespace Precompiles {
         output_len: felt, output: felt*, gas_used: felt
     ) {
         // Compute the corresponding offset in the jump table:
-        // count 1 for "next line" and 4 steps per precompile address: call, precompile, ret
+        // count 1 for "next line" and 3 steps per precompile address: call, precompile, ret
         tempvar offset = 1 + 3 * address;
 
         // Prepare arguments
--- a/tests/unit/src/kakarot/precompiles/test_precompiles.cairo
+++ b/tests/unit/src/kakarot/precompiles/test_precompiles.cairo
@@ -14,7 +14,7 @@ from utils.utils import Helpers
 from kakarot.constants import Constants
 from kakarot.model import model
 from kakarot.stack import Stack
-from kakarot.instructions.system_operations import SystemOperations 
+from kakarot.instructions.system_operations import SystemOperations
 from tests.unit.helpers.helpers import TestHelpers
 
 @external
--- a/tests/unit/src/kakarot/precompiles/test_precompiles.cairo
+++ b/tests/unit/src/kakarot/precompiles/test_precompiles.cairo
@@ -15,26 +15,23 @@ from kakarot.constants import Constants
 from kakarot.model import model
 from kakarot.stack import Stack
 from kakarot.instructions.system_operations import SystemOperations
+from kakarot.precompiles.precompiles import Precompiles
 from tests.unit.helpers.helpers import TestHelpers
 
 @external
 func test__precompiles_should_throw_on_not_implemented{
     syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr, bitwise_ptr: BitwiseBuiltin*
-}(_address: felt) {
-    // Given
-    alloc_locals;
-
-    let (bytecode) = alloc();
-    local bytecode_len = 0;
-
-    let stack: model.Stack* = Stack.init();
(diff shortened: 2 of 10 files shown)


Eikix (commented):
I thought this approach to be fine but my worry is that it changes the signature of the precompile
My understanding is that precompiles have a fixed signature

src/kakarot/precompiles/precompiles.cairo
```diff
@@ -90,14 +90,15 @@ namespace Precompiles {
         output_len: felt, output: felt*, gas_used: felt
     ) {
         // Compute the corresponding offset in the jump table:
-        // count 1 for "next line" and 4 steps per precompile address: call, opcode, ret
+        // count 1 for "next line" and 4 steps per precompile address: call, precompile, ret
```
ClementWalter, line 93: It’s also 3 and not 4..!

ClementWalter (approved):

ClementWalter (approved):

ClementWalter (approved):

tests/unit/src/kakarot/precompiles/test_precompiles.py
```diff
@@ -0,0 +1,29 @@
...
+    async def test_precompiles(self, precompiles):
+        # we choose an out of range precompile address to get the NotImplementedPrecompile error, and test if it is including the address in the error
+        # note: in our implementation, `Precompiles.is_precompile` checks if an address is within a given range before dispatching, so usually an out of range address would never be passed to `Precompiles.run`
+        address = 10
+        with pytest.raises(Exception) as e:
+            await precompiles.precompiles_should_throw_on_not_implemented(
+                address=address
+            ).call()
+            message = re.search(r"Error message: (.*)", e.value.message)[1]  # type: ignore
+            assert message == "Kakarot: StackUnderflow " + address
```
ClementWalter, line 29: this error is not related to what you try to test but only to the fact that `STATICCALL` is trying to pop values from the stack but your test creates an ExecutionContext with an empty stack.

tests/unit/src/kakarot/precompiles/test_precompiles.cairo
```diff
@@ -0,0 +1,40 @@
+// SPDX-License-Identifier: MIT
```
ClementWalter, line 1: in this file, you may want to test:
- `not_implemented_precompile` raises with the expected message
- `is_precompile` returns TRUE/FALSE appropriately
- `_exec_precompile` raises when `address` is greater than `LAST_PRECOMPILE_ADDRESS` (no particular warnings implemented, we should have a unknown memory location I guess)
- `run` returns an appropriate `ExecutionContext`

This may be out of the scope of this PR to add them all but at least `not_implemented_precompile` should be

tests/unit/src/kakarot/precompiles/test_precompiles.py
```diff
@@ -0,0 +1,29 @@
...
+    return await starknet.deploy(
+        source="./tests/unit/src/kakarot/precompiles/test_precompiles.cairo",
+        cairo_path=["src"],
+        disable_hint_validation=True,
+    )
+
+
+@pytest.mark.asyncio
+class TestPrecompiles:
+    async def test_precompiles(self, precompiles):
```
ClementWalter, line 20: it's better to name this func the same as the one in the corresponding cairo file `precompiles_should_throw_on_not_implemented` for log reason

ClementWalter (changes requested):

tests/unit/src/kakarot/precompiles/test_precompiles.py
jobez: really good catch. makes me realize i should be more careful when I am catching exceptions in testing.

tests/unit/src/kakarot/precompiles/test_precompiles.py
```diff
@@ -0,0 +1,31 @@
...
+        # we choose an out of range precompile address to get the NotImplementedPrecompile error, and test if it is including the address in the error
+        # note: in our implementation, `Precompiles.is_precompile` checks if an address is within a given range before dispatching, so usually an out of range address would never be passed to `Precompiles.run`
+        address = 10
+        with pytest.raises(Exception) as e:
+            await precompiles.test__precompiles_should_throw_on_not_implemented(
+                address=address
+            ).call()
+
+        message = re.search(r"Error message: (.*)", e.value.message)[1]
+        assert message == "Kakarot: NotImplementedPrecompile " + str(address)
```
ClementWalter, line 31: if `LAST_PRECOMPILE_ADDRESS` is `0x9`, I'm quite surprised that `address = 10` raises this error

tests/unit/src/kakarot/precompiles/test_precompiles.cairo
ClementWalter: so the functions should be for example

```
func test_is_precompile_should_return_false_when_address_is_greater_than_last_precompile
func test_is_precompile_should_return_true_when_address_is_lower_than_last_precompile
func test_not_implemented_precompile_should_raise_with_detailed_error_message
func test_run_should_return_a_stopped_execution_context
func test_run_should_return_an_execution_context_with_input_ctx_as_calling_context
func test_run_should_return_an_execution_context_with_precompile_address_as_evm_address
```

I'd not test return data without mocking `_exec_precompile` and I don't know how to mock in our setting

tests/unit/src/kakarot/precompiles/test_precompiles.py
```diff
@@ -0,0 +1,31 @@
...
+async def precompiles(starknet: Starknet):
+    return await starknet.deploy(
+        source="./tests/unit/src/kakarot/precompiles/test_precompiles.cairo",
+        cairo_path=["src"],
+        disable_hint_validation=True,
+    )
+
+
+@pytest.mark.asyncio
+class TestPrecompiles:
```
ClementWalter, line 20: here you create one test function per cairo test function. You can also create sub classes for grouping tests of the same function, like

```
class TestPrecompiles:
    class TestIsPrecompile:
        async def test_should_return_false_when_address_is_greater_than_last_precompile():
            await precompile.func test_run_should_return_an_execution_context_with_proper_calling_context()
...
```

tests/unit/src/kakarot/precompiles/test_precompiles.py
jobez: I am calling `Precompile.runs` directly, which would normally get called after a `Precompiles.is_precompile`

ClementWalter: `run` calls `_exec_precompile` with `address = 10`, consequently `offset = 1 + 30` and I don't know where it jumps, but should not jump to a `not_implemented_precompile` right?

jobez: I see what you mean! When the address is 10, we do see an error w/ message of `Kakarot: NotImplementedPrecompile 10` Am experimenting to to see what the behavior is from 10-100, for curiosity. Will edit back results.

ClementWalter: si in summary, we have:

```
Kakarot: NotImplementedPrecompile 0
don't raise
Kakarot: NotImplementedPrecompile 2
Kakarot: NotImplementedPrecompile 3
don't raise
Kakarot: NotImplementedPrecompile 5
don't raise
Kakarot: NotImplementedPrecompile 7
Kakarot: NotImplementedPrecompile 8
Kakarot: NotImplementedPrecompile 9
Kakarot: NotImplementedPrecompile 10
Kakarot: NotImplementedPrecompile 11
raise something else (unknown memory location?)
```

sounds weird!

ClementWalter (changes requested):

tests/unit/src/kakarot/precompiles/test_precompiles.py
ClementWalter: so bottom line is:
- the function `not_implemented_precompile` is defined right after the jump table
- when jumping to 10, one actually jump into `not_implemented_precompile`...
- when jumping to 11, one jumps somewhere in `not_implemented_precompile` and the program raises *but* error_message is set to `Kakarot: NotImplementedPrecompile 11`
- when jumping to 12 or higher, it does raise an invalid location error

ClementWalter (approved):

ClementWalter (approved):

ClementWalter:
resources evolution is 
image
go for merge

PR #412, opened 2023-01-05
Fix state update is persisted across tests

ClementWalter:
## Pull request type

- [x] Bugfix
- [x] Feature
- [ ] Build related changes

When using a Solidity contracts in tests, state modifications are persisted across different tests.
Furthermore it's not possible to deploy to times the same contract (with different constructor args for example).

## What is the new behavior?

The `StarknetState` is cached in an `autouse` fixture; contracts can be deployed at module level, cached while state is reset after each test to its initial value.

## Other information

I have moved some tests according to TODOs that were written herein (test concerning EXTCODECOPY inside test_erc20 and test_counter.
`Counter.sol` actually had no test so I wrote a bunch of them.

Time spent on this PR:  1 day


Eikix (approved):
lgtm

PR #415, opened 2023-01-06
add unchecked and in place cases for decrementing

jobez:
## Pull request type

Please check the type of change your PR introduces:

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [X] Other (please describe):

Minimal working example of error cases that came up when testing the solmates erc721 impl. The decrementing comes up in `burn` and `transferFrom` and tests involving these had failure. See: https://github.com/transmissions11/solmate/blob/main/src/tokens/ERC721.sol#L178
## What is the current behavior?

Resolves #<Issue number>

## What is the new behavior?

-
-
-

## Other information

Time spent on this PR:  1/4 of a day

PR #418, opened 2023-01-07
Feat/erc721 testing (#353)

### LINKED ISSUE #353: Complete integration tests coverage (solmate tests) for ERC721

jobez:
## Pull request type

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes

Minimal integration testing for the solmates erc721 contract

Resolves #353 

## What is the new behavior?

Richer tests based on those found [here](https://github.com/transmissions11/solmate/blob/main/src/test/ERC721.t.sol)

- All tests involving burning or transferring will fail, because they involve unchecked arithmetic ops.
   - I reproduced a MWE here #415 and in the telegram danilo suggested a fix, which I added to the example and resolved it
   - With those changes in #415, tests should pass 100% (or atleast they do locally)
- I made two changes to the solmates logic to adapt to our test runner, will comment in line but briefly
  - safeMint / safeTransfer were methods that had dispatching based on arguments.  I renamed the variants off each method that take a data argument safeMint2 and safeTransfer2. 
     - Justification: when I tried to call the methods as is, I was seeing the following error:
  ```python
    def get_function_by_identifier(
        fns: Sequence[ContractFunction], identifier: str
    ) -> ContractFunction:
        if len(fns) > 1:
           raise ValueError(
                'Found multiple functions with matching {0}. '
                'Found: {1!r}'.format(identifier, fns)
              ValueError: Found multiple functions with matching name. Found: [<Function safeMint(address,uint256,bytes)>, <Function safeMint(address,uint256)>
   ```
  - In the `ERC721Recipient` contract, there is a public address field named 'from.' I renamed this to from_. 
    - Justification: when trying to call the getter 'from' in python I was getting this error, which I presumed to mean there was some overlap in the kakarot wrapping:
    ```shell
     File "./starknet/kakarot/tests/integration/solidity_contracts/Solmate/test_erc721.py", line 399
       recipient_from = await erc_721_recipient.from()
                                               SyntaxError: invalid syntax
    ```

## Other information

Time spent on this PR:  1 day


tests/integration/solidity_contracts/Solmate/ERC721.sol
```diff
@@ -147,7 +147,7 @@ contract ERC721 {
         );
     }
 
-    function safeTransferFrom(
+    function safeTransferFrom2(
```
jobez, line 150: Changed to avoid ambiguity in method names, otherwise was seeing this error:

   ```
    def get_function_by_identifier(
        fns: Sequence[ContractFunction], identifier: str
    ) -> ContractFunction:
        if len(fns) > 1:
           raise ValueError(
                'Found multiple functions with matching {0}. '
                'Found: {1!r}'.format(identifier, fns)
              ValueError: Found multiple functions with matching name. Found: [<Function safeMint(address,uint256,bytes)>, <Function safeMint(address,uint256)>
  ```

tests/utils/reporting.py
```diff
@@ -24,6 +25,13 @@
 T = TypeVar("T", bound=Callable[..., Any])
 
 
+class CustomJSONEncoder(json.JSONEncoder):
```
jobez, line 28: Changed because `safeMint2` and `safeTransfer2` takes a byte argument, and reporting is dumping arguments via json.dumps.

Without this change, the test runner throws an error that bytes is not json serializable.

tests/fixtures/1_starknet.py
```diff
@@ -24,6 +25,8 @@
 logging.getLogger("asyncio").setLevel(logging.ERROR)
 logger = logging.getLogger()
 
+sys.setrecursionlimit(10000)
```
jobez, line 28: Tests were running successfully on my local computer, but failing in the github action because of errors involving the recursive limit. 

The recursive limit being hit in copying datastructures (in `starknet_snapshot` wasnt because of cyclical references (if they were, the error would have showed up locally), but probably because the datastructures were more complex and hitting the recursive limit.

To test this, I made this (temporary) change to the recursive limit to something high (on my local computer it is set to 1000) to see if that allowed the tests to successfully run.

ClementWalter (approved):

PR #431, opened 2023-01-12
Load revert reason short string

### LINKED ISSUE #416: bug: REVERT does not return the revert reason
# Bug Report

#f03cf627ff871b5795cf561e84720244cf7eca71

**Current behavior:**

REVERT does not use `size` to load the proper data from memory

**Expected behavior:**

REVERT uses all the popped value to load the right amount of data.

**Steps to reproduce:**

Test against the raised error in `test_system_operations::test_revert`

**Related code:**

```
        // TODO: implement loading of the revert reason based on size value,
        // currently limited by short string size
        let size = popped[0];
        let offset = popped[1];

        // Load revert reason from offset
        let (memory, revert_reason_uint256, gas_cost) = Memory.load(memory, offset.low);
        local revert_reason = revert_reason_uint256.low;

        // revert with loaded revert reason short string
        with_attr error_message("Kakarot: Reverted with reason: {revert_reason}") {
            assert TRUE = FALSE;
        }
```

**Other information:**

I don't know the actual limitation of `with_attr error_message`

ClementWalter:
## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

The revert opcode does not load the revert reason but only returns a random number.

Resolves #416 

## What is the new behavior?

The reason is loaded as a bytes array. Only the short string is propagated with the `error_attr`.

## Other information

Time spent on this PR:  1 day

We definitely need to fix REVERT (see #432) because this is close to a nightmare in terms of hard fixing to have something ok for testing.

### DIFF
--- a/src/kakarot/instructions/system_operations.cairo
+++ b/src/kakarot/instructions/system_operations.cairo
@@ -221,9 +221,16 @@ namespace SystemOperations {
         let (memory, gas_cost) = Memory.load_n(memory, size.low, revert_reason_bytes, offset.low);
 
         // revert with loaded revert reason short string: 31 bytes of the last word
-        let revert_reason_uint256 = Helpers.bytes_i_to_uint256(
-            revert_reason_bytes + size.low - 32, 31
-        );
+        let reason_is_single_word = is_le(size.low, 32);
+        if (reason_is_single_word != FALSE) {
+            tempvar initial_byte: felt* = revert_reason_bytes;
+            tempvar actual_size = size.low;
+        } else {
+            tempvar byte_shift = size.low - 32;
+            tempvar initial_byte: felt* = revert_reason_bytes + byte_shift;
+            tempvar actual_size = 31;
+        }
+        let revert_reason_uint256 = Helpers.bytes_i_to_uint256(initial_byte, actual_size);
         local revert_reason = Helpers.uint256_to_felt(revert_reason_uint256);
 
         with_attr error_message("Kakarot: Reverted with reason: {revert_reason}") {
--- a/src/kakarot/instructions/system_operations.cairo
+++ b/src/kakarot/instructions/system_operations.cairo
@@ -222,14 +222,14 @@ namespace SystemOperations {
 
         // revert with loaded revert reason short string
         let truncate_reason = is_le(32, size.low);
-        tempvar reason_actual_size;
+        tempvar reason_short_string_size;
         if (truncate_reason != FALSE) {
-            reason_actual_size = 31;
+            reason_short_string_size = 31;
         } else {
-            reason_actual_size = size.low;
+            reason_short_string_size = size.low;
         }
         let revert_reason_uint256 = Helpers.bytes_i_to_uint256(
-            revert_reason_bytes, reason_actual_size
+            revert_reason_bytes, reason_short_string_size
         );
         local revert_reason = Helpers.uint256_to_felt(revert_reason_uint256);
 
--- a/src/kakarot/instructions/system_operations.cairo
+++ b/src/kakarot/instructions/system_operations.cairo
@@ -220,16 +220,9 @@ namespace SystemOperations {
         let (revert_reason_bytes: felt*) = alloc();
         let (memory, gas_cost) = Memory.load_n(memory, size.low, revert_reason_bytes, offset.low);
 
-        // revert with loaded revert reason short string
-        let truncate_reason = is_le(32, size.low);
-        tempvar reason_short_string_size;
-        if (truncate_reason != FALSE) {
-            reason_short_string_size = 31;
-        } else {
-            reason_short_string_size = size.low;
-        }
+        // revert with loaded revert reason short string: 31 bytes of the last word
         let revert_reason_uint256 = Helpers.bytes_i_to_uint256(
-            revert_reason_bytes, reason_short_string_size
+            revert_reason_bytes + size.low - 32, 31
         );
         local revert_reason = Helpers.uint256_to_felt(revert_reason_uint256);
 
--- a/src/kakarot/instructions/system_operations.cairo
+++ b/src/kakarot/instructions/system_operations.cairo
@@ -218,7 +218,7 @@ namespace SystemOperations {
 
         // Load revert reason from offset
         let (revert_reason_bytes: felt*) = alloc();
-        let (memory, gas_cost) = Memory.load_n(memory, 32, revert_reason_bytes, offset.low);
+        let (memory, gas_cost) = Memory.load_n(memory, size.low, revert_reason_bytes, offset.low);
 
         // revert with loaded revert reason short string
         let truncate_reason = is_le(32, size.low);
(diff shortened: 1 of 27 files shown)


src/kakarot/instructions/system_operations.cairo
```diff
@@ -212,19 +212,31 @@ namespace SystemOperations {
...
-        // currently limited by short string size
+
         let size = popped[0];
         let offset = popped[1];
 
         // Load revert reason from offset
-        let (memory, revert_reason_uint256, gas_cost) = Memory.load(memory, offset.low);
-        local revert_reason = revert_reason_uint256.low;
+        let (revert_reason_bytes: felt*) = alloc();
+        let (memory, gas_cost) = Memory.load_n(memory, 32, revert_reason_bytes, offset.low);
```
Eikix, line 221: why use `load_n` rather than `load` for 32 bytes? I remember @spapinistarkware mentioned that load is more optimized

src/kakarot/instructions/system_operations.cairo
```diff
@@ -212,19 +212,31 @@ namespace SystemOperations {
...
         let offset = popped[1];
 
         // Load revert reason from offset
-        let (memory, revert_reason_uint256, gas_cost) = Memory.load(memory, offset.low);
-        local revert_reason = revert_reason_uint256.low;
+        let (revert_reason_bytes: felt*) = alloc();
+        let (memory, gas_cost) = Memory.load_n(memory, size.low, revert_reason_bytes, offset.low);
 
         // revert with loaded revert reason short string
+        let truncate_reason = is_le(32, size.low);
```
Eikix, line 224: Is this less than or equal? In which case maybe there is a problem with the edge case 32?

src/kakarot/instructions/system_operations.cairo
```diff
@@ -212,19 +212,31 @@ namespace SystemOperations {
...
+        tempvar reason_actual_size;
+        if (truncate_reason != FALSE) {
+            reason_actual_size = 31;
```
Eikix, line 227: maybe `reason_truncated_size`? Or `reason_short_string_size`?

src/kakarot/instructions/system_operations.cairo
ClementWalter: a felt can contains `31` bytes, so if the request size is `32` of greater, we need to truncate to `31`, right?

src/kakarot/instructions/system_operations.cairo
```diff
@@ -212,19 +212,31 @@ namespace SystemOperations {
...
+        let (memory, gas_cost) = Memory.load_n(memory, size.low, revert_reason_bytes, offset.low);
+
+        // revert with loaded revert reason short string: 31 bytes of the last word
+        let reason_is_single_word = is_le(size.low, 32);
+        if (reason_is_single_word != FALSE) {
+            tempvar initial_byte: felt* = revert_reason_bytes;
+            tempvar actual_size = size.low;
+        } else {
+            tempvar byte_shift = size.low - 32;
+            tempvar initial_byte: felt* = revert_reason_bytes + byte_shift;
```
Eikix, line 230: does this mean that we take the 31 last bytes and shift to the last 32 bytes? Does this mean that essentially the very last byte is ignored?

tests/integration/solidity_contracts/Solmate/test_erc20.py
```diff
@@ -80,6 +72,20 @@ async def test_should_transfer(self, erc_20, owner, other):
             assert await erc_20.balanceOf(owner.address) == 0
             assert await erc_20.balanceOf(other.address) == TEST_SUPPLY
 
+        async def test_transfer_should_fail_when_insufficient_balance(
+            self, erc_20, owner, other
+        ):
+            await erc_20.mint(
+                owner.address, TEST_AMOUNT, caller_address=owner.starknet_address
+            )
+            with kakarot_error():
```
Eikix, line 81: are we asserting on the error message itself? 
as was done before:  `assert message == "Kakarot: Reverted with reason: 12884901888"`

tests/integration/solidity_contracts/Solmate/test_erc20.py
```diff
@@ -193,18 +175,14 @@ async def test_transfer_from_should_fail_when_insufficient_balance(
                 TEST_SUPPLY,
                 caller_address=other.starknet_address,
             )
-            with pytest.raises(Exception) as e:
+            with kakarot_error():
```
Eikix, line 178: same here `assert message == "Kakarot: Reverted with reason: 12884901888"`

tests/integration/solidity_contracts/Solmate/test_erc20.py
```diff
@@ -285,7 +259,7 @@ async def test_permit_should_fail_with_bad_deadline(
                 deadline,
             )
             v, r, s = ec_sign(digest, owner.private_key)
-            with pytest.raises(Exception) as e:
+            with kakarot_error("PERMIT_DEADLINE_EXPIRED"):
```
Eikix, line 262: nice!

tests/integration/solidity_contracts/UniswapV2/test_uniswap_v2_erc20.py
```diff
@@ -86,26 +86,20 @@ async def test_should_transfer_token_when_signer_is_owner(
         async def test_should_fail_when_amount_is_greater_than_balance_and_balance_not_zero(
             self, token, owner, other
         ):
-            with pytest.raises(Exception) as e:
+            with kakarot_error():
```
Eikix, line 89: same here `assert message == "Kakarot: Reverted with reason: 147028384"` we cannot check that the returned error is relevant?

src/kakarot/instructions/system_operations.cairo
ClementWalter: yes I've assumed that the last word (`bytes32`)  is more relevant than the first one (because of how `string` encoding works) but eventually the revert_reason is a `bytes31` (a `felt`) so we can keep it all, and consequently I've decided to use the first 31 bytes of the last word, or in other words we drop the very last byte.

I don't know tbh if it's possible to have a size that is not a multiple of `32`.

tests/integration/solidity_contracts/Solmate/test_erc20.py
ClementWalter: the numbers returned before didn't necessarily have sense. It's possible that the code revert with no reason actually, and it's often what is used in Solmate's contrat (see the `require` are only used for `permit`).


They use the fact that underflow will revert to avoid writing `require` themselves and in this case, the `revert_reason` is `[]` (size = 0).

tests/integration/solidity_contracts/Solmate/test_erc20.py
ClementWalter: see above

tests/integration/solidity_contracts/UniswapV2/test_uniswap_v2_erc20.py
ClementWalter: see above

Eikix (approved):
lgtm

PR #440, opened 2023-01-16
Add tests for CREATE2 and require in PlainOpcodes

ClementWalter:
Time spent on this PR: 1 day

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Facing an error when testing Univ2Factory, I have eventually manage to isolate the error case in a `PlainOpcode.sol` test.

The error is illustrated in `class TestRequire` that is currently skipped (see also create issue https://github.com/sayajin-labs/kakarot/issues/439)

### DIFF
--- a/tests/fixtures/2_kakarot.py
+++ b/tests/fixtures/2_kakarot.py
@@ -13,7 +13,7 @@ async def eth(starknet: Starknet):
 
 
 @pytest_asyncio.fixture(scope="session")
-async def contract_account_class(starknet: Starknet):
+async def contract_account_class(starknet: Starknet) -> DeclaredClass:
     return await starknet.declare(
         source="./src/kakarot/accounts/contract/contract_account.cairo",
         cairo_path=["src"],
--- a/tests/integration/helpers/helpers.py
+++ b/tests/integration/helpers/helpers.py
@@ -73,22 +73,16 @@ def get_domain_separator(name: str, token_address: str) -> bytes:
 
 
 def get_create2_address(
-    factory_address: str, token_a: str, token_b: str, bytecode: bytes
+    sender_address: str, salt: bytes, initialization_code: bytes
 ) -> str:
-    token_0, token_1 = sorted([token_a, token_b])
-    create2_inputs = [
-        b"\xff",
-        decode_hex(factory_address),
+    """
+    See [CREATE2](https://www.evm.codes/#f5)
+    """
+    return to_checksum_address(
         keccak(
-            encode_abi(
-                ["address", "address"],
-                [token_0, token_1],
-            )
-        ),
-        keccak(bytecode),
-    ]
-    sanitized_inputs = b"".join(create2_inputs[2:])
-    return to_checksum_address(keccak(sanitized_inputs)[-40:])
+            b"\xff" + decode_hex(sender_address) + salt + keccak(initialization_code)
+        ).hex()[-40:]
+    )
 
 
 def get_approval_digest(
--- a/tests/integration/helpers/helpers.py
+++ b/tests/integration/helpers/helpers.py
@@ -81,7 +81,7 @@ def get_create2_address(
     return to_checksum_address(
         keccak(
             b"\xff" + decode_hex(sender_address) + salt + keccak(initialization_code)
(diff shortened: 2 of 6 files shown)


tests/fixtures/2_kakarot.py
```diff
@@ -13,7 +13,7 @@ async def eth(starknet: Starknet):
 
 
 @pytest_asyncio.fixture(scope="session")
-async def contract_account_class(starknet: Starknet):
+async def contract_account_class(starknet: Starknet) -> DeclaredClass:
```
jobez, line 16: I was wondering if you could share your suggestion about when adding typing info to code we write in python tests.

Should we always be looking to do that or is it more important at the level of fixtures/helpers?

tests/integration/solidity_contracts/PlainOpcodes/test_plain_opcodes.py
```diff
@@ -345,3 +345,44 @@ async def test_should_emit_log3(self, plain_opcodes, addresses, event):
...
+            self,
+            plain_opcodes,
+            counter,
+            addresses,
+            get_starknet_address,
+            get_solidity_contract,
+        ):
+            salt = 1234
+            evm_address = await plain_opcodes.create2(
+                counter.constructor().data_in_transaction,
```
jobez, line 360: doing an assignment of the bytecode before passing it to the invocation would help the readability of this test, imo. had to scroll back up to the defined method in plaincodes.sol otherwise 
```
bytecode = counter.constructor().data_in_transaction
```

tests/integration/helpers/helpers.py
```diff
@@ -73,22 +73,16 @@ def get_domain_separator(name: str, token_address: str) -> bytes:
...
-                ["address", "address"],
-                [token_0, token_1],
-            )
-        ),
-        keccak(bytecode),
-    ]
-    sanitized_inputs = b"".join(create2_inputs[2:])
-    return to_checksum_address(keccak(sanitized_inputs)[-40:])
+            b"\xff" + decode_hex(sender_address) + salt + keccak(initialization_code)
+        ).hex()[-40:]
```
jobez, line 84: Doing the slice on the bytes that keccak returns (which would be -20) then converting to hex might be easier to follow, as it would mirror the cairo version and would require one less byte -> hex mapping a dev has to do in their head in a day.

jobez (commented):
Some small readability comments. This looks great!

tests/fixtures/2_kakarot.py
ClementWalter: good question. Bottom line is: we could (should ? but in a couple of months we'll probably move to rust) strive to use mypy (python type checking) but it'd be quite a lot of rework. In the meantime, I somehow add these randomly when I want my IDE to understand better what I'm doing.

tests/integration/solidity_contracts/PlainOpcodes/test_plain_opcodes.py
ClementWalter: make sense when using args like that. I think that `wrap_for_kakarot` is supposed to handle kwargs, I'll check one or the other option

ClementWalter: I'll use `kwarg` : 

```
            evm_address = await plain_opcodes.create2(
                bytecode=counter.constructor().data_in_transaction,
                salt=salt,
                caller_address=addresses[0].starknet_address,
            )
```

if it's ok to you

jobez: it is, indeed, okay with me 👍

tests/fixtures/2_kakarot.py
jobez: Yeah, as a python noob it took me a while to realize some of the comments like 

```
# type: ignore
```

was conveying information meant for mypy haha. For a while I thought it was just people copy and pasting an old comment.

At readtime, I like the type annotations. I think using them at the level of fixtures/helpers makes sense!

jobez (approved):
lgtm!

PR #441, opened 2023-01-16
Unit tests for RLP

Flydexo:
Added unit tests for the RLP helper used in the Ethereum Account Abstraction feature.

Time spent on this PR: 2 days

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Other (please describe): unit tests

The tests found out that the bytes_to_felt function was incorrect. (endian)

## What is the new behavior?

bytes_to_felt returns the correct results and rlp decoding and decoding are tested


tests/unit/src/utils/test_rlp.py
```diff
@@ -0,0 +1,113 @@
+import os
+from random import randint
+
+import pytest
+import pytest_asyncio
+from rlp import decode
```
ClementWalter, line 6: plz add this lib to the project requirements (`poetry add {package-name} --group dev`)

tests/unit/src/utils/test_rlp.py
```diff
@@ -0,0 +1,113 @@
...
+    return await starknet.deploy(
+        source="./tests/unit/src/utils/test_rlp.cairo",
+        cairo_path=["src"],
+        disable_hint_validation=False,
+    )
+
+
+@pytest.fixture
+def data_to_list_long():
+    return list(
```
ClementWalter, line 20: I'd rather have

```
list(bytes.fromhex("80850430e2340082520894dc544d1aa88ff8bbd2f2aec754b1f1e99e1812fd01801ba0110d8fee1de53df0870e6eb599ed3bf68fb3f1e62c82dfe5976c467c97253b15a03450b73d2aef2009f026bcbf097a257ae7a37eb5d3b73dc0760aefad2b98e327"))
```

as it's shorter and char interpolation in your byte string is confusing

tests/unit/src/utils/test_rlp.py
```diff
@@ -0,0 +1,113 @@
...
+            "f90135010a840adc656b8508d6c03c4a8303088e941111111254fb6c44bac0bed2854e76f90643097d80b8c82e95b6c800000000000000000000000095ad61b0a150d79219dcf64e1e6cc01f0b64c4ce000000000000000000000000000000000000000000006c36df1cfb2498bfc4fa000000000000000000000000000000000000000000000000000000000051a4a60000000000000000000000000000000000000000000000000000000000000080000000000000000000000000000000000000000000000000000000000000000100000000000000003b6d0340773dd321873fe70553acc295b1b49a104d968cc80bd34b36c001a03ef4c950835e3402d10615a9ca96b9143921e378e80d8ceebd5b07710cb03657a028e3f573e7484eff9867045923e61e75d29f65b9af2fb82b7a57a163742eda70"
+        ),
+        bytes.fromhex(
+            "f86f830216c58506676ef5ec826b6c941cedc0f3af8f9841b0a1f5c1a4ddc6e1a1629074880101009bbb1fb5aa8026a0044e77af97e063a12b87fbcc083eae2b4b8daeaac46f967b5dcc82cfa1725192a06a9626195a8430f83676b3c1ca8037bbf5d2108161b0aaf07968a2cd442dc8ef"
+        ),
+    ]
+
+
+@pytest.mark.asyncio
+class TestRLP:
```
ClementWalter, line 53: classes names are usually (in this project):

```
class Test{Namespace/Class}:
    class Test{function under testing}:

        def test_should_{}
```

tests/unit/src/utils/test_rlp.py
ClementWalter: also, I don't really understand what `data_to_list` means

tests/unit/src/utils/test_rlp.py
```diff
@@ -0,0 +1,113 @@
...
+    class TestRLPListEncode:
+        async def test__encode_rlp_list_longer_55_bytes(self, rlp, data_to_list_long):
+            rlp_list = await rlp.test__encode_rlp_list(data_to_list_long).call()
+            data_len = len(data_to_list_long)
+            data_len_len = len(
```
ClementWalter, line 58: `len(int.to_bytes(_len, 'big'))` will just be `_len`; in short

```
data_len_len = math.ceil(data_len.bit_length() / 8)
# or as you wrote
data_len_len = (data_len.bit_length() + 7) // 8
```

tests/unit/src/utils/test_rlp.py
```diff
@@ -0,0 +1,113 @@
...
+                data_len.to_bytes((data_len.bit_length() + 7) // 8, "big")
+            )
+            prefix = 0xF7 + data_len_len
+            expected_list = [
+                prefix,
+                *list(data_len.to_bytes((data_len.bit_length() + 7) // 8, "big")),
```
ClementWalter, line 64: `data_len_len`

tests/unit/src/utils/test_rlp.py
```diff
@@ -0,0 +1,113 @@
...
+
+        async def test__decode_string_le_55(self, rlp):
+            string = list(os.urandom(randint(2, 54)))
+            prefix = 0x80 + len(string)
+            decoded = await rlp.test__rlp_decode_at_index([prefix, *string], 0).call()
+            assert decoded.result.data == string
+            assert decoded.result.is_list == False
+
+        async def test__decode_string_gt_55(self, rlp):
+            data_len = randint(56, 100000)
```
ClementWalter, line 91: fix seed when using random numbers in tests

tests/unit/src/utils/test_rlp.py
```diff
@@ -0,0 +1,113 @@
...
+def data_to_list_short():
+    return list(
+        bytes.fromhex(
+            "80850430e2340082520801801ba0110d8fee1de53df0870e6eb599ed3bf68fb3f1e62c82dfe5976c467c97253b15"
+        )
+    )
+
+
+# using https://etherscan.io/getRawTx?tx= to get samples
+def raw_tx_samples():
```
ClementWalter, line 35: you don't really need to make this a function, you could instead use

```
RAW_TX_SAMPLES = [
...
]
```

ClementWalter (changes requested):

tests/unit/src/utils/test_rlp.py
Flydexo: Included in web3 package

pyproject.toml
```diff
@@ -23,11 +23,12 @@ setproctitle = "^1.3.2"
 pysha3 = "^1.0.2"
 web3 = "^5.31.1"
 python-dotenv = "^0.21.0"
-pycrypto = "2.6.1"
 eth-utils = "<2.0.0"
 eth-abi = "<3.0.0"
 eth-account = "<0.8.0"
 eth-keys = "<0.4.0"
+autoflake = "^2.0.0"
```
ClementWalter, line 30: I think that you need to rebase

tests/unit/src/utils/test_rlp.py
ClementWalter: yes but what if web3 stops including it in a coming version?

The purpose of the `pyproject.toml` (as `package.json` for npm/yarn for example) is to list everything that is **directly** used by the project.
Then the package manager generates a `.lock` files that actually resolves all packages and their dependencies to find corresponding appropriate versions.

In our case, you will see:

```
$ poetry add rlp --group dev
Using version ^3.0.0 for rlp

Updating dependencies
Resolving dependencies... (0.2s)

Because no versions of rlp match >3.0.0,<4.0.0
 and rlp (3.0.0) depends on eth-utils (>=2.0.0,<3), rlp (>=3.0.0,<4.0.0) requires eth-utils (>=2.0.0,<3).
So, because kakarot depends on both eth-utils (<2.0.0) and rlp (^3.0.0), version solving failed.
```

ie the resolver tells you that not all versions of `rlp` are fine but that we need to fix a limit to the version, so:

```
$ poetry add "rlp<3" --group dev

Updating dependencies
Resolving dependencies... (1.0s)

Writing lock file

No dependencies to install or update
```

and in this later case, it:
- works
- does not actually install anything

tests/unit/src/utils/test_rlp.py
```diff
@@ -0,0 +1,112 @@
...
+    ),
+    bytes.fromhex(
+        "f86f830216c58506676ef5ec826b6c941cedc0f3af8f9841b0a1f5c1a4ddc6e1a1629074880101009bbb1fb5aa8026a0044e77af97e063a12b87fbcc083eae2b4b8daeaac46f967b5dcc82cfa1725192a06a9626195a8430f83676b3c1ca8037bbf5d2108161b0aaf07968a2cd442dc8ef"
+    ),
+]
+
+
+@pytest.mark.asyncio
+class TestRLP:
+    class TestRLPListEncode:
```
ClementWalter, line 55: looking back at the `rlp.cairo`, I think that you should updaet the cairo to be 

```
namspace RLP {
    func encode{}
    func decode{}
}
```

so tests would be:

```
class TestRLP:
    class TestEncode:
    class TestDecode:
```

the `rlp` in the function name in redundant with the namespace, and I don't really understand why you would add `list` to encode

tests/unit/src/utils/test_rlp.py
```diff
@@ -0,0 +1,112 @@
...
+        async def test_should_encode_rlp_list_longer_55_bytes(self, rlp, long_list):
```
ClementWalter, line 56: can't we just do one single parametrized test for `encode` and one for `decode`? and just comparing the output of the lib to the python reference implementation ?

```
@pytest.mark.parametrize("bytes_len", [54, 55, 56, 126, 127, 128])
async def test_encode_should_match_reference_implementation(self, bytes_len):
    bytes_array = random.getrandbits(8 * bytes_len)
    res = await rlp.test__encode_rlp_list(bytes_array).call()
    assert encode(bytes_array) == res
```

etc.

ClementWalter (changes requested):

tests/unit/src/utils/test_rlp.py
Flydexo: Because only the list encoding is needed, not all the data types encoding

tests/unit/src/utils/test_rlp.py
```diff
@@ -0,0 +1,101 @@
...
+        "f86f830216c58506676ef5ec826b6c941cedc0f3af8f9841b0a1f5c1a4ddc6e1a1629074880101009bbb1fb5aa8026a0044e77af97e063a12b87fbcc083eae2b4b8daeaac46f967b5dcc82cfa1725192a06a9626195a8430f83676b3c1ca8037bbf5d2108161b0aaf07968a2cd442dc8ef"
+    ),
+]
+
+
+@pytest.mark.asyncio
+class TestRLP:
+    class TestRLPListEncode:
+        async def test_should_encode_list_longer_55_bytes(self, rlp, long_list):
+            rlp_list = await rlp.test__encode_list(long_list).call()
```
ClementWalter, line 63: wouldn't it be easier to just compare against the reference `encode` func you've imported from `rlp` ?

tests/unit/src/utils/test_rlp.py
```diff
@@ -0,0 +1,101 @@
+import pytest
+import pytest_asyncio
+from rlp import decode, encode
+
+# cannot parametrize and use random data at the same time
```
ClementWalter, line 5: I don't understand why you can't parametrize on `data_len` and use
```
random.seed(data_len)
random.getrandbits(data_len * 8)
```
inside a test
Ju

ClementWalter (changes requested):
seems almost good

tests/unit/src/utils/test_rlp.py
```diff
@@ -0,0 +1,106 @@
...
+    # to encode: ["0x","0x0430e23400","0x5208","0x01","0x","0x1b","0x110d8fee1de53df0870e6eb599ed3bf68fb3f1e62c82dfe5976c467c97253b15"]
+    return list(
+        bytes.fromhex(
+            "ee80850430e2340082520801801ba0110d8fee1de53df0870e6eb599ed3bf68fb3f1e62c82dfe5976c467c97253b15"
+        )
+    )
+
+
+# using https://etherscan.io/getRawTx?tx= to get samples
+RAW_TX_SAMPLES = [
```
ClementWalter, line 51: looks like we could use the lib instead of these copy/pasted bytes?
https://pyrlp.readthedocs.io/en/latest/quickstart.html

ClementWalter (changes requested):
plz review tests are they still look a bit erratic

ClementWalter (approved):
you need to rebase and re-run `poetry add`  to fix the `poetry.lock` conflict

jobez:
I also used the RLP namespace using the old function name:

Flydexo:
> Just a heads up, I also used the RLP namespace using the old function name. That would need a small change to have the tests passing:
> 
> https://github.com/sayajin-labs/kakarot/blob/fe03ff003c0f04b96a02572899ca6c8a8e613e0a/src/kakarot/instructions/system_operations.cairo#L585

Yes I changed it

Issue #342, opened 2022-12-06
bug: deployed contracts bytecode differs from local bytecode


jobez:
# Bug Report

When doing integration tests for excodecopy, I ran into issues comparing the bytecode as it is stored in the registry (as written by deploy logic) from the bytecode as it exists locally.

**Kakarot version:**

v0.1.0.
**Current behavior:**

Extcodecopy writes to memory a contracts bytecode, but the bytecode in memory does not match the bytecode of the contract as compiled locally.

**Expected behavior:**

We expect the contract bytecode as written to memory to be the same as the locally compiled bytecode. Although there may be exceptions: https://docs.soliditylang.org/en/develop/contracts.html#call-protection-for-libraries. Needs clarification/investigation.

**Steps to reproduce:**

The two minimal cases can be seen by unskipping and running:

```
poetry run pytest -s -k test_extcodecopy_offset_and_size_gte_twenty_c
```

and

```
poetry run pytest -s -k test_extcodecopy_erc20
```

**Related code:**

You can see the logic in each of these tests.

```
insert short code snippets here
```

**Other information:**

@ClementWalter mentioned on telegram when I shared my first investigation of this 

> interesting. I realized as well that the test on bytecode (deployed VS stored) was failing for ERC20 as well, we had once an issue that I closed after refactoring the test. Codecopy is part of the non-tested opcodes so we’ll probably see something here 👀

Eikix:
#369 does not solve #342, right ?

ClementWalter:
I don't think so because when doing #369 we still add issues with deployed bytecode, which probably uses CODECOPY (not sure though)

PR #448, opened 2023-01-19
Add test for ORIGIN and CALLER (+some refacto)

ClementWalter:
Time spent on this PR: 1 day

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Resolves https://github.com/sayajin-labs/kakarot/issues/342

No test to illustrate the bug of the ORIGIN opcode

## What is the new behavior?

Added a `originAndSender` function in `PlainOpcodes.sol` to check that `tx.origin` and `msg.sender` are correctly returned when called both from an EAO and from a contract.

## Other information

Some minor refacto along the way, especially I added a forge test to check my .sol function to be tested in kakarot (!).
We may want to start diff testing one day or another


tests/integration/solidity_contracts/PlainOpcodes/PlainOpcodes.t.sol
```diff
@@ -0,0 +1,45 @@
+pragma solidity 0.8.10;
```
jobez, line 1: Is this test file run against forge and then we compare results with kakarot somewhere?

would be good to have a top level comment contextualizing it.

tests/integration/solidity_contracts/PlainOpcodes/test_plain_opcodes.py
```diff
@@ -71,242 +67,14 @@ async def test_should_return_self_address(
...
-
-            expected_memory_result = hex_string_to_bytes_array(erc_20.bytecode.hex())
-            memory_result = extract_memory_from_execute(res.result)
-            # asserting to the first discrepancy
-            assert memory_result[dest_offset:2] == expected_memory_result[offset:2]
+        @pytest.mark.parametrize("offset, size", [[0, 32], [32, 32], [0, None]])
+        async def test_should_return_counter_code(
+            self, plain_opcodes, counter, offset, size
+        ):
+            deployed_bytecode = counter.bytecode[counter.bytecode.index(0xFE) + 1 :]
```
jobez, line 74: Can you leave a comment here explaining how this range was derived ?

I asked monsieur chat gpt and it filled me in:

> The reason for this is that the "0xFE" hex value is used as a marker for the beginning of the deployed bytecode. This is because the bytecode that is stored on the blockchain is different from the bytecode of the contract's source code. The bytecode stored on the blockchain includes additional information like the contract's deployed address and any libraries that it references.
> 
> So this python code is looking for the first occurrence of "0xFE" in the bytecode, which marks the beginning of the deployed bytecode, and then slicing the bytecode to include everything after that point. This ensures that the "deployed_bytecode" variable only includes the actual bytecode that is deployed on the blockchain, and not any additional information.
> 
> 

This was sort of the context I was missing for the #342 issue, right?

tests/integration/solidity_contracts/StarkEx/interfaces/ProxySupport.sol
```diff
@@ -14,11 +14,15 @@ import "./ContractInitializer.sol";
...
+abstract contract ProxySupport is
+    Governance,
+    BlockDirectCall,
+    ContractInitializer
+{
     using Addresses for address;
 
     // The two function below (isFrozen & initialize) needed to bind to the Proxy.
-    function isFrozen() external view virtual returns (bool) {
+    function isFrozen() external pure virtual returns (bool) {
```
jobez, line 25: were these changes made by a linter or something to that effect?

tests/integration/solidity_contracts/UniswapV2/conftest.py
```diff
@@ -0,0 +1,72 @@
...
+
+@pytest_asyncio.fixture(scope="module")
+async def pair(
+    deploy_solidity_contract: Callable,
+    token_a,
+    token_b,
+    factory,
+):
+    # TODO: the fixture should use factory.createPair but this currently fails
+    # TODO: with OUT_OF_RESOURCES so we do it "manually" for the sake of running
```
jobez, line 55: do you have a model / explanation for why the `factory.createPair` fails but manual doesn't? i am curious.

tests/integration/solidity_contracts/PlainOpcodes/PlainOpcodes.t.sol
```diff
@@ -0,0 +1,45 @@
...
+contract PlainOpcodesTest is Test {
+    PlainOpcodes plainOpcodes;
+    Counter counter;
+
+    function setUp() public {
+        counter = new Counter();
+        plainOpcodes = new PlainOpcodes(address(counter));
+    }
+
+    function testOpcodeExtCodeCopyReturnsCounterCode(
```
jobez, line 16: having a couple high level comments around the assembly blocks explaining what the outcome is would help with readability, imo.

jobez (changes requested):
Two contextualizing comments, particularly 

- one around the file `PlainOpcodes.t.sol` and its main testing method
- the usage of `counter.bytecode.index(0xFE)` and its relevance to differentiate the deployed bytecode and source bytcode

and it looks good to me! ggs!

tests/integration/solidity_contracts/PlainOpcodes/PlainOpcodes.t.sol
ClementWalter: yes it's run against `forge`, I wrote this test to debug my `opcodeExtCodeCopyReturnsCounterCode`.

The idea was that we write solidity to test kakarot, but if this solidity is not trivial, how can we debug it? Since we use forge to compile, I guesses that the easiest way was to write a forge test...
This is not run in the CI, was just for debug purpose, but left it here because:
- I expect this (testing the solidity before testing it with kakarot) to be a common need
- we may use `forge test` and diff testing for kakarot

ClementWalter: I'm updating the README#test section instead to detail this usecase

tests/integration/solidity_contracts/PlainOpcodes/test_plain_opcodes.py
ClementWalter: Actually when you deploy a contract:
- you send some bytecode to address 0
- this bytecode is run as any bytecode
- the returned data of this run is the one deployed

And I thinks that it is only a convention (maybe depending on the lib?) to have the bytecode as follows:

```
   constructor bytecode                          contract bytecode                                 calldata
|------------------------FE|-------------------------------------------|--------------------|
```

the constructor bytecode is indeed not stored in the contract as it would be useless (constructor is only called at deploy time).

In other words, when you deploy a contract, you only run "constructor bytecode" (up to FE), and the output of this run is "contract bytecode" + possible initialization (Counter as no constructor indeed), which is stored

tests/integration/solidity_contracts/StarkEx/interfaces/ProxySupport.sol
ClementWalter: yes and no, this `view` to `pure` I made it myself because forge was complaining, the other parts of the file were linted automatically

tests/integration/solidity_contracts/UniswapV2/conftest.py
ClementWalter: it does some more ops so if we're close to the limit we then are just above the 1M step. This needs to be confirmed though, I'll push the Univ2Pair tests in a forthcoming PR

jobez (dismissed):
lgtm! great introduction of using forge to verify solidity functions.

jobez (approved):

jobez (approved):

Makefile
```diff
@@ -84,4 +84,4 @@ get-blockhashes:
 	poetry run python scripts/get_latest_blockhashes.py
 
 build-sol:
-	forge build --contracts tests/integration/solidity_contracts -o tests/integration/solidity_contracts/build
+	forge build --force
```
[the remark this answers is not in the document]
ClementWalter: I don't undertstand what's going on, I don't have any error locally

jobez (approved):
Maybe we can make an issue to add some documentation of best practices when devs are introducing solidity files, as there seems to be some necessary treatments to get things to play nice with forge, right?

PR #455, opened 2023-01-23
create opcode uses deterministic evm contract address

### LINKED ISSUE #426: bug: the EVM address of a deployed contract is wrong
# Bug Report

**Current behavior:**

The EVM address of a contract account is computed as follows:

```
        // Generate EVM_contract address from the new cairo contract
        // TODO: Use RLP to compute proper EVM address, see https://www.evm.codes/#f0
        let (_, low) = split_felt(starknet_contract_address);
        local evm_contract_address = 0xAbdE100700000000000000000000000000000000 + low;
```

**Expected behavior:**

Depending on wether the call is a `CREATE` or a `CREATE2`, the address should be computed with

- CREATE (see https://www.evm.codes/#f0)
```
address = keccak256(rlp([sender_address,sender_nonce]))[12:]
```
- CREATE2 (see https://www.evm.codes/#f5)
```
address = keccak256(0xff + sender_address + salt + keccak256(initialisation_code))[12:]
```

**Related code:**

See `ContractAccount.deploy` function.
RLP utils are added in https://github.com/sayajin-labs/kakarot/pull/422

jobez:
Time spent on this PR: 1/2 day

## Pull request type

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes

Create opcode uses an easter egg 🥚 to determine the evm contract mapping.

Resolves #426 

## What is the new behavior?

The create opcode uses `CreateHelper.get_create_address` which derives the address via:

```
address = keccak256(rlp([sender_address,sender_nonce]))[12:]
```

-
-
-

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,6 +29,7 @@ eth-account = "<0.8.0"
 eth-keys = "<0.4.0"
 autoflake = "^2.0.0"
 pycryptodome = "^3.16.0"
+rlp = "<3"
 
 [tool.pytest.ini_options]
 filterwarnings = [
--- a/src/kakarot/instructions/system_operations.cairo
+++ b/src/kakarot/instructions/system_operations.cairo
@@ -10,7 +10,7 @@ from starkware.cairo.common.cairo_keccak.keccak import keccak_bigend, finalize_k
 from starkware.cairo.common.math import split_felt
 from starkware.cairo.common.math_cmp import is_le, is_not_zero, is_nn
 from starkware.cairo.common.memcpy import memcpy
-from starkware.cairo.common.uint256 import Uint256, uint256_and
+from starkware.cairo.common.uint256 import Uint256
 from starkware.starknet.common.syscalls import deploy as deploy_syscall
 from starkware.starknet.common.syscalls import get_contract_address
 
@@ -29,6 +29,7 @@ from kakarot.interfaces.interfaces import IEvmContract, IRegistry, IEth
 from kakarot.memory import Memory
 from kakarot.model import model
 from kakarot.stack import Stack
+from utils.rlp import RLP
 from utils.utils import Helpers
 
 // @title System operations opcodes.
@@ -551,9 +552,58 @@ namespace CreateHelper {
         pedersen_ptr: HashBuiltin*,
         range_check_ptr,
         bitwise_ptr: BitwiseBuiltin*,
-    }(sender_address: felt, nonce: felt) -> (evm_contract_address: felt) {
-        // TODO: see https://github.com/sayajin-labs/kakarot/issues/426
-        return (0xAbdE100700000000000000000000000000000000 + nonce,);
+    }(sender_address: felt, salt: felt) -> (evm_contract_address: felt) {
+        alloc_locals;
+        let (keccak_ptr: felt*) = alloc();
+        local keccak_ptr_start: felt* = keccak_ptr;
+
+        let (local address_packed_bytes: felt*) = alloc();
+
+        // pack sender address, padded twenty bytes
+        // the address should be twenty bytes, so we skip the leading 12 elements
+        let (sender_address_high, sender_address_low) = split_felt(sender_address);
+
+        let (address_packed_bytes_len) = Helpers.uint256_to_dest_bytes_array(
+            value=Uint256(low=sender_address_low, high=sender_address_high),
+            byte_array_offset=12,
+            byte_array_len=Constants.ADDRESS_BYTES_LEN,
+            dest_offset=0,
+            dest_len=0,
+            dest=address_packed_bytes,
+        );
+
+        // encode address rlp
+        let (local packed_bytes: felt*) = alloc();
+        let (packed_bytes_len) = RLP.encode_element_byte_array(
+            address_packed_bytes_len, address_packed_bytes, 0, packed_bytes
+        );
+
+        // encode salt rlp
+        let (packed_bytes_len) = RLP.encode_element_felt(salt, packed_bytes_len, packed_bytes);
+
+        let (local rlp_list: felt*) = alloc();
+        let (rlp_list_len: felt) = RLP.encode_rlp_list(packed_bytes_len, packed_bytes, rlp_list);
+
+        let (local packed_bytes8: felt*) = alloc();
+        Helpers.bytes_to_bytes8_little_endian(
+            bytes_len=rlp_list_len,
+            bytes=rlp_list,
+            index=0,
+            size=rlp_list_len,
+            bytes8=0,
+            bytes8_shift=0,
+            dest=packed_bytes8,
+            dest_index=0,
+        );
+
+        with keccak_ptr {
+            let (create_hash) = keccak_bigend(inputs=packed_bytes8, n_bytes=rlp_list_len);
+
+            finalize_keccak(keccak_ptr_start=keccak_ptr_start, keccak_ptr_end=keccak_ptr);
+        }
+
+        let create_address = Helpers.keccak_hash_to_evm_contract_address(create_hash);
+        return (create_address,);
     }
 
     // @notice Constructs an evm contract address for the create2 opcode
--- a/src/kakarot/instructions/system_operations.cairo
+++ b/src/kakarot/instructions/system_operations.cairo
@@ -574,12 +574,12 @@ namespace CreateHelper {
 
         // encode address rlp
         let (local packed_bytes: felt*) = alloc();
-        let (packed_bytes_len) = RLP.encode_element_byte_array(
+        let (packed_bytes_len) = RLP.encode_byte_array(
             address_packed_bytes_len, address_packed_bytes, 0, packed_bytes
         );
 
         // encode salt rlp
-        let (packed_bytes_len) = RLP.encode_element_felt(salt, packed_bytes_len, packed_bytes);
+        let (packed_bytes_len) = RLP.encode_felt(salt, packed_bytes_len, packed_bytes);
 
         let (local rlp_list: felt*) = alloc();
         let (rlp_list_len: felt) = RLP.encode_rlp_list(packed_bytes_len, packed_bytes, rlp_list);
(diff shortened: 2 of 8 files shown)


src/utils/rlp.cairo
```diff
@@ -140,4 +141,86 @@ namespace RLP {
...
+            // single byte needs no prefix
+            assert [rlp + rlp_len] = element;
+            return (rlp_len + 1,);
+        } else {
+            // otherwise, we need to convert the value to an unpadded byte array
+            let (local reversed_element_bytes: felt*) = alloc();
+            let (local element_bytes: felt*) = alloc();
+
+            let (element_high, element_low) = split_felt(element);
+            let (element_bytes_len) = Helpers.uint256_to_bytes_no_padding(
```
jobez, line 170: when testing against the case where the salt was sufficiently large [255,` 2**55-1), the Helpers.felt_to_bytes_array was returning incorrect results.

This motivated creating this helper that encoded a uint256 with no padding for RLP encoding.

src/utils/utils.cairo
```diff
@@ -236,6 +243,29 @@ namespace Helpers {
         return (bytes_array_len=32, bytes_array=value_as_bytes_array);
     }
 
+    // @notice This function is a variant of `uint256_to_bytes_array` that encodes the uint256 with no padding
+    // @param value: value to convert.
+    // @param idx: index of res array
+    // @param res: resultant encoded bytearray, but in reverse order
+    // @param dest: reversed res, putting byte array in right order
+    // @return bytes array len
+    func uint256_to_bytes_no_padding{bitwise_ptr: BitwiseBuiltin*, range_check_ptr}(
```
jobez, line 252: To reiterate the motivation for this helper:

We encode the salt/nonce in rlp, which means translating it to a byte array when its value is greater than 127.

When I tested the salt in the case 2^55-1, `Helpers.felt_to_bytes_array` was returning a byte array of length one with [2^55-1], downstream causing incorrect rlp encoding than the python implementation.

So I opted to lift the value as a uint256 to handle a greater range accurately and needed to write this helper to get the byte array that took the uint256 value and translated it into a byte array with no padding.

tests/unit/src/kakarot/instructions/test_system_operations.py
```diff
@@ -71,8 +71,9 @@ async def test_call(
             contract_account_class.class_hash, account_registry.contract_address
         ).call()
 
+    @pytest.mark.parametrize("salt", [127, 256, 2**55 - 1])
```
jobez, line 74: 127 would fit in a single byte, meaning in rlp terms we wouldn't need a prefix.

256 and 2**55-1 both need prefixes

src/utils/rlp.cairo
```diff
@@ -140,4 +141,86 @@ namespace RLP {
             return (rlp_len=data_len + 1);
         }
     }
+
+    // @notice encodes felt into an rlp element
+    // @param element The felt that is encoded into rlp
+    // @param rlp_len The length of the rlp array
+    // @param rlp The pointer receiving the rlp encoded felt
+    // @return rlp_len The length of the encoded felt in bytes
+    func encode_element_felt{
```
jobez, line 150: motivation: to encode the salt/nonce value in rlp element (which has distinct logic from an rlp list).

see: https://ethereum.org/en/developers/docs/data-structures-and-encoding/rlp/

jobez (commented):
Currently in draft to have some overview of the motivation/logic of the PR.

pyproject.toml
```diff
@@ -29,6 +29,7 @@ eth-account = "<0.8.0"
 eth-keys = "<0.4.0"
 autoflake = "^2.0.0"
 pycryptodome = "^3.16.0"
+rlp = "<3"
```
jobez, line 32: seems as if i need to commit an updated lock file

the output of the error is saying to run

```
poetry lock --no-update
```

Is it as simple as that and then pushing the updated lock file?

src/utils/rlp.cairo
ClementWalter: could be named just `encode_felt`?

src/utils/rlp.cairo
```diff
@@ -140,4 +141,86 @@ namespace RLP {
...
+        syscall_ptr: felt*,
+        pedersen_ptr: HashBuiltin*,
+        bitwise_ptr: BitwiseBuiltin*,
+        range_check_ptr,
+    }(element: felt, rlp_len: felt, rlp: felt*) -> (rlp_len: felt) {
+        alloc_locals;
+        // 127 is the largest value that can be represented by one byte
+        let element_is_single_byte = is_le(element, 127);
```
ClementWalter, line 158: is `element` an "item" (as per your link)? if yes, then plz name it `item`

src/utils/rlp.cairo
ClementWalter: we have a terrible naming imho from the very beginning regarding `bytes`, `padding` etc.

We should strive to use the same naming convention as the EVM, ie:
- word = `bytes32` = `uint256`
- `bytes_i_` is the _i_ first bytes from the left:
    - `bytes31(bytes32)` will drop the right most byte
    - `bytes1(bytes32)` will keep the left most byte
- `uint_i_` works the opposite (as uint are big endian)
    - for example, with no overflow, `uint64(uint256)` = `uint64`
    - so `uint248(uint256)` will drop the left most byte
    - `uint8(uint256)` will keep the right most byte

`bytes` is a `bytes1[]`, ie an array of bytes and consequently should not have any leading or trailing 0s added for padding

In short, I think that we should once for all create a `BytesLib`, put all these helpers with standard naming conventions and tests, and stop to think about it!

src/utils/rlp.cairo
```diff
@@ -140,4 +141,86 @@ namespace RLP {
...
+    // @return rlp_len The length of the encoded list in bytes
+    func encode_element_byte_array{
+        syscall_ptr: felt*,
+        pedersen_ptr: HashBuiltin*,
+        bitwise_ptr: BitwiseBuiltin*,
+        range_check_ptr,
+    }(byte_array_len, byte_array: felt*, rlp_len: felt, rlp: felt*) -> (rlp_len: felt) {
+        alloc_locals;
+
+        let element_is_le_55 = is_le(byte_array_len, 55);
```
ClementWalter, line 197: `item` ?

tests/unit/src/kakarot/instructions/test_system_operations.py
```diff
@@ -71,8 +71,9 @@ async def test_call(
...
     async def test_create(
-        self, system_operations, contract_account_class, account_registry
+        self, system_operations, contract_account_class, account_registry, salt
```
ClementWalter, line 76: you don't use `salt` in the test 🤔

ClementWalter (changes requested):

pyproject.toml
ClementWalter: no, it's easier 😇

actually when you want to add a lib, you don't (never) update the `pyproject.toml` by hand. Instead, you want `poetry` (or the other package manager you use) to make it for you as it'll check dependencies versions consistency.

So you should use instead

```
poetry add "rlp<3" --group dev
```

and it'll be update the `pyproject.toml`, the `poetry.lock` file and install the lib in your local interpreter.

jobez: thank you!

src/utils/rlp.cairo
jobez: Are you ok with `Helpers.uint256_to_bytes_no_padding` as a name for now?

src/utils/utils.cairo
```diff
@@ -236,6 +243,29 @@ namespace Helpers {
         return (bytes_array_len=32, bytes_array=value_as_bytes_array);
     }
 
+    // @notice This function is a variant of `uint256_to_bytes_array` that encodes the uint256 with no padding
+    // @param value: value to convert.
```
jobez, line 247: convention question:

What is the preferred format for these document comments? In some places it looks like:

https://github.com/sayajin-labs/kakarot/blob/main/src/kakarot/stack.cairo#L55

and here I went with the flow and used colons.

src/utils/rlp.cairo
ClementWalter: can you add an issue for this refactoring?

src/kakarot/instructions/system_operations.cairo
```diff
@@ -551,9 +552,58 @@ namespace CreateHelper {
         pedersen_ptr: HashBuiltin*,
         range_check_ptr,
         bitwise_ptr: BitwiseBuiltin*,
-    }(sender_address: felt, nonce: felt) -> (evm_contract_address: felt) {
-        // TODO: see https://github.com/sayajin-labs/kakarot/issues/426
-        return (0xAbdE100700000000000000000000000000000000 + nonce,);
+    }(sender_address: felt, salt: felt) -> (evm_contract_address: felt) {
```
jobez, line 555: I thought I would preserve our back and forth here about potential next steps re: the salt usage here.

> me: just to sanity check my understanding, in the case of create, the evm documentation refer to a nonce. in our codebase we have salt storage variable. 

> these terms are pretty much interchangeable?

> clement: actually I think that the nonce is sender related while our salt is currently global
> actually I think that the nonce is sender related while our salt is currently global

src/utils/utils.cairo
ClementWalter: we are supposed to use natspec https://docs.soliditylang.org/en/v0.8.17/natspec-format.html

could you create an issue for cleaning the codebase and adding a tool like https://github.com/onlydustxyz/kaaper to generate the doc from the comment?

src/kakarot/instructions/system_operations.cairo
ClementWalter: yes, let's keep like that (`salt`)  and create an issue for investigating the usage of a true sender's nonce

src/utils/rlp.cairo
jobez: np! see #457

ClementWalter (approved):

PR #462, opened 2023-01-25
fix: do uint256 comparison for condition validity in jumpi opcode

### LINKED ISSUE #439: bug: require(address != 0) does not work for address == 2**128
# Bug Report

**Current behavior:**

Fix the test marked as `skip` in `test_plain_opcodes::TestRequire`

**Expected behavior:**

Test passes unskipped

**Steps to reproduce:**

Run the test

**Other information:**

I have only tested cases with `_address = 1 << 4*i`, ie one single nibble, and it failed from `2**128`. I don't know the behaviour when `_address` has both `low != 0` & `high != 0`

jobez:
Time spent on this PR: 1.1 days

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

The exec_jumpi opcode uses the low field of the popped skipped condition to do the check if it is 'anything other than zero'

https://github.com/sayajin-labs/kakarot/blob/b5423b73dc8c32ecbd6701c5a566dc3ae3fc495b/src/kakarot/instructions/memory_operations.cairo#L224
This causes the bug because for the case of the input address of 2^128, it creates a uint with a low of zero and a high of 1. The highest range of representation the low field can do is 2^128-1.

Resolves #439 

## What is the new behavior?

Now we do a uint256_le check to have consistent results with the evm.

## Other information

I made a branch that preserves the minimized changes I added to enable debugging only when the execution_at_address entry point. I generated a log file for the passing and failing case and diffed them to better understand what was going on. Might be a useful example/starting point for anyone else needing to debug integration tests: https://github.com/jobez/kakarot/tree/cool_messy_debugging

### DIFF
--- a/tests/integration/solidity_contracts/PlainOpcodes/test_plain_opcodes.py
+++ b/tests/integration/solidity_contracts/PlainOpcodes/test_plain_opcodes.py
@@ -154,7 +154,7 @@ class TestPlainOpcodes:
                     caller_address=addresses[0].starknet_address,
                 )
 
-        @pytest.mark.parametrize("address", [2**127 , 2**128])
+        @pytest.mark.parametrize("address", [2**127, 2**128])
         async def test_should_not_revert_when_address_is_not_zero(
             self, plain_opcodes, addresses, address
         ):
--- a/tests/integration/solidity_contracts/PlainOpcodes/test_plain_opcodes.py
+++ b/tests/integration/solidity_contracts/PlainOpcodes/test_plain_opcodes.py
@@ -154,13 +154,15 @@ class TestPlainOpcodes:
                     caller_address=addresses[0].starknet_address,
                 )
 
-        @pytest.mark.skip("Test fails when address is 2**128 or greater")
-        @pytest.mark.parametrize("address", [2**127, 2**128])
+        @pytest.mark.parametrize("address", [2**127 , 2**128])
         async def test_should_not_revert_when_address_is_not_zero(
             self, plain_opcodes, addresses, address
         ):
+            address_bytes = address.to_bytes(20, byteorder="big")
+            address_hex = Web3.toChecksumAddress(address_bytes)
+
             await plain_opcodes.requireNotZero(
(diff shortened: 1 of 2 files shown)


tests/integration/solidity_contracts/PlainOpcodes/test_plain_opcodes.py
```diff
@@ -154,13 +154,15 @@ async def test_should_revert_when_address_is_zero(
                     caller_address=addresses[0].starknet_address,
                 )
 
-        @pytest.mark.skip("Test fails when address is 2**128 or greater")
         @pytest.mark.parametrize("address", [2**127, 2**128])
         async def test_should_not_revert_when_address_is_not_zero(
             self, plain_opcodes, addresses, address
         ):
+            address_bytes = address.to_bytes(20, byteorder="big")
```
jobez, line 161: change was interpreted as necessary to deal w/the following error:

```
>           raise ValidationError(message)
E           web3.exceptions.ValidationError: 
E           Could not identify the intended function with name `requireNotZero`, positional argument(s) of type `(<class 'str'>,)` and keyword argument(s) of type `{}`.
E           Found 1 function(s) with the name `requireNotZero`: ['requireNotZero(address)']
E           Function invocation failed due to no matching argument types.

```

jobez (commented):
I was thinking that maybe it would be worth it to 'smoke out' any uint256<>felt bugs by 
  - adding one issue or maybe a series of issues (particularly for logic that is calling uint256.low anywhere) to
      - add parameterized cases in the unit tests for the case of a value  being 2^128

ClementWalter (changes requested):
awesome thanks.

can you also update the uniswap test that helped us figuring out this bug?
see https://github.com/sayajin-labs/kakarot/blob/fe03ff003c0f04b96a02572899ca6c8a8e613e0a/tests/integration/solidity_contracts/UniswapV2/test_uniswap_v2_factory.py#L12

Issue #452, opened 2023-01-22
bug: CALLER returns same address in sub_context


seshanthS:
# Bug Report

**Kakarot version:**

**Current behavior:**
`CALLER` (msg.sender) returns the EVM address of transaction initiator in sub-contexts.

EOA -> Contract_A -> Contract_B
`msg.sender` in Contract_B returns EOA

**Expected behavior:**
I. `CALLER` should return EVM address of transaction initiator when root_context is the calling_context.

II. `CALLER` should return the calling context's `evm_contract_address` when, 
1. The calling context is **NOT** the root context
2. The sub_context is **NOT** created as a result of `DELEGATECALL`

**Steps to reproduce:**
See test `PlainOpcodes.originAndSender`.

**Related code:**

[environmental_information.cairo#L165](https://github.com/sayajin-labs/kakarot/blob/main/src/kakarot/instructions/environmental_information.cairo#L165)

**Other information:**

`ctx.calling_context.evm_contract_address` - To get the calling context's evm_address

mradolux:
Hello ! Can I take this one please ?

ClementWalter:
@mradolux sure!
can you plz also fix https://github.com/sayajin-labs/kakarot/pull/449 before actually starting to work on this?

mradolux:
yes of course !

PR #471, opened 2023-01-28
UniswapV2Factory full test

### LINKED ISSUE #467: feat: Fix UniswapV2Factory tests
# Feature Request

When deploying the `UniswapV2Factory` we face a couple of bugs that have been addressed in dedicated PR. We are now in a position to pass these tests.

The reference file for this test is https://github.com/Uniswap/v2-core/blob/master/test/UniswapV2Factory.spec.ts
The corresponding file in kakarot is https://github.com/sayajin-labs/kakarot/blob/main/tests/integration/solidity_contracts/UniswapV2/test_uniswap_v2_factory.py

- set the `TEST_ADDRESSES` to what they are in the original file 

```
const TEST_ADDRESSES: [string, string] = [
  '0x1000000000000000000000000000000000000000',
  '0x2000000000000000000000000000000000000000'
]
```

Unskip `TestCreatePair` and make it pass!

You'll also need to 

- move the `deployer` fixture in `tests/fixtures/2_kakarot.py`
- update `get_solidity_contract` fixture (see `tests/integration/conftest.py`):
    - add the `deployer` fixture in the args
    - make the `starknet_contract_address=None`
    - make the `tx=None`
    - if `starknet_contract_address is None` get the address from the deployer

ClementWalter:
Time spent on this PR: 0.5

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Token addresses are random instead the same one as in the original test.
Test using `createPair` is skipped

Resolves #467 #452 

## What is the new behavior?

All tests pass as in the original file

### DIFF
--- a/src/kakarot/instructions/environmental_information.cairo
+++ b/src/kakarot/instructions/environmental_information.cairo
@@ -163,7 +163,9 @@ namespace EnvironmentalInformation {
         alloc_locals;
         // Get caller address.
         let (current_address) = get_caller_address();
-        let (evm_address) = IAccount.get_eth_address(current_address);
+        let (current_evm_address) = IAccount.get_eth_address(current_address);
+        let is_root = ExecutionContext.is_root(ctx);
+        let evm_address = (1 - is_root) * ctx.calling_context.evm_contract_address + is_root * current_evm_address;
         let evm_address_uint256 = Helpers.to_uint256(evm_address);
         let stack: model.Stack* = Stack.push(self=ctx.stack, element=evm_address_uint256);
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -122,6 +122,7 @@ markers = [
     "SolmateERC20",
     "SolmateERC721",
     "UniswapV2ERC20",
+    "UniswapV2Factory",
     "Utils",
 ]
 
--- a/src/kakarot/execution_context.cairo
+++ b/src/kakarot/execution_context.cairo
@@ -223,7 +223,7 @@ namespace ExecutionContext {
     // @param self The pointer to the execution context.
     // @return TRUE if the execution context is root, FALSE otherwise.
     func is_root(self: model.ExecutionContext*) -> felt {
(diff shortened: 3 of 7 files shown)


src/kakarot/instructions/environmental_information.cairo
```diff
@@ -163,7 +163,9 @@ namespace EnvironmentalInformation {
         alloc_locals;
         // Get caller address.
         let (current_address) = get_caller_address();
-        let (evm_address) = IAccount.get_eth_address(current_address);
+        let (current_evm_address) = IAccount.get_eth_address(current_address);
+        let is_root = ExecutionContext.is_root(ctx);
+        let evm_address = (1 - is_root) * ctx.calling_context.evm_contract_address + is_root * current_evm_address;
```
Eikix, line 168: nice catch!

Eikix (approved):
lgtm

Eikix (changes requested):
Nvm I didn't notice tests were failing

src/kakarot/instructions/environmental_information.cairo
ClementWalter: tribute to @seshanthS actually

ClementWalter:
> Nvm I didn't notice tests were failing

It took me quite a lot of time to understand what was the issue. Actually the `ExecutionContext.is_root` and `ExecutionContext.is_leaf` (note: I wrote them myself two months ago) were not checking that a given _true_ context was root or leaf ; it was working only on the dummy `empty_ctx`.

These methods are not tested (!) and actually `is_root` was used only once with 

```
is_parent_root = ExecutionContext.is_root(ctx.calling_context);
```

which was quite misleading imao against past me. Consequently I've updated these helpers to be usable on the actual current execution context. For now, I've just considered that checking against an empty `evm_contract_address` was enough, wdyt?

Eikix:
> 

Are we still maintaining kakarot.execute?

It looked to me as though we were slowly deprecating kakarot.execute in favor of kakarot.execute_at_address

PR #475, opened 2023-02-02
feat: add modexp precomile MVP - only for uint256 numbers

### LINKED ISSUE #361: feat: Precompile 0x05 modexp
# Feature Request

feat: Explore / propose strategy (if applicable, MVP implementation) for EVM Precompiled contract -> 0x05 modexp

https://www.evm.codes/precompiled

dragan2234:
ModExp MVP precompile implementation. MVP is missing support for bigint since there is no bigint or arbitrary size and arithmetics implementation in cairo 0.10.3. Both integration and unit tests are passing.

Time spent on this PR: 7 days

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Resolves #361 

## What is the new behavior?

- Calculates modular exponentiation via algorithm that requires log_2(exponent) loops.
- In cairo this requires ~500k steps for 32-bit numbers
- It's not possible to do this in the hint(if it could get whitelisted) since checking the result after the hint is too heavy to compute


src/utils/utils.cairo
```diff
@@ -812,4 +819,34 @@ namespace Helpers {
...
+
+        let (quotient, remainder) = uint256_unsigned_div_rem(y,Uint256(low=2, high=0));
+        let (is_equal_to_one) = uint256_eq(remainder,Uint256(low=1, high=0));
+        if ((is_equal_to_one) == 0) {
+            let (x_res_quotient,  x_res_quotient_high,  x_res_remainder) = uint256_mul_div_mod(x,x,p);
+            return uint256_expmod_recursive_call(x=x_res_remainder,y=quotient,res=res,p=p);
+        } else {
+            let (x_res_res_quotient, x_res_res_quotient_high, x_res_res_remainder) = uint256_mul_div_mod(res,x,p);
+            let (x_res_quotient,  x_res_quotient_high,  x_res_remainder) = uint256_mul_div_mod(x,x,p);
+            return uint256_expmod_recursive_call(x=x_res_remainder,y=quotient,res=x_res_res_remainder,p=p);
```
ClementWalter, line 849: can't we use the `if/else` only for setting the first param `a` of the `uint256_mul_div_mod` like

```
if (is_equal_to_one == 0) {
            tempvar a = x;
        } else {
            tempvar a = res;
}
let (x_res_quotient,  x_res_quotient_high,  x_res_remainder) = uint256_mul_div_mod(a,x,p);
return uint256_expmod_recursive_call(x=x_res_remainder,y=quotient,res=x_res_res_remainder,p=p);
```

tests/unit/src/kakarot/precompiles/test_modexp.cairo
```diff
@@ -0,0 +1,41 @@
...
+from kakarot.memory import Memory
+from kakarot.constants import Constants
+from kakarot.stack import Stack
+from kakarot.execution_context import ExecutionContext
+from kakarot.instructions.memory_operations import MemoryOperations
+from kakarot.instructions.system_operations import SystemOperations, CallHelper, CreateHelper
+from tests.unit.helpers.helpers import TestHelpers
+
+
+const MOD_EXP_BYTES_LEN = 32;
```
ClementWalter, line 24: why re defining this?

ClementWalter (changes requested):

tests/unit/src/kakarot/precompiles/test_modexp.py
```diff
@@ -0,0 +1,43 @@
...
+        m_size_bytes = m_size.to_bytes(32, "big")
+        m_bytes = m.to_bytes(m_size, "big")
+
+        bytes_array = list(
+            b_size_bytes + e_size_bytes + m_size_bytes + b_bytes + e_bytes + m_bytes
+        )
+        expected_result = pow(b,e,m)
+        cairo_result = (await modexp.test__modexp_impl(bytes_array).call()).result[0]
+        cairo_uint256 = uint256_to_int(cairo_result.low,cairo_result.high)
+        assert expected_result == cairo_uint256
```
ClementWalter, line 43: naming is confusing, it seems like it's more

```
cairo_uint256 = (await modexp.test__modexp_impl(bytes_array).call()).result[0]
cairo_result = uint256_to_int(cairo_uint256.low, cairo_uint256.high)
```

dragan2234:
added implementation for gast cost

Eikix (commented):
What is the status of this PR after the TLV hacker house?
@ClementWalter @dragan2234?

dragan2234:
I have a few more updates to push regarding gas cost calculation. Then @ClementWalter can review again and we can merge. Hopefully, I'll push the updates today/tomorrow, I'll ping you when it's ready. Also, we should consider adding benchmarks for the precompiles(if there aren't already some). But first I'll finish the gas calculation. cc @Eikix

Eikix:
Awesome! Benchmarks can be done in a different PR yeah:)

dragan2234:
I just realized that this explanation is a bit confusing:
```
max_length = max(Bsize, Msize)
words = (max_length + 7) / 8
multiplication_complexity = words**2

iteration_count = 0
if Esize <= 32 and exponent == 0: iteration_count = 0
elif Esize <= 32: iteration_count = exponent.bit_length() - 1
elif Esize > 32: iteration_count = (8 * (Esize - 32)) + ((exponent & (2**256 - 1)).bit_length() - 1)
calculate_iteration_count = max(iteration_count, 1)

static_gas = 0
dynamic_gas = max(200, multiplication_complexity * iteration_count / 3)
```
 `ESize` in this explanation refers to length of bits in exponent and not the byte size of the exponent (which is semantic of `ESize` for the modexp calcuation). So I'll need to refactor the gas calcuation.
 
 The correct explanation is given here: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-198.md
 
 Generally speaking, I need to investigate more on geth's and nethermind's implementation of this calcuation.

dragan2234:
just for the reference, gas cost calculation is defined in EIP-198 but redefined in EIP-2565: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-2565.md

There is some confusion and bugs on https://www.evm.codes/precompiled website, but i think i finally know what should i do here.

dragan2234:
> ESize in this explanation refers to the length of bits in the exponent and not the byte size of the exponent (which is semantic of ESize for the modexp calcuation). So I'll need to refactor the gas calculation.

I was wrong about this. So much confusion about this calculation, but I'm getting there slowly. Another useful link: https://ethereum.github.io/execution-specs/autoapi/ethereum/berlin/vm/precompiled_contracts/modexp/index.html

src/utils/utils.cairo
```diff
@@ -16,11 +16,18 @@ from starkware.cairo.common.memcpy import memcpy
...
     uint256_check,
+    uint256_mul_div_mod,
+    uint256_mul,
     uint256_eq,
+    uint256_lt,
     uint256_shr,
+    uint256_sub,
+    uint256_unsigned_div_rem,
 )
+
```
ClementWalter, line 30: unused imports ?

src/utils/modexp/modexp_utils.cairo
```diff
@@ -0,0 +1,452 @@
...
+        let (quotient, remainder) = uint256_unsigned_div_rem(r_res, p);
+        return (remainder=remainder);
+    }
+
+    func uint256_expmod_recursive_call{range_check_ptr: felt}(
+        x: Uint256, y: Uint256, res: Uint256, p: Uint256
+    ) -> (r_x: Uint256, r_y: Uint256, r_res: Uint256) {
+        alloc_locals;
+        let (is_greater_than_zero) = uint256_lt(Uint256(low=0, high=0), y);
+        if ((is_greater_than_zero) == 0) {
```
ClementWalter, line 123: extra parenthesis

ClementWalter: use FALSE instead plz

ClementWalter (changes requested):
as discussed in DM, plz polish the precompile only, add a test in `test_cases.py` and remove all commented or not ready for merge code
thx!

src/utils/modexp/modexp_utils.cairo
```diff
@@ -0,0 +1,452 @@
...
+
+// @title ModExpHelper Functions
+// @notice This file contains a selection of helper functions for modular exponentiation and gas cost calculation.
+// @author @dragan2234
+// @custom:namespace ModExpHelpers
+namespace ModExpHelpers {
+    const GAS_COST_MOD_EXP = 200;
+
+    func calculate_modexp_gas{range_check_ptr: felt, bitwise_ptr: BitwiseBuiltin*}(
+        b_size: Uint256, e_size: Uint256, m_size: Uint256, b: Uint256, e: Uint256, m: Uint256
```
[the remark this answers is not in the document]
dragan2234: @ClementWalter you are right about the `b`,`e` and `m`. I just checked nethermind's implementation of this precompile and they are representing `b` `e` and `m` as `bytes[]` and using `gmp_lib` for arthimetics. This would maybe require different algorithm for modular exponentiation.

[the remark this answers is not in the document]
dragan2234: @ClementWalter Thanks for pointing this out, now i realized the current implementation of ModExp in this PR only supports uint256, but actually bigint or `bytes` or `felt*` is needed. So this implementation is not correct. And there is no bigint library or implementation in cairo. There is this repo: https://github.com/NethermindEth/research-basic-Cairo-operations-big-integers
and this file: https://github.com/starkware-libs/cairo-lang/blob/master/src/starkware/cairo/common/cairo_secp/bigint.cairo

which are both not bigints with arbitrary size. What's interesting is that the modexp with big integers is needed for example in RSA verification as stated on EIP-198: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-198.md

There is bigint library in rust for example: https://docs.rs/num-bigint/0.2.3/num_bigint/struct.BigUint.html


At least gas cost calcuation is done correctly as it operates with uint256 numbers, only for calcuation of gas cost. But the **bigint** modular exponentiation requires more work, as we would need to think how to abstract bigint of arbitrary size (for example 4096-bit number) with felts. I guess felt* is the way to go but then ModExp algorithm needs to be applied over that type

src/kakarot/precompiles/modexp.cairo
```diff
@@ -0,0 +1,52 @@
...
+        let b_size: Uint256 = Helpers.bytes32_to_uint256(input);
+        let e_size: Uint256 = Helpers.bytes32_to_uint256(input + MOD_EXP_BYTES_LEN);
+        let m_size: Uint256 = Helpers.bytes32_to_uint256(input + MOD_EXP_BYTES_LEN * 2);
+        let b: Uint256 = Helpers.bytes_i_to_uint256(input + MOD_EXP_BYTES_LEN * 3, b_size.low);
+        let e: Uint256 = Helpers.bytes_i_to_uint256(
+            input + MOD_EXP_BYTES_LEN * 3 + b_size.low, e_size.low
+        );
+        let m: Uint256 = Helpers.bytes_i_to_uint256(
+            input + MOD_EXP_BYTES_LEN * 3 + b_size.low + e_size.low, m_size.low
+        );
```
ClementWalter, line 43: how can you make sure that `.low` is enough? in other words that all numbers `_n_ < 2**128`?

src/utils/modexp/modexp_utils.cairo
```diff
@@ -0,0 +1,66 @@
...
+// @notice This file contains a selection of helper functions for modular exponentiation and gas cost calculation.
+// @author @dragan2234
+// @custom:namespace ModExpHelpers
+namespace ModExpHelpers {
+    /// @title Modular exponentiation calculation
+    /// @author dragan2234
+    /// @dev Computes x ** y % p for Uint256 numbers via fast modular exponentiation algorithm.
+    /// Time complexity is log_2(y).
+    /// Loop is implemented via uint256_modexp_recursive_call() function.
+    func uint256_modexp{range_check_ptr: felt}(x: Uint256, y: Uint256, p: Uint256) -> (
```
ClementWalter, line 28: according to you naming convention (namespace is `ModExpHelpers`) and starkware oen (e.g. `uint256_mul_div_mod`) I guess that this should be `uint256_mod_exp`

tests/unit/src/kakarot/precompiles/test_modexp.cairo
```diff
@@ -0,0 +1,36 @@
...
+func test__modexp_impl{
+    syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr, bitwise_ptr: BitwiseBuiltin*
+}(data_len: felt, data: felt*) -> (result: Uint256, gas_cost: felt) {
+    alloc_locals;
+
+    let (output_len, output, gas_used) = PrecompileModExp.run(
+        PrecompileModExp.PRECOMPILE_ADDRESS, data_len, data
+    );
+    let result = Helpers.bytes_i_to_uint256(output, output_len);
+    assert output_len = 32;
```
ClementWalter, line 33: why should the output_len be always 32? shouldn't it be instead `<= 32`?

ClementWalter (changes requested):

dragan2234:
closing this one since cairo doesn't support bigint of arbitrary size and arithmetics over that type. It make sense to write this in cairo 1.0 since the development will start very soon

src/kakarot/precompiles/modexp.cairo
dragan2234: @ClementWalter since this implementation only support uint256, byte size of `uint256` is never more than 32

dragan2234:
@ClementWalter I've reopened the PR since @Eikix thought it's ok to have an MVP implementation in the repo. Both unit and integration tests are passing and I have left comments about missing features in this MVP. Also, I have added gas cost calculation logic from #497 which is calculating correctly for exponent `uint256` . I'll drop the progress(except if the review won't require too much input) on this now because of https://github.com/sayajin-labs/kakarot/issues/361#issuecomment-1428707435 . Let me know your thoughts

src/utils/utils.cairo
dragan2234: in the end i kept them since i added gas cost calcuation again

src/kakarot/precompiles/modexp.cairo
```diff
@@ -0,0 +1,64 @@
...
+
+// Internal dependencies
+from utils.utils import Helpers
+from utils.modexp.modexp_utils import ModExpHelpers
+
+// @title ModExpMVP Precompile related functions.
+// @notice It is an MVP implementation since it only supports uint256 numbers with m_size<=16 and not bigint which requires bigint library in cairo 0.10.
+// @author @dragan2234
+// @custom:namespace PrecompileModExpMVP
+namespace PrecompileModExpMVP {
```
ClementWalter, line 19: prefer a meaningful naming, ie `MVP` doesn't make any obvious sense here, maybe `Uint256`?

src/utils/modexp/modexp_utils.cairo
```diff
@@ -0,0 +1,440 @@
...
+    uint256_lt,
+    uint256_sub,
+    uint256_unsigned_div_rem,
+    uint256_add,
+)
+from starkware.cairo.common.bitwise import bitwise_and
+from starkware.cairo.common.registers import get_label_location
+from starkware.cairo.common.bool import TRUE, FALSE
+
+// @title ModExpHelperMVP Functions
```
ClementWalter, line 16: remove the MVP, see before

src/utils/utils.cairo
ClementWalter: but they are not used in the file

ClementWalter (changes requested):

src/utils/utils.cairo
dragan2234: correct, those are for `modexp.utils`, will remove

dragan2234:
@ClementWalter renamed `MVP` to `Uint256`. Run tests again, works fine

ClementWalter (approved):

PR #478, opened 2023-02-02
feat: support legacy transaction in EOA

Flydexo:
Added support for legacy transactions, see reference here:
https://github.com/ethereum/EIPs/blob/master/EIPS/eip-155.md
https://ethereum-magicians.org/t/eip-2718-typed-transaction-envelope/4355

Time spent on this PR: 3 days

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Legacy Transactions are refused by default

## What is the new behavior?

Legacy transactions are handled, verified.

### DIFF
--- a/src/kakarot/accounts/eoa/aa/library.cairo
+++ b/src/kakarot/accounts/eoa/aa/library.cairo
@@ -1,6 +1,7 @@
 %lang starknet
 
 from utils.utils import Helpers
+from kakarot.constants import Constants
 from starkware.cairo.common.cairo_builtins import HashBuiltin, SignatureBuiltin, BitwiseBuiltin
 from starkware.cairo.common.cairo_secp.signature import verify_eth_signature_uint256
 from starkware.starknet.common.syscalls import get_tx_info
@@ -10,6 +11,7 @@ from starkware.cairo.common.math import split_felt
 from starkware.cairo.common.cairo_keccak.keccak import keccak, finalize_keccak, keccak_bigend
 from utils.rlp import RLP
 from starkware.cairo.common.math_cmp import is_le
+from starkware.cairo.common.registers import get_label_location
 
 namespace ExternallyOwnedAccount {
     // Constants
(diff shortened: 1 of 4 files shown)


src/kakarot/accounts/eoa/aa/library.cairo
```diff
@@ -165,15 +207,38 @@ namespace ExternallyOwnedAccount {
...
+        tx_type: felt,
+        keccak_ptr: felt*,
+        keccak_ptr_start: felt*,
+    ) -> (is_valid: felt) {
         alloc_locals;
-        let (keccak_ptr: felt*) = alloc();
-        local keccak_ptr_start: felt* = keccak_ptr;
-        with keccak_ptr {
-            verify_eth_signature_uint256(msg_hash=msg_hash, r=r, s=s, v=v, eth_address=eth_address);
+        if (tx_type == 2) {
```
ClementWalter, line 219: so `type = 2` introduces only a shift (3) in the indexes `V_IDX`, `R_IDX` and `S_IDX`. We could better just compute the correct indices and avoid the `if/else` and duplication.

```
let shift = tx_type / 2 * 3
let v_idx = V_IDX - shift
...
```

TotalPizza:
In general there are some more things we need to validate besides the signature:
Correct nonce, gasLimit is withing valid range, and a few more.
But it's probably best we do this with the TLV-Hackathon pr and not this one.

Flydexo:
> In general there are some more things we need to validate besides the signature: Correct nonce, gasLimit is withing valid range, and a few more. But it's probably best we do this with the TLV-Hackathon pr and not this one.

Yes, agree was starting for gas but the testing library doesn't allow spending gas on tx

Eikix (changes requested):
What is the status of this PR after the TLV hacker house? 
@Flydexo @jobez @TotalPizza did anyone "fork" Flydexo's branch to complete this PR? Or is it still "relevant"?

jobez:
> What is the status of this PR after the TLV hacker house? @Flydexo @jobez @TotalPizza did anyone "fork" Flydexo's branch to complete this PR? Or is it still "relevant"?

I believe it is still relevant, as it extends the fork we were working with to add legacy support.

PR #489, opened 2023-02-04
Fix/revert should not stop execution in sub context

### LINKED ISSUE #432: bug: REVERT should not stop execution in sub context
# Bug Report

**Current behavior:**

The `REVERT` opcode reverts the whole starknet transaction.

**Expected behavior:**

As per the doc:

> Stop the current context execution, revert the state changes (see [STATICCALL](https://www.evm.codes/#FA) for a list of state changing opcodes) and return the unused gas to the caller. It also reverts the gas refund to its value before the current context. If the execution is stopped with REVERT, the value 0 is put on the stack of the calling context, which continues to execute normally. The [return data](https://www.evm.codes/about) of the calling context is set as the given chunk of memory of this context.

**Related code:**

`exec_revert` opcode.

**Other information:**

The list of opcodes modifying the state is:  CREATE, CREATE2, LOG0, LOG1, LOG2, LOG3, LOG4, SSTORE, SELFDESTRUCT.

A possible strategy could be:
- add in the `ExecutionContext` keys for storing `LOG_i` (events) to be emitted and emit only when context is stopped
- add in the `ExecutionContext` a `dict` of all updated storage and revert them if ctx fails.
- add in the `ExecutionContext` a `list` of addresses created with CREATE/CREATE2 and use the same logic as `exec_selfdestruct` if ctx fails.

jobez:
Time spent on this PR:

## Pull request type

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes

Resolves #432 

## What is the new behavior?

-
-
-


ClementWalter:
@jobez is this ready for review?

jobez:
> @jobez is this ready for review?

Still need to do handle the logic between the calling context getting the revert issue/status from the subcontext. After that I will clean up everything and ping for a request 👍

src/kakarot/instructions/system_operations.cairo
```diff
@@ -924,6 +1051,12 @@ namespace SelfDestructHelper {
             sub_context=ctx.sub_context,
             destroy_contracts_len=0,
             destroy_contracts=empty_destroy_contracts,
+            events_len=0,
+            events=cast(0, model.Event*),
```
jobez, line 1055: Probably should just get rid of all the casts because it can cause unexpected memory bugs.

src/kakarot/library.cairo
```diff
@@ -125,6 +126,14 @@ namespace Kakarot {
 
         // Start execution
         let ctx = EVMInstructions.run(ctx);
+        let is_reverted = ExecutionContext.is_reverted(self=ctx);
```
jobez, line 129: I am throwing at the end of 'execute_at_address' for now, but probably with discussion from rpc people about maybe we can propogate the error information to the rpc and let them decide how to convey.

tests/integration/solidity_contracts/PlainOpcodes/PlainOpcodes.sol
```diff
@@ -127,4 +128,52 @@ contract PlainOpcodes {
             extcodecopy(target, add(extcode, 0x20), offset, size)
         }
     }
+
```
jobez, line 131: Currently putting exception handling tests in plainopcodes. I think it makes sense to refactor into a whole ExceptionHandling directory in integration testing. Thoughts?

tests/unit/src/kakarot/instructions/test_system_operations.cairo
```diff
@@ -115,10 +115,16 @@ func test__exec_revert{
     let ctx: model.ExecutionContext* = ExecutionContext.update_stack(ctx, stack);
 
     // When
-    SystemOperations.exec_revert(ctx);
+    let ctx = SystemOperations.exec_revert(ctx);
+    let is_reverted = ExecutionContext.is_reverted(ctx);
 
     // Then
-    // TODO: update test when revert does not break the execution
+    assert is_reverted = 1;
```
jobez, line 122: as in `execute_at_address` we are still conveying the exception via a kakarot throw, for now.

src/kakarot/instructions/system_operations.cairo
```diff
@@ -509,32 +559,84 @@ namespace CallHelper {
...
-        // ret_offset, see prepare_args
-        let memory = Memory.store_n(
-            ctx.memory,
-            ctx.sub_context.return_data_len,
-            ctx.sub_context.return_data,
-            [ctx.sub_context.return_data - 1],
-        );
-        let ctx = ExecutionContext.update_memory(ctx, memory);
+        if (sub_context_reverted != 0) {
+            SelfDestructHelper._finalize_loop(
```
jobez, line 567: Reusing the loop in the SelfDestructHelper. Need to refactor a name to generalize usecases across concerns.

jobez (commented):
Ok, I wanted to surface what I had so far to get general flow feedback as I continue to refine.

What is still left is adding more tests to ensure we are propagating the revert condition following the evm semantics and are doing the 'reverting' logic of undoing writes and deleting create contracts at the right time.

 There is one skipped test currently that marks that.

src/kakarot/library.cairo
jobez: @Eikix thoughts of how exceptions should be relayed to the RPC?

src/kakarot/instructions/system_operations.cairo
```diff
@@ -832,6 +954,8 @@ namespace CreateHelper {
     }(ctx: model.ExecutionContext*) -> model.ExecutionContext* {
         alloc_locals;
 
+        let is_reverted: felt = ExecutionContext.is_reverted(self=ctx);
```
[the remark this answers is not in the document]
ClementWalter: I would try not to mix REVERT ops and other parts of the EVM

ClementWalter (commented):
pushing this first line of thoughts !

src/kakarot/instructions/memory_operations.cairo
```diff
@@ -375,14 +384,47 @@ namespace MemoryOperations {
...
+
             IContractAccount.write_storage(
                 contract_address=starknet_contract_address, key=key, value=value
             );
         }
 
-        // Increment gas used.
-        let ctx = ExecutionContext.increment_gas_used(ctx, GAS_COST_SSTORE);
-        return ctx;
+        let (__fp__, _) = get_fp_and_pc();
```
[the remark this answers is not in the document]
jobez: I saw it in the amm example in the cairo 0.10 repo that had an example of using dicts. Will check if it was erroneously duplicated!

src/kakarot/library.cairo
```diff
@@ -125,7 +126,7 @@ namespace Kakarot {
 
         // Start execution
         let ctx = EVMInstructions.run(ctx);
-
+        ExecutionContext.maybe_throw_revert(ctx);
```
jobez, line 129: This is where I refactored the logic that takes the revert bytes and gets a short string.

src/kakarot/instructions/system_operations.cairo
```diff
@@ -204,32 +214,17 @@ namespace SystemOperations {
         // 1 - offset: byte offset in the memory in bytes
         let (stack, popped) = Stack.pop_n(self=stack, n=2);
 
-        let size = popped[0];
-        let offset = popped[1];
+        let offset = popped[0];
```
jobez, line 217: This change was the crux of the issues I was seeing with revert reason propagating from sub to calling context.

jobez (commented):
Now the first round of tests pass that assert errors propagating across contexts along with the revert messages. 

Now will look to add another round that make sure reversal of writes/addresses is correct and that emits are only emitted in a successful and completed context.

src/kakarot/instructions.cairo
```diff
@@ -632,8 +635,13 @@ namespace EVMInstructions {
                     contract_address=ctx.starknet_contract_address
                 );
                 if (bytecode_len == 0) {
-                    let ctx = CreateHelper.finalize_calling_context(ctx);
-                    return run(ctx=ctx);
+                    if (is_reverted != 0) {
+                        // TODO need to reason about the 'when' a create contract gets deleted
```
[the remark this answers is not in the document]
jobez: Just to make sure I understand, you're saying something like:

```
        // Terminate execution
        if (stopped != FALSE) {
            // TODO should only emit if stopped and not reverted
            let ctx = LoggingHelper.finalize(ctx);
            if (is_root != FALSE) {
                if (ctx.destroy_contracts_len != 0) {
                    let ctx = SelfDestructHelper.finalize(ctx);
                    return ctx;
                }
                return ctx;
            } else {
                let is_precompile = Precompiles.is_precompile(address=ctx.evm_contract_address);
                if (is_precompile != FALSE) {
                    let ctx = CallHelper.finalize_calling_context(ctx);
                    return run(ctx=ctx);
                }
                
                if (is_reverted != 0) {
                    // do state management stuff here
                }
        
                let (bytecode_len) = IAccount.bytecode_len(
                    contract_address=ctx.starknet_contract_address
                );
                if (bytecode_len == 0) {
                        let ctx = CreateHelper.finalize_calling_context(ctx);
                        return run(ctx=ctx);                
                } else {
                    let ctx = CallHelper.finalize_calling_context(ctx);
                    return run(ctx=ctx);
                }
            }
        }

```

src/kakarot/instructions/memory_operations.cairo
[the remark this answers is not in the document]
jobez: commented out and seems to run fine.

here is the relevant section of code from the amm I am talking about:
https://github.com/starkware-libs/cairo-lang/blob/master/src/demo/amm_demo/amm.cairo#L51-L60

I think the `get_fp_and_pc` call is necessary to get the pointer of the struct with the ampersand (&) syntax.

jobez (commented):
I began refactoring by consolidating most of the revert logic in execution context.

Two tasks left (in my mind) that would take it out of draft:

- making sure that tests for events emitted are adequate / passing
- looking into how gas is handled in revert  (or maybe we can skip this task, will leave that up for discussion)

tests/integration/solidity_contracts/PlainOpcodes/test_plain_opcodes.py
```diff
@@ -167,6 +168,82 @@ async def test_should_not_revert_when_address_is_not_zero(
...
+
+            # we ignore the first 4 bytes because it is the function selector
+            # see conventions for the structure of return payloads on reverting cases
+            # https://docs.soliditylang.org/en/latest/control-structures.html#revert
+            assert "Revert reason" == Web3().codec.decode(["string"], data[4:])[0]
+            assert success == False
+            assert await counter.count() == 0
+            # TODO: is this just asserting that there were no events firing from a contract,
+            # or that there were no events fired from all touched contracts in an execution?
+            assert caller.events._events == []
```
ClementWalter, line 195: actually this `_events` key is from Web3 but not updated by the kakarot wrapper. The wrap_kakarot only `setattr` matching emitted events of the given contract
see https://github.com/sayajin-labs/kakarot/blob/91999728588bfa9efb047b16c2972dfa4e83838b/tests/integration/helpers/wrap_kakarot.py#L95

in other words, currentl, events emitted in subcontexts would be filtered out in the wrapper so there is no way to assert here something about them.

Because it's difficult from the wrapper to know which sub context could have emitted, I think that an option would to add the raw `log_receipts` from the tx and a method to filter these receipts for any given target even abi

tests/unit/helpers/helpers.cairo
```diff
@@ -40,6 +41,48 @@ namespace TestHelpers {
         return ctx;
     }
 
+    func init_context_at_stark_address{
```
ClementWalter, line 44: could we stick to full names (is `starknet_address` here)?

tests/integration/solidity_contracts/PlainOpcodes/PlainOpcodes.sol
```diff
@@ -127,4 +128,82 @@ contract PlainOpcodes {
...
+        } catch Error(string memory errorMessage) { 
+            // Ensure that the created child contract in parent is deleted. 
+            require(address(parent.child()) == address(0), "Parent is deleted"); 
+        
+            // Revert with the error message from the subcontext.
+            revert(errorMessage);
+        } 
+    }    
+
+    function testCallingContextShouldCatchRevertFromSubContextWithReversedWrites() public {
```
ClementWalter, line 183: I don't understand the purpose of this test

tests/integration/solidity_contracts/PlainOpcodes/PlainOpcodes.sol
```diff
@@ -127,4 +128,82 @@ contract PlainOpcodes {
...
+        } catch { 
+            // Ensure that the created child contract in parent is deleted. 
+            //  
+            require(address(parent.child()) == address(0), "Parent is deleted"); 
+        } 
+    }   
+
+    function testCallingContextShouldCatchRevertFromSubContextAndPropogateRevertMessage() public {
+        Parent parent = new Parent();
+
```
ClementWalter, line 170: these two tests seem redundant after `testShouldRevertViaCall`

tests/integration/solidity_contracts/PlainOpcodes/PlainOpcodes.sol
```diff
@@ -127,4 +128,82 @@ contract PlainOpcodes {
...
+    function testCallingContextShouldPropogateRevertFromSubContextOnCreate() public {
+        ContractRevertsOnConstruction doomedContract = new ContractRevertsOnConstruction();
+
+    }    
+    
+    function testShouldRevertViaCall() external returns (address address_, bool triggerRevert ) {
+        Parent parent = new Parent();
+
+        (bool success, ) = address(parent).call(abi.encodeWithSignature("triggerRevert()"));
```
ClementWalter, line 150: Does the `returnData` return the revert reason in case of revert?
You could then add here your 

```
assert(address(parent.child()) == address(0)); 
```

tests/integration/solidity_contracts/PlainOpcodes/PlainOpcodes.t.sol
```diff
@@ -48,4 +49,53 @@ contract PlainOpcodesTest is Test {
...
+        parent.triggerRevert();
+
+    }    
+    
+    function testShouldRevertViaCall() public {
+        Parent parent = new Parent();
+        // child contract is created in the same execution context as a revert       
+        (bool success, ) = address(parent).call(abi.encodeWithSignature("triggerRevert()"));
+
+        require(!success, "trigger revert... should revert");
```
ClementWalter, line 70: could yse `assert` instead of adding unnecessary revert reason `"trigger revert... should revert"`

ClementWalter (changes requested):
This is quite a big piece of work!
I think that the solidity tests should be skimmed to keep only what's strictly necessary:

- let's create a `RevertTestCases.sol` file with two contracts in:
    - a contract that raises in the constructor
    - a contract with a function that deploys a `Counter`, updates a storage, emit and raises
- let's keep `PlainOpcode` as the main sol contract with two functions, one for each

wdyt?

tests/integration/solidity_contracts/PlainOpcodes/PlainOpcodes.sol
jobez: I was basically stress testing the reverting writes logic--that it is restoring the value to the pre-context state. You feel its overkill?

tests/integration/solidity_contracts/PlainOpcodes/PlainOpcodes.sol
jobez: I pushed in PlainOpcodes.t.sol what it would look like to assert the error message in soliditiy. trying to run the same solidity code in kkrt resulted in a much longer test time (i never bothered waiting for it to finish once it got to the 5x time difference) so I opted to just return the returndata and test for the revert in python

jobez:
I am having an issue finding the comment you made about what is necessary to test for emitted events per transaction. Is it cool if we take that on in another issue?

src/kakarot/instructions.cairo
```diff
@@ -21,7 +21,7 @@ from kakarot.instructions.comparison_operations import ComparisonOperations
 from kakarot.instructions.duplication_operations import DuplicationOperations
 from kakarot.instructions.environmental_information import EnvironmentalInformation
 from kakarot.instructions.exchange_operations import ExchangeOperations
-from kakarot.instructions.logging_operations import LoggingOperations
+from kakarot.instructions.logging_operations import LoggingOperations, LoggingHelper
```
ClementWalter, line 24: unused import

tests/integration/solidity_contracts/PlainOpcodes/PlainOpcodes.sol
```diff
@@ -127,4 +128,22 @@ contract PlainOpcodes {
             extcodecopy(target, add(extcode, 0x20), offset, size)
         }
     }
+
+    function testCallingContextShouldPropogateRevertFromSubContextOnCreate() public {
+        ContractRevertsOnConstruction doomedContract = new ContractRevertsOnConstruction();
+        doomedContract.value();
```
ClementWalter, line 134: if it's reverting on construction, this line is never reached right?

tests/integration/solidity_contracts/PlainOpcodes/PlainOpcodes.sol
```diff
@@ -127,4 +128,22 @@ contract PlainOpcodes {
...
+    }    
+    
+    function testShouldRevertViaCall() external returns (bytes memory returnData) {
+        ContractRevertsOnMethodCall doomedContract = new ContractRevertsOnMethodCall();
+
+        (bool success, bytes memory returnData1) = address(doomedContract).call(abi.encodeWithSignature("triggerRevert()"));
+
+        assert(!success);
+
```
ClementWalter, line 143: maybe add a `assert(doomedContract.counter.address) = 0` ?

ClementWalter (approved):
looks good. So we still need to add a check for the events, maybe in another PR if you prefer

tests/integration/solidity_contracts/PlainOpcodes/PlainOpcodes.sol
jobez: yeah, but necessary to get rid of a linting error of an used variable

src/kakarot/instructions.cairo
jobez: incredible eye!

jobez:
> looks good. So we still need to add a check for the events, maybe in another PR if you prefer

 are you thinking we expose something in wrap_kakarot or just use the web3 filter stuff in a one-off?

jobez:
> > looks good. So we still need to add a check for the events, maybe in another PR if you prefer
> 
>  are you thinking we expose something in wrap_kakarot or just use the web3 filter stuff in a one-off?

For some reason i am able to see your comment on mobile but not desktop

tests/integration/solidity_contracts/PlainOpcodes/PlainOpcodes.sol
ClementWalter: can't you just write

```
function testCallingContextShouldPropogateRevertFromSubContextOnCreate() public {
    new ContractRevertsOnConstruction();
}
```

ClementWalter (approved):
lfg!

PR #501, opened 2023-02-14
fix: update get_evm_address to view

Eikix:
Fix


jobez (approved):
lgtm

ClementWalter (approved):
It's ok for me to update this though it shouldn't change anything as there are no differences actually between view and external afaik

PR #502, opened 2023-02-15
Feat/eoa_deploys_contract

### LINKED ISSUE #491: feat: add deploy support in EOA / Kakarot
# Feature Request

Deploy is done in EVM by sending a raw transaction to address "empty string" 
And calldata := bytecode! 

EOA needs to handle this case. 
Idea: by branching ifs on the `to` field inside the transaction (e.g. if to if empty string -> kakarot.deploy, else kakarot.execute_at_address)

TotalPizza:
Time spent on this PR: 0.5d

## Pull request type

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes

Account contracts have to be deployed by interaction with kakarot and can't be deployed via the EOA. 

Resolves #491 

## What is the new behavior?

- Sending a transaction (legacy or default) with the "to" address set to "" will result in a contract deployment instead of a standard transaction.
- An event containing the new contracts EVM address is emitted on contract creation


src/kakarot/accounts/eoa/library.cairo
```diff
@@ -122,18 +124,34 @@ namespace ExternallyOwnedAccount {
...
         let (current_tx_calldata: felt*) = alloc();
-        assert [current_tx_calldata] = destination;
-        assert [current_tx_calldata + 1] = amount;
-        assert [current_tx_calldata + 2] = gas_limit;
-        assert [current_tx_calldata + 3] = payload_len;
-        memcpy(current_tx_calldata + 4, payload, payload_len);
+        local offset;
+        local selector;
+        // If destination is 0, we are deploying a contract
+        if (destination == FALSE) {
```
TotalPizza, line 132: Alternatively I could move the logic into their own functions:
```
if  (destination == FALSE) {
    deploy_contract();
}else{
    execute_contract_bytecode();
}
```
But imo this is still fine.

tests/unit/src/kakarot/accounts/eoa/test_library.py
```diff
@@ -45,17 +45,31 @@ async def test_execute_should_make_all_calls_and_return_concat_results(
...
-                transaction["value"],
-                transaction["gas"],
-                len(
-                    bytes.fromhex(transaction.get("data", "").replace("0x", ""))
-                ),  # calldata_len
-                *list(
-                    bytes.fromhex(transaction.get("data", "").replace("0x", ""))
-                ),  # calldata
-            ]
+            # Execute contract bytecode
```
TotalPizza, line 48: @ClementWalter  kinda hijacked your test here.
I didn't think that this deserved its own one, as we're still only testing whether transactions are decoded and forwarded to kakarot correctly.

TotalPizza:
I could extend this PR to include the new contracts evm address being determined using the EVMs nonce instead of the global kakarot salt.
(But it's probably better as it's own PR, as each account_contract now also requires an individual nonce that it uses when CREATE or CREATE2 is being executed)

src/kakarot/accounts/library.cairo
```diff
@@ -27,6 +27,10 @@ from kakarot.constants import (
 )
 from kakarot.interfaces.interfaces import IAccount
 
+@event
+func evm_contract_deployed(evm_contract_address: felt, starknet_contract_address: felt) {
```
TotalPizza, line 31: @Eikix Is there more info we might need for the rpc/block-explorer?
Maybe the evm address that deployed the contract?

tests/unit/src/kakarot/accounts/eoa/mock_kakarot.cairo
```diff
@@ -13,3 +13,13 @@ func execute_at_address{
 ) {
     return (address, value, gas_limit, calldata_len, calldata);
 }
+
+// @dev mock kakarot contract that returns inputs for execute_at_address
```
ClementWalter, line 17: update plz

tests/utils/constants.py
```diff
@@ -74,4 +74,23 @@
...
         "chainId": CHAIN_ID,
     },
+    {
+        "type": 2,
+        "gas": 100_000,
+        "maxFeePerGas": 2_000_000_000,
+        "maxPriorityFeePerGas": 2_000_000_000,
+        "data": "0x616263646566",
+        "nonce": 34,
+        "to": "",
```
ClementWalter, line 84: is `"to": ""` the same as `"to": ADDRESS_ZERO`?

ClementWalter (approved):
lgtm, just few questions

tests/utils/constants.py
TotalPizza: I remember @Eikix saying during the hackathon that `""` was what they would be sending via the RPC, which is why I used it. But yes, `""` and `ZERO_ADDRESS` both result in `to` being **0** inside the contract.

src/utils/eth_transaction.cairo
```diff
@@ -245,6 +245,11 @@ namespace EthTransaction {
         let (gas_limit, destination, amount, payload_len, payload, tx_hash, v, r, s) = decode(
             tx_data_len, tx_data
         );
+        if (destination == FALSE){
+            with_attr error_message("ExternallyOwnedAccount: account creations are not payable") {
```
Eikix, line 249: I think they are? granted the constructor is payable as well (i think)

src/kakarot/kakarot.cairo
```diff
@@ -167,17 +167,18 @@ func set_native_token{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_chec
...
+// @param gas_limit: maximum amount of gas the constructor can use
 // @param bytes_len: the constructor + contract bytecode lenght
 // @param bytes: the constructor + contract bytecode
 // @return evm_contract_address The evm address that is mapped to the newly deployed starknet contract address
 // @return starknet_contract_address The newly deployed starknet contract address
 @external
 func deploy_contract_account{
     syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr, bitwise_ptr: BitwiseBuiltin*
-}(bytecode_len: felt, bytecode: felt*) -> (
+}(gas_limit: felt, bytecode_len: felt, bytecode: felt*) -> (
```
Eikix, line 178: I don't think we need gas_limit as a param, we decided in the community call to deprecate the use of gas_limit in our logic.
I think a pragmatic approach would be to remove this most recent addition of gas_limit inside the function

tests/utils/constants.py
Eikix: very cool !

src/kakarot/accounts/eoa/library.cairo
Eikix: still fine imo as well

src/kakarot/accounts/library.cairo
Eikix: I'm not sure yet, does emitting an event pose a problem to the app?

I recall @ClementWalter saying it was ruining the event parsing in his tests

Eikix (changes requested):
very cool PR! let's merge it asap and try to deploy via remix

src/utils/eth_transaction.cairo
TotalPizza: Just noticed, you're 100% correct.
Fixing that asap.

src/kakarot/kakarot.cairo
TotalPizza: Okidoki. Removing it.

src/kakarot/accounts/library.cairo
TotalPizza: Oh, I thought it just got removed because it wasn't being used.

ClementWalter: Actually, there is not real problem it was only for events parsing I’m the web3.pu lib.

Eikix: We can leave it there!

Eikix (approved):
lgtm

src/kakarot/kakarot.cairo
```diff
@@ -167,6 +167,7 @@ func set_native_token{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_chec
 }
 
 // @notice Deploy a new contract account and execute constructor
+// @param gas_limit: maximum amount of gas the constructor can use
```
ClementWalter, line 170: this need to be removed then

src/kakarot/library.cairo
```diff
@@ -164,6 +164,7 @@ namespace Kakarot {
     // @notice deploy contract account
     // @dev First deploy a contract_account with no bytecode, then run the calldata as bytecode with the new address,
     //      then set the bytecode with the result of the initial run
+    // @param gas_limit: maximum amount of gas the constructor can use
```
ClementWalter, line 167: remove then plz

ClementWalter (dismissed):

ClementWalter (approved):

src/kakarot/kakarot.cairo
TotalPizza: whooops,
removed!

PR #506, opened 2023-02-16
feat: test compute starknet address

### LINKED ISSUE #477: feat: test compute_starknet_address
# Feature Request

Provide a unit test for `compute_starknet_address` in `accounts.library`

Test should be made in `unit.src.kakarot.accounts`.

Create a `test_accounts.cairo` where you import `compute_starknet_address` from `kakarot.accounts.library` and test it in `test_accounts.py`. You could compare our function returned value to the one returned by starkware python lib.

dragan2234:
Resolves #477 

Time spent on this PR: 1.5 days

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Resolves #<Issue number>

## What is the new behavior?

-
-
-

### DIFF
--- a/tests/integration/test_kakarot.py
+++ b/tests/integration/test_kakarot.py
@@ -62,8 +62,14 @@ class TestKakarot:
                 int(evm_address, 16)
             ).execute()
 
-            kakarot_starknet_address = eoa_deploy_tx.call_info.internal_calls[0].contract_address
+            kakarot_starknet_address = eoa_deploy_tx.call_info.internal_calls[
+                0
+            ].contract_address
 
-            computed_starknet_address = (await kakarot.compute_starknet_address(evm_address=int(evm_address, 16)).call()).result[0]
+            computed_starknet_address = (
+                await kakarot.compute_starknet_address(
+                    evm_address=int(evm_address, 16)
+                ).call()
+            ).result[0]
 
             assert kakarot_starknet_address == computed_starknet_address


dragan2234:
@ClementWalter Assert is failing, our `compute_starknet_address` is giving different number from `kakarot_starknet_address`. I tried debugging [library.cairo](https://github.com/sayajin-labs/kakarot/blob/main/src/kakarot/accounts/library.cairo) file but still no conclusions. So far I realized when we deploy the EOA starknet address is assigned via `create`, then some parameters are passed to `initialize` and there is some `deploy` syscall? I tried playing with `constructor_calldata` since there are some differences in `create` and `compute_starknet_address` but still couldn't debug. If you have any ideas/suggestions let me know.

ClementWalter:
> @ClementWalter Assert is failing, our `compute_starknet_address` is giving different number from `kakarot_starknet_address`. I tried debugging [library.cairo](https://github.com/sayajin-labs/kakarot/blob/main/src/kakarot/accounts/library.cairo) file but still no conclusions. So far I realized when we deploy the EOA starknet address is assigned via `create`, then some parameters are passed to `initialize` and there is some `deploy` syscall? I tried playing with `constructor_calldata` since there are some differences in `create` and `compute_starknet_address` but still couldn't debug. If you have any ideas/suggestions let me know.

you cannot test the computing of the starknet address like that because it depends on the address calling the `compute_starknet_address`, see

```
let (_deployer_address: felt) = get_contract_address();
```

what you want to do instead is to compare what kakarot returns with what should be returned when using the deploy syscall with these exact same params. Actually the Issue described the possible implementation 

> You could compare our function returned value to the one returned by starkware python lib.

dragan2234:
@ClementWalter  So my first intuition was:

In the PR there are 2 starknet addresses:
```
        computed_starknet_address = (await test_starknet_address_computation.test__compute_starknet_address(evm_address=int(evm_address, 16)).call()).result[0]
```
so this is calculated [here](https://github.com/sayajin-labs/kakarot/blob/main/src/kakarot/accounts/library.cairo#L39)

and this:
```
        eoa_deploy_tx = await kakarot.deploy_externally_owned_account(
            int(evm_address, 16)
        ).execute()

        kakarot_starknet_address = eoa_deploy_tx.call_info.internal_calls[0].contract_address
```

which is calculated through [here](https://github.com/sayajin-labs/kakarot/blob/main/src/kakarot/library.cairo#L248) so it's calling [this](https://github.com/sayajin-labs/kakarot/blob/main/src/kakarot/accounts/library.cairo#L79)

> what you want to do instead is to compare what kakarot returns with what should be returned when using the deploy syscall with these exact same params. Actually, the Issue described the possible implementation

And this `create` is calling `deploy_syscall`. And the resulting addresses are different.

At least that's my understanding of this sentence:
```
You could compare our function returned value to the one returned by starkware python lib.
```

so "our function" is the `compute_starknet_address` and `one returned by starkware python lib.` is the `kakarot.deploy_externally_owned_account(
            int(evm_address, 16)
        ).execute()`, which is calling `deploy_syscall` in the `create` function. I couldn't find any other reference to `one returned by starkware python lib.`.
        
So my question is: **am I looking at the correct calls?**

But since `compute_starknet_addreses` depends on:
```
let (_deployer_address: felt) = get_contract_address();
```

**how should I call this then?** 

OR

first call should be the `kakarot.deploy_externally_owned_account(
            int(evm_address, 16)
        ).execute()`
        
And second should be only:
```
        let (account_address) = deploy_syscall(
            _account_proxy_class_hash,
            contract_address_salt=evm_address,
            constructor_calldata_size=0,
            constructor_calldata=constructor_calldata,
            deploy_from_zero=0,
        );
```


which I would put in separate cairo file and call it from python test file. but that's the same actually?


sorry I'm a bit confused with this

hurrikaanig:
Hey @dragan2234,

One simpler way of testing would be importing `deploy_externally_owned_account` from kakarot.cairo in your test file. Then call it with `evm_address` and compare its result to `Accounts.compute_starknet_address(evm_address)`. Since address returned by `deploy_externally_owned_account` is the one returned by the syscall it should be ok.

Also you will need to set `externally_owned_account_class_hash` and `account_proxy_class_hash` in your test for this to work.

dragan2234:
@hurrikaanig thanks for the input.

Tests are passing finally. I think I got confused with this sentence:

> Create a test_accounts.cairo where you import compute_starknet_address from kakarot.accounts.library


If this test_accounts.cairo is deployed by "starknet" object in python, wrong `deployer_address` is assigned in `compute_starknet_address` function.

So in the end I made it via "kakarot: StarknetContract" python object where I called and asserted these 2 functions from that object. That way `externally_owned_account_class_hash` and `account_proxy_class_hash` are already set.

Anyway, let me know your thoughts and if this is correct now

ClementWalter (changes requested):
ok so this is much better indeed!

So few comments to address:
- this is a test regarding the `Kakarot` contract right? so this test should be in `test_kakarot.py` file. This file however is in a misleading directory `tests/integration/bytecode` so just move it upward into `tests/integrations/.` ; leave however the `test_cases` therein and update the import in the `test_kakarot.py` file
- so your system under testing is `compute_starknet_address` right? so create a class `class TestComputeStarknetAddress` in `TestKakarot`
- your test tests that the function returns the same address as the one got from a `deploy`, so name your test function `def test_should_return_same_as_deployed_address`.
- this part could be simplified:

```
        kakarot_starknet_address = eoa_deploy_tx.call_info.internal_calls[
            0
        ].contract_address

        test_compute_starknet_address = await kakarot.compute_starknet_address(
            evm_address=int(evm_address, 16)
        ).execute()

        computed_starknet_address = test_compute_starknet_address.result[0]
        assert computed_starknet_address == kakarot_starknet_address
```

- you don't need to create that many variables, plz also keep the naming explicit and simple, and eventually you can `.call` instead of `.execute` for `compute_starknet_address`.

dragan2234:
@ClementWalter I've refactored the code. Note that in the issue description, it says unit test, but it can be an integration test also, so I've refactored it per your instructions

> update the import in the test_kakarot.py file

no need to do this since the import is:
```
from tests.integration.bytecode.test_cases import test_cases
```

ClementWalter (approved):

PR #511, opened 2023-02-17
Add starknet devnet

ClementWalter:
Time spent on this PR: 0.5

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [x] Build related changes

No starknet devnet

## What is the new behavior?

Starknet-devnet added to the project.
`make run` will start a starknet devnet
`make deploy` will by default target the starknet devnet

## Other information

Don't forget about https://devnet.starkscan.co/ to have a block explorer for your local devnet!


jobez (dismissed):
LGTM. 

I was wondering if we can have a canonical devnet dump that people can download. 

Might help make it a little more frictionless if we want to invite people to test a system locally: https://github.com/Shard-Labs/starknet-devnet/blob/master/page/docs/guide/dumping-and-loading.md

ClementWalter:
> LGTM.
> 
> I was wondering if we can have a canonical devnet dump that people can download.
> 
> Might help make it a little more frictionless if we want to invite people to test a system locally: https://github.com/Shard-Labs/starknet-devnet/blob/master/page/docs/guide/dumping-and-loading.md

actually I have hesitated with the dump BUT I'm scared that it becomes bug prone as it would not be guaranteed that the deployed version is actually the one the dev has locally. here, though it takes few minutes, I think that it's not so long to run `make deploy` after `make run`, thus insuring a sound local devnet.

jobez:
> > LGTM.
> > I was wondering if we can have a canonical devnet dump that people can download.
> > Might help make it a little more frictionless if we want to invite people to test a system locally: https://github.com/Shard-Labs/starknet-devnet/blob/master/page/docs/guide/dumping-and-loading.md
> 
> actually I have hesitated with the dump BUT I'm scared that it becomes bug prone as it would not be guaranteed that the deployed version is actually the one the dev has locally. here, though it takes few minutes, I think that it's not so long to run `make deploy` after `make run`, thus insuring a sound local devnet.

Yeah, makes sense. If the time isn't too extravagant, then the risk of desync isn't really work it.

PR #527, opened 2023-02-22
refactor: eoa library file, remove unnecessary imports, use constants

dragan2234:
While working on the other task I noticed this library has some imports which are not used. I've run `test_eoa.py` and it's passing.

Time spent on this PR:

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Resolves #<Issue number>

## What is the new behavior?

-
-
-

### DIFF
--- a/src/kakarot/accounts/eoa/library.cairo
+++ b/src/kakarot/accounts/eoa/library.cairo
@@ -1,20 +1,12 @@
 %lang starknet
 
-from utils.utils import Helpers
-from kakarot.constants import Constants
 from kakarot.interfaces.interfaces import IEth, IKakarot
-from starkware.cairo.common.cairo_builtins import HashBuiltin, SignatureBuiltin, BitwiseBuiltin
-from starkware.cairo.common.cairo_secp.signature import verify_eth_signature_uint256
-from starkware.starknet.common.syscalls import get_tx_info, get_caller_address, call_contract
+from starkware.cairo.common.cairo_builtins import HashBuiltin, BitwiseBuiltin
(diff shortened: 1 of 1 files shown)


src/kakarot/accounts/eoa/library.cairo
```diff
@@ -90,7 +82,7 @@ namespace ExternallyOwnedAccount {
         bitwise_ptr: BitwiseBuiltin*,
         range_check_ptr,
     }(call_array_len: felt, call_array: CallArray*, calldata_len: felt, calldata: felt*) -> () {
-        if (call_array_len == 0) {
+        if (call_array_len == FALSE) {
```
TotalPizza, line 85: Imo in this scenario:
`call_array_len == 0` is preferable over `call_array_len == FALSE`,
As we're actually checking what the call_array length is and not whether it is true or false.

src/kakarot/accounts/eoa/library.cairo
```diff
@@ -120,7 +112,7 @@ namespace ExternallyOwnedAccount {
         response: felt*,
     ) -> (response_len: felt) {
         alloc_locals;
-        if (call_array_len == 0) {
+        if (call_array_len == FALSE) {
```
ClementWalter, line 115: I'd keep this as "0" as we're comparing a len and not a bool

ClementWalter (changes requested):
thx for the cleaning of the imports, we definitely miss such a tool in cairo format

ClementWalter (approved):

Issue #520, opened 2023-02-21
Workshop: Add tests that expect "fail" to PUSH4 Opcode


danilowhk:
# Kakarot Workshop Issue

Level: Medium

Right now we only have, "ok" tests for PUSH4 Opcode.

For this issue please implement at least 1 extra test where calling PUSH4 would fail or revert.

https://github.com/sayajin-labs/kakarot/blob/286aaea710b22bf78facd5257a9da4b608e68a6e/tests/unit/src/kakarot/instructions/test_push_operations.cairo#L56

(Hint, you can play with Uint256 overflow/underflow , Bytes underflow/overflow)

PR #530, opened 2023-02-22
Overflow tests push operations

### LINKED ISSUE #519: Workshop: Add tests that expect "fail" to PUSH2 Opcode
# Kakarot Workshop Issue

Level: Medium

Right now we only have, "ok" tests for PUSH2 Opcode.

For this issue please implement at least 1 extra test where calling PUSH2 would fail or revert.

https://github.com/sayajin-labs/kakarot/blob/286aaea710b22bf78facd5257a9da4b608e68a6e/tests/unit/src/kakarot/instructions/test_push_operations.cairo#L32

(Hint, you can play with Uint256 overflow/underflow , Bytes underflow/overflow)

dbejarano820:
Time spent on this PR: 1 day

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [X] Other (please describe): Adding more tests

Resolves #519 #520 #521 and #522 

## What is the new behavior?

Testing for overflow when carrying out different push operation to the stack.

## Other information

image


ClementWalter (changes requested):
hey, look at how `init_context_with_bytecode` works more closely you are not testing what you think you test!

Furthermore, I think that kakarot should raise so I would expect you to use 
```
with kakarot_error():
    await push_operations.test__exec_push2_should_not_add_2_byte_to_stack().call()
```

dbejarano820:
> with kakarot_error():

@ClementWalter Hey, thanks for the review! That `kakarot_error` negation helper was exactly what I needed. 

In order to overflow a uint256, the value has to go over 32 bytes correct? I adjusted the numbers I set in the `init_context_with_bytecode` for each case to slightly overflow.

ClementWalter (changes requested):
you are still missing something with the `init_context_with_bytecode`!

The doc says

> @notice Init an execution context where bytecode has "bytecode_count" entries of "value".

here you are giving `value = 0xFFFFF` which is a non sense since value should be a bytecode (so 0 <= value < 256)

furthermore, for triggering an overflow, you want to have `bytecode_count < i` (for PUSH _i)

dbejarano820:
> you are still missing something with the `init_context_with_bytecode`!
> 
> The doc says
> 
> > @notice Init an execution context where bytecode has "bytecode_count" entries of "value".
> 
> here you are giving `value = 0xFFFFF` which is a non sense since value should be a bytecode (so 0 <= value < 256)
> 
> furthermore, for triggering an overflow, you want to have `bytecode_count < i` (for PUSH _i)

Gotcha, thanks for the clarification... makes sense.

tests/unit/src/kakarot/instructions/test_push_operations.cairo
```diff
@@ -40,6 +40,17 @@ func test__exec_push2_should_add_2_byte_to_stack{
...
 }
 
+@external
+func test__exec_push2_should_not_add_2_byte_to_stack{
+    syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr, bitwise_ptr: BitwiseBuiltin*
+}() {
+    alloc_locals;
+    let ctx: model.ExecutionContext* = TestHelpers.init_context_with_bytecode(1, 0xFF);
+    let result = PushOperations.exec_push2(ctx);
+    TestHelpers.assert_stack_last_element_contains_uint256(result.stack, Uint256(0xFFFF, 0));
```
ClementWalter, line 50: this is probably never read, the test should fail at `PushOperations.exec_push2`

ClementWalter: actually for this task you may want to have instead one single test using `exec_push_i` and a `parametrize`, like:

```
@external
func test_exec_push_should_raise{
    syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr, bitwise_ptr: BitwiseBuiltin*
}(i: felt) {
    alloc_locals;
    let ctx: model.ExecutionContext* = TestHelpers.init_context_with_bytecode(i - 1, 0xFF);
    PushOperations.exec_push_i(ctx, i);
}
```

and in the python file

```
@pytest.mark.parametrize("i", range(2, 32))
async def test_should_raise(i):
    with kakarot_error():
        await ...
```

ClementWalter (changes requested):

tests/unit/src/kakarot/instructions/test_push_operations.cairo
dbejarano820: Yea with this we could test all of the push amounts. Much better approach, I like it.

dbejarano820: I'll look into abstracting the other existing tests and using a `parametrize` as well. Seems to me that there is a lot of repetition.

dbejarano820: @ClementWalter Let me know your thoughts, I added another commit which refactors the other tests. I also updated the screenshot in the description for this change.

tests/unit/src/kakarot/instructions/test_push_operations.cairo
```diff
@@ -16,441 +17,40 @@ from kakarot.instructions.push_operations import PushOperations
...
+    let ctx: model.ExecutionContext* = TestHelpers.init_context_with_bytecode(i - 1, 0xFF);
+    PushOperations.exec_push_i(ctx, i);
+    
     return ();
-}
+} 
 
 @external
-func test__exec_push20_should_add_20_byte_to_stack{
+func test__exec_push_should_add_1_through_16_bytes_to_stack{
```
ClementWalter, line 31: naming here is odd. You want to keep it symmetric with the first test, like

```
func test__exec_push_should_raise
func test__exec_push_should_push
```

ClementWalter: use only one `test__exec_push_should_push` and assert in python, it'll be easier

tests/unit/src/kakarot/instructions/test_push_operations.cairo
```diff
@@ -16,441 +17,40 @@ from kakarot.instructions.push_operations import PushOperations
...
-    let result = PushOperations.exec_push32(ctx);
-    TestHelpers.assert_stack_last_element_contains_uint256(
-        result.stack,
-        Uint256(0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF, 0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF),
-    );
+    TestHelpers.assert_stack_last_element_contains_uint256(result.stack, Uint256(0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF,  res - 0x01));
 
     return ();
-}
+}
```
ClementWalter, line 56: enable workflow in your repo to have the CI format your code
and run `make format` locally as well!

tests/unit/src/kakarot/instructions/test_push_operations.py
```diff
@@ -1,7 +1,7 @@
 import pytest
 import pytest_asyncio
 from starkware.starknet.testing.starknet import Starknet
-
+from tests.utils.errors import kakarot_error
```
ClementWalter, line 4: see above mentioned comment for formatting

tests/unit/src/kakarot/instructions/test_push_operations.py
```diff
@@ -11,101 +11,17 @@ async def push_operations(starknet: Starknet):
...
-    async def test__exec_push32_should_add_32_byte_to_stack(self, push_operations):
-        await push_operations.test__exec_push32_should_add_32_byte_to_stack().call()
+    @pytest.mark.parametrize('i', range(2, 33))
+    async def test__exec_push_should_raise(self, push_operations, i):
+        with kakarot_error():
+            await push_operations.test__exec_push_should_raise(i).call()
+
+    @pytest.mark.parametrize('i', range(1, 17))
+    async def test__exec_push_should_add_1_through_16_bytes_to_stack(self, push_operations, i):
+        await push_operations.test__exec_push_should_add_1_through_16_bytes_to_stack(i).call()
```
ClementWalter, line 23: so update this

```
peek = await push_operations.test_should_push(i).call()
assert peek = 256**i - 1
```

tests/unit/src/kakarot/instructions/test_push_operations.py
```diff
@@ -11,101 +11,17 @@ async def push_operations(starknet: Starknet):
...
+
+    @pytest.mark.parametrize('i', range(17, 33))
+    async def test__exec_push_should_add_17_through_32_bytes_to_stack(self, push_operations, i):
+        await push_operations.test__exec_push_should_add_17_through_32_bytes_to_stack(i).call()
```
ClementWalter, line 27: again, `make format`

ClementWalter (changes requested):

tests/unit/src/kakarot/instructions/test_push_operations.py
dbejarano820: How can I get the stack's last value here?

ClementWalter: you can return it in the cairo test func, can't you?

dbejarano820: Yea, I had to look into the struct of `StarknetCallInfo`. Wasn't aware of it.

ClementWalter (approved):

tests/unit/src/kakarot/instructions/test_push_operations.cairo
dbejarano820: Asserting in python is cleaner for sure, nice tip

dbejarano820:
@ClementWalter I force pushed the branch to clean up the import of `pow` I was no longer using in `test_push_operations.cairo`

ClementWalter (approved):

PR #531, opened 2023-02-22
Workshop: update boolean constants

### LINKED ISSUE #517: Workshop: Booleans in Cairo
Level: Easy

As Cairo doesn't have a native boolean type, it is not uncommon to see developers using 0 and 1 as a replacement.
E.g. 
```
if (result == 0):
    //Do Something
```
However, for readability purposes we prefer to utilize the FALSE and TRUE constants from the common cairo-lang library:
https://github.com/starkware-libs/cairo-lang/blob/master/src/starkware/cairo/common/bool.cairo

In the following file 1 is still used instead of the TRUE constant.
https://github.com/sayajin-labs/kakarot/blob/f519876a3acb67507ccea05e1f87d82f36093142/src/utils/utils.cairo#L257-L260

To fix this import the relevant cairo-lang library and replace 1 with the TRUE constant.

weiihann:
Time spent on this PR:

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Cairo boolean values are represented with 0 and 1, which might not be good for readability.

Resolves #517

## What is the new behavior?
FALSE and TRUE constants from the common cairo-lang library are utilized instead of 0 and 1.

## Other information
N/A


src/kakarot/precompiles/blake2f.cairo
```diff
@@ -121,7 +122,7 @@ namespace Blake2 {
 
         // Compute state[14].
         local state14;
-        if (f == 1) {
+        if (f == TRUE) {
```
ClementWalter, line 125: actually it's better to test against FALSE, so can you switch the if/else : 

```
if (f == FALSE) {
    state14 = 0x1f83d9abfb41bd6b;
} else {
    state14 = 0xe07c265404be4294;
}
```

not that for this exact case, you case as well do

```
let state14 = (1 - f) * 0x1f83d9abfb41bd6b + f * 0xe07c265404be4294
```

src/kakarot/precompiles/ripemd160.cairo
```diff
@@ -127,7 +128,7 @@ func compress_data{dict_ptr: DictAccess*, range_check_ptr, bitwise_ptr: BitwiseB
 ) -> (res: felt*, rsize: felt, new_msg: felt*) {
     alloc_locals;
     let len_lt_63 = is_le(input_len, 63);
-    if (len_lt_63 == 1) {
+    if (len_lt_63 == TRUE) {
```
ClementWalter, line 131: inverse `is_le` test and use `FALSE` plz

ClementWalter (commented):
plz try to use FALSE everywhere

ClementWalter (changes requested):
plz use `FALSE` instead `TRUE` when possible

weiihann:
> plz use `FALSE` instead `TRUE` when possible

Added necessary changes. Please review again :)

src/kakarot/precompiles/blake2f.cairo
weiihann: May I ask why is it better to test against FALSE?

src/utils/utils.cairo
```diff
@@ -23,7 +23,7 @@ from starkware.cairo.common.uint256 import (
 )
 from starkware.cairo.common.registers import get_label_location
 from starkware.cairo.common.cairo_secp.bigint import BigInt3, bigint_to_uint256, uint256_to_bigint
-from starkware.cairo.common.bool import FALSE
+from starkware.cairo.common.bool import FALSE, TRUE
```
ClementWalter, line 26: remove unused TRUE svp

ClementWalter (changes requested):

src/utils/utils.cairo
weiihann: Done

ClementWalter (approved):

ClementWalter (approved):

PR #534, opened 2023-02-24
Feat/new nonce handling

### LINKED ISSUE #505: feat: use individual contract nonces
# Feature Request

Currently new contract_accounts are deployed using the kakarot `salt` which is a global nonce that is being incremented for every deployed contract.
By tracking an individual nonce for each contract_account and eoa, the deployed contract will have the same address as they would on ethereum.

For the eoa it might be viable to use the nonce provided by starknet (which can accessed via the syscall `get_tx_info`).
That way we save storage costs.
Kakarot RPC can fetch the nonce via the Starknet cli.
But for the Create and Create2 opcodes we require the nonces of the contract_account that holds the respective opcodes. 
Therefore I'd suggest adding a `nonce` storage var, as well as a `@view` and `increment` method.

**Describe Alternatives**

Alternatively we can let the current nonce system be. It's not a big blocker as far as I can tell.
It's just another (imo important) step to EVM compatibility. It would enable DAOs to have the same address for their DAPP or multisig as on every other EVM chain.

Yes

TotalPizza:
Instead of using a global salt to create unique evm addresses, we utilize the individual EOA or contract account nonces.

Time spent on this PR: 3d

## Pull request type

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes

Resolves #505 

## What is the new behavior?

- Contract deployments utilize contract account and eoa nonce for the generation of the evm address instead of a global salt.
- The starknet account nonce is used for the EOA nonce
- Contract accounts receive a new storage var to store it's individual nonce.


tests/unit/src/kakarot/accounts/contract/test_contract_account.py
```diff
@@ -90,3 +90,29 @@ async def test_should_give_inifinite_allowance_to_kakarot(
...
+
+            # Get new nonce
+            assert initial_nonce == await contract_account.get_nonce().call()
+
+
+        async def test_should_influence_address_generation(
+            self, contract_account: StarknetContract, kakarot
+        ):
+            x = 1
+    """
```
TotalPizza, line 118: @ClementWalter 
I'm looking to create some tests to see if the nonces are handled correctly.

What I'm going for here is more an integration test, right?
What scope do you think would be fitting?

TotalPizza:
It seems my local spellchecker has failed me. Will get that updated and fixed asap.

src/kakarot/library.cairo
```diff
@@ -185,21 +185,16 @@ namespace Kakarot {
         pedersen_ptr: HashBuiltin*,
         range_check_ptr,
         bitwise_ptr: BitwiseBuiltin*,
-    }(bytecode_len: felt, bytecode: felt*) -> (
+    }(nonce: felt, bytecode_len: felt, bytecode: felt*) -> (
```
ClementWalter, line 188: why do you need to get the `nonce` as input here instead of retrieving it like in `CREATE`?

tests/unit/src/kakarot/accounts/contract/test_contract_account.py
ClementWalter: so the only way for a contract_account to have its nonce increased is to have a CREATEright?

At the unit test level you may only want to test that `get_nonce` returns the nonce and `increment_nonce` increment it.

At the integration test level, you may want to test that when deploying a contract with CREATE the nonce increases

ClementWalter (commented):
need to stop here so posting half a review

src/kakarot/accounts/eoa/library.cairo
```diff
@@ -128,10 +128,14 @@ namespace ExternallyOwnedAccount {
 
         // If destination is 0, we are deploying a contract
         if (destination == 0) {
+            let (tx_info) = get_tx_info();
```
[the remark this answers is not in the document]
TotalPizza: `CREATE` opcodes utilize the contract account nonce and not the eoa nonce, so that should not be an issue, right?
I think you might have a point if we do a multicall in an eoa.

I discussed this in DMs with elias and he posted the summary in the issue:
https://github.com/sayajin-labs/kakarot/issues/505

src/kakarot/instructions/system_operations.cairo
```diff
@@ -109,7 +108,7 @@ namespace SystemOperations {
         // 0 - value: value in wei to send to the new account
         // 1 - offset: byte offset in the memory in bytes (initialization code)
         // 2 - size: byte size to copy (size of initialization code)
-        // 3 - salt: salt for address generation
+        // 3 - nonce: nonce for address generation
```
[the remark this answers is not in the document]
TotalPizza: oh dang. true.
Will fix.

src/kakarot/library.cairo
TotalPizza: hmmm true, I guess i can also fetch the transaction info here.
No unnecessary passing of arguments.
will fix!

src/kakarot/accounts/eoa/library.cairo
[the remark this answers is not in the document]
ClementWalter: indeed, and I even commented this thread ... sorry for the irrelevant comment

src/kakarot/accounts/eoa/library.cairo
```diff
@@ -129,7 +129,7 @@ namespace ExternallyOwnedAccount {
         // If destination is 0, we are deploying a contract
         if (destination == 0) {
             // deploy_contract_account signature is
-            // calldata_len: felt, calldata: felt*
+            // nonce: felt, calldata_len: felt, calldata: felt*
```
ClementWalter, line 132: forgot to remove?

tests/unit/src/kakarot/accounts/contract/test_contract_account.py
```diff
@@ -90,3 +90,22 @@ async def test_should_give_inifinite_allowance_to_kakarot(
...
+            # Increment nonce
+            await contract_account.increment_nonce().execute(
+                caller_address=kakarot.contract_address
+            )
+
+            # Get new nonce
+            assert initial_nonce + 1 == (await contract_account.get_nonce().call()).result.nonce
+
+            # Only kakarot should be able to increment
+            with kakarot_error():
```
ClementWalter, line 110: could you move this to another unit test function?

```
async def test_should_raise_when_caller_is_not_kakarot()
...
```

tests/unit/src/kakarot/instructions/test_system_operations.cairo
```diff
@@ -48,6 +47,16 @@ func constructor{
     return ();
 }
 
+// ///////////////////////////////////////////////////////////////////////////////////////////////////////////////
+//                                          MOCK FUNCTIONS                                                      //
+// The Kakarot, EOA, Account Contract and ETH contracts often times require communication between each other.   //
```
ClementWalter, line 52: ... Contract Account and ETH ...

ClementWalter (approved):
minor changes but lgtm!

src/kakarot/accounts/eoa/library.cairo
TotalPizza: ....potentially

Eikix (commented):
Nice PR! Does this somehow affect `compute_starknet_address` for Kakarot?

TotalPizza:
> Nice PR! Does this somehow affect `compute_starknet_address` for Kakarot?

I'm currently creating integration tests for this feature and exploring why other tests are failing.
My hunch is that `tx_info.nonce` doesn't increment in the tests, resulting in some contracts failing to be deployed as the same address is generated as a contract that already exists.

TotalPizza:
@ClementWalter, @Eikix 
Sigh... yeah so from what I can tell the issue is that the `StarknetContract` class makes use of the `execute()` function which always uses `nonce=0`.
So if we save the state between contract deployments (and use  the `tx_info.nonce` as the salt) then we get conflicts in the tests because we generate the same addresses.
I'll play around a bit and try to find a good solution.

ClementWalter:
> @ClementWalter, @Eikix Sigh... yeah so from what I can tell the issue is that the `StarknetContract` class makes use of the `execute()` function which always uses `nonce=0`. So if we save the state between contract deployments (and use the `tx_info.nonce` as the salt) then we get conflicts in the tests because we generate the same addresses. I'll play around a bit and try to find a good solution.

might be related but when trying to test ORIGIN with @hurrikaanig we faced an issue with `tx_info` as well, this one being that `origin` was 0.

Eikix:
> @ClementWalter, @Eikix Sigh... yeah so from what I can tell the issue is that the `StarknetContract` class makes use of the `execute()` function which always uses `nonce=0`. So if we save the state between contract deployments (and use the `tx_info.nonce` as the salt) then we get conflicts in the tests because we generate the same addresses. I'll play around a bit and try to find a good solution.

it's a good question to ask at the starknet meetup tonight to the teams that develop `starknet-in-rust` and `cairo1` since it's really hard to simulate nonce behaviour locally

TotalPizza:
Yeah, so I don't think we can influence the tx_info.nonce unless we make changes to the StarkWare library.
An alternative is, that we ensure that contracts are deployed by different users.
So if we have a test where a token_a, token_b and factory is deployed, they have to be deployed by different addresses.
That also means that the kakarot state wouldn't be allowed to transfer between different testes (if they want to use some of the same addresses to deploy contracts). --> IMO that generally sounds like an awful solution.

...So I'd almost be for adding the same nonce handling for the EOAs that we use for the contract accounts. (And add some comments to the code why we are currently not using the tx_info)

ClementWalter (dismissed):

ClementWalter:
> Yeah, so I don't think we can influence the `tx_info.nonce` unless we make changes to the StarkWare library. An alternative is, that we ensure that contracts are deployed by different users. So if we have a test where a `token_a`, `token_b` and `factory` is deployed, they have to be deployed by different addresses. That also means that the kakarot state wouldn't be allowed to transfer between different testes (if they want to use some of the same addresses to deploy contracts). --> IMO that generally sounds like an awful solution. Especially as we're heavily using the `owner` and `addresses` fixtures.
> 
> ...So I'd almost be for adding the same nonce handling for the EOAs that we use for the contract accounts. (And add some comments to the code why we are currently not using the tx_info).
> 
> Does that work for you guys?

I'm fine with adding a `nonce` in the EAO as we have un the CA; I'm less ok with having to use different deployers in test just because the nonce does not change in the test suite because this is very very unexplicit.

TotalPizza:
@ClementWalter
The changes I made to `tests/integration/conftest.py` are probably the most controversial ones.
Optimally we would send the deployment transactions entirely through the eoa. However, that requires a bunch of setup to build the transaction...this was sort of the easy/quick fix.

I could refactor the integration tests to make deployments through the EOA... but maybe that should be a separate PR?

ClementWalter (approved):

ClementWalter:
> @ClementWalter The changes I made to `tests/integration/conftest.py` are probably the most controversial ones. Optimally we would send the deployment transactions entirely through the eoa. However, that requires a bunch of setup to build the transaction...this was sort of the easy/quick fix.
> 
> I could refactor the integration tests to make deployments through the EOA... but maybe that should be a separate PR?

just merged the PR, was better not to keep this ongoing work, gg! plz can you create an issue for the work you're mentioning?

PR #540, opened 2023-03-07
feature: add `query_logs` method to kkrt wrapper

jobez:
Time spent on this PR: 0.5

## Pull request type

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes

Revert tests were unable to assert that a given event emitted from a subcontext was not emitted.

Resolves #<Issue number>

## What is the new behavior?

The kakarot wrapper now has a `query_logs` method that, when called with 
- no args -> returns the raw logs of transaction.
- contract key -> returns all the events in logs that match the event abi of the contract
-  contract key  x event name -> returns the events that match the particular event name

### DIFF
--- a/src/kakarot/execution_context.cairo
+++ b/src/kakarot/execution_context.cairo
@@ -86,7 +86,9 @@ namespace ExecutionContext {
         let (empty_create_addresses: felt*) = alloc();
 
         let (local revert_contract_state_dict_start) = default_dict_new(0);
-        tempvar revert_contract_state: model.RevertContractState* = new model.RevertContractState(revert_contract_state_dict_start, revert_contract_state_dict_start);
+        tempvar revert_contract_state: model.RevertContractState* = new model.RevertContractState(
+            revert_contract_state_dict_start, revert_contract_state_dict_start
+        );
 
         // Define initial program counter
         let initial_pc = 0;
@@ -125,7 +127,7 @@ namespace ExecutionContext {
             revert_contract_state=revert_contract_state,
             reverted=FALSE,
             read_only=FALSE,
-            );
+        );
         return self;
     }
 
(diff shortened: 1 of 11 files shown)


src/kakarot/execution_context.cairo
```diff
@@ -181,7 +183,9 @@ namespace ExecutionContext {
         let (empty_events: model.Event*) = alloc();
         let (empty_create_contracts: felt*) = alloc();
         let (local revert_contract_state_dict_start) = default_dict_new(0);
-        tempvar revert_contract_state: model.RevertContractState* = new model.RevertContractState(revert_contract_state_dict_start, revert_contract_state_dict_start);
+        tempvar revert_contract_state: model.RevertContractState* = new model.RevertContractState(
```
jobez, line 186: i updated my env by calling `make setup` and these syntax revisions was the result of calling `make format` with that updated env

tests/integration/helpers/wrap_kakarot.py
```diff
@@ -123,10 +171,14 @@ async def _wrapped(contract, *args, **kwargs):
 # Example: get_contract("Solmate", "ERC721") will load the ERC721.sol file in the tests/integration/solidity_contracts/Solmate folder
 # Example: get_contract("StarkEx", "StarkExchange") will load the StarkExchange.sol file in the tests/integration/solidity_contracts/StarkEx/starkex folder
 #
-def get_contract(contract_app: str, contract_name: str) -> Contract:
+def get_contract(
```
jobez, line 174: I am seeing this warning when running tests locally and I assume it is because I added an optional argument here:
```
/home/jhnn/.cache/pypoetry/virtualenvs/kakarot-mTn7Smmv-py3.9/lib/python3.9/site-packages/marshmallow/fields.py:945: RemovedInMarshmallow4Warning: The 'default' argument to fields is deprecated. Use 'dump_default' instead.
    super().__init__(**kwargs)
```
Not sure how to resolve this. 

This optional argument was so I could use `get_contract` to get the contract object of `ContractRevertsOnMethodCall` which is one of two contracts defined in the file `RevertTestCases`

jobez (commented):
I am seeing a warning output when running test input that looks like:

```
/home/jhnn/.cache/pypoetry/virtualenvs/kakarot-mTn7Smmv-py3.9/lib/python3.9/site-packages/marshmallow/fields.py:945: RemovedInMarshmallow4Warning: The 'default' argument to fields is deprecated. Use 'dump_default' instead.
    super().__init__(**kwargs)
```
I left a subcomment about what change I believe is causing it and my justification for that change.

ClementWalter (changes requested):
could you just update the pyproject.toml to ignore these warnings so as not to pollute the test output? there's also a warning concerning the `MODEXP` mark, probably a typo somewhere, can you fix this as well plz?

ClementWalter (approved):

Issue #411, opened 2023-01-04
test: Refactor test fixtures


matthieuauger:
The test fixtures are prefixed in their name, initially to ease readability and sorting them in the file tree by their loading order.

![Image](https://user-images.githubusercontent.com/1172099/210599715-c4efce85-23be-495c-ab4b-6e04896544ae.png)

However the prefixes are hard to maintain and to keep up to date with reality. For example, we have 5_blockhashes which contain fixtures loaded before the starknet fixtures.

We are also loading all fixtures in the top conftest.py, consequently even for running a smol unit test we deploy all the contracts of the repo, we’d need to fix this by carefully placing the fixture right where they are required in subfolder’s conftest.py.

We want to : 
- remove the prefix in the test names
-  place the fixtures in the right conftest to optimize test execution time

mradolux:
Hello ! I'd like to take this issue please

ptisserand:
Hi @mradolux , are you still working on this issue?

mradolux:
Hi @ptisserand, I didn't get assigned to it

Issue #380, opened 2022-12-19
dev: Update mermaid diagram (interaction diagram) to better reflect current state of Kakarot


Eikix:

Flydexo:
What parts are missing ?

Eikix:
non-exhaustive list:
- we don't have a  `generate_instruction()` method anymore

we haven't diagramed `execute_at_address` nor `deploy`

PR #554, opened 2023-03-22
Add eth_call and eth_send_transaction entrypoints to kakarot

### LINKED ISSUE #551: feat: refacto execute_at_address into eth_call
# Feature Request

The logic between call and deployment is split in between kakarot (execute_at_address) and EOAccounts (https://github.com/sayajin-labs/kakarot/blob/ecc216e86642e946ed2af4c378c36d33c043444c/src/kakarot/accounts/eoa/library.cairo#L131)

We need to merge this in one single `eth_call` entrypoint for easing inter operability with the starknet ecosystem.

Rename `execute_at_address` into `eth_call``
Remove `deploy_contract_account`
Add `gas_price` to `eth_call`
Add logic from https://github.com/sayajin-labs/kakarot/blob/ecc216e86642e946ed2af4c378c36d33c043444c/src/kakarot/accounts/eoa/library.cairo#L131 into this function
- ⚠️ the value transfer (if amount > 0) should be made **before** and not **after** the call

ClementWalter:
Time spent on this PR: 2.5

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

The EOAccount owns the logic of transferring native tokens and dispatching between contract deployment and contract invocation.
On the other hand, the kakarot contract exposes two methods (execute_at_address & deploy_contract_account) that require the caller to know what they want to do and how they need to send the data. This is an extra complexity compared to what the EVM does: the kakarot contract should be able to handle a plain (encoded or decoded) EVM tx without requiring the caller to do any logic themselves.

As a matter of fact, the EOAccount responsibility is (almost) narrowed to a standard AA responsibility: accepte a sign tx, verify signature and propagate the call. Almost because it still needs to **decode** the tx, and actually Yoav (from Braavos) thinks that we should go further and remove this **decoding** part from the EOA: only validate a signature and send the payload (encoded) to the kakarot contract. This would allow us to use Braavos contract out of the box (with a proper signer type)

Resolves #551 #411 

## What is the new behavior?

The Kakarot contract has two new entrypoints that behave as in the [ethereum RPC doc](https://ethereum.org/en/developers/docs/apis/json-rpc)
- `eth_call`: Executes a new message call immediately without creating a transaction (`@view` function)
- `eth_send_transaction`: Creates new message call transaction or a contract creation, if the data field contains code. (`@external` function)

Along the way, the test refactoring (#441) was eventually done, and the `ExecutionContext` `init` & `init_at_address` have been merged into one single general purpose `init` method.

## Other information

Following a community call discussion, we could put more of these entrypoints in the main Kakarot contract:
- eth_getBalance
- eth_getStorageAt
- eth_getTransactionCount
- eth_getCode
- etc.

to put more logic in the core contract and less in the RPC thay may eventually require a complete refacto depending on where we actually make this kakarot live.

I have added a check `compute_starknet_address(IAccount.get_evm_address(contract_address)) != contract_address` to avoid unexpected behaviors resulting from a starknet contract implementing `get_evm_address` without being **the** `ContractAccount`. I'd like to remove this sooner or later but still need to think about all the edge cases.


docs/general/kakarot_components.md
```diff
@@ -9,54 +9,84 @@ The entire Kakarot protocol is composed of 4 different contracts:
 
 ## Kakarot
 
-The main Kakarot contract is located at: `./src/kakarot/kakarot.cairo`.
+The main Kakarot contract is located at:
+[`./src/kakarot/kakarot.cairo`](../../src/kakarot/kakarot.cairo).
 
-This is the core contract which is capable of parsing and executing ethereum bytecode.
+This is the core contract which is capable of executing decoded ethereum
```
jobez, line 15: might be good here to talk about how in our system, account abstraction is responsible for validating signed transactions. maybe we can create an issue for a diagram or something if one doesnt exist.

tests/unit/src/kakarot/accounts/eoa/test_library.py
```diff
@@ -32,9 +37,10 @@ class TestLibrary:
     async def test_execute_should_make_all_calls_and_return_concat_results(
         self, mock_externally_owned_account, eth
     ):
-        private_key = generate_random_private_key()
+        private_key = keys.PrivateKey(int.to_bytes(0xDEAD, 32, "big"))
```
jobez, line 40: is 0xDEAD being used here because of a convention of an account deployed on devnet? 

maybe it can be defined as a constant where the variable name offers some clarity.

src/kakarot/kakarot.cairo
```diff
@@ -204,12 +91,41 @@ func deploy_externally_owned_account{
     return Kakarot.deploy_externally_owned_account(evm_address);
 }
 
-// @notice Compute the starknet address of a contract given its EVM address
-// @param evm_address The EVM address of the contract
-// @return contract_address The starknet address of the contract
+// @notice The eth_call function as described in the spec, see https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_call
```
jobez, line 94: appreciate linking out to the spec!

one short little comment that describes the difference of how eth_call differs from invoke might be useful for people new to blockchains/eth/starknet would be good here also.

the simplest way to put it is (maybe) its an evaluation of bytecode that doesn't update any state.

jobez (changes requested):
Great stuff. Just some comments around some documentation and a curiousity about if this gas_price change was for a particular blocker.

src/kakarot/accounts/eoa/library.cairo
```diff
@@ -117,76 +117,31 @@ namespace ExternallyOwnedAccount {
             return (response_len=0);
         }
 
+        Accounts.increment_nonce();
+
         let (
-            gas_limit, destination, amount, payload_len, payload, tx_hash, v, r, s
+            gas_price, gas_limit, destination, amount, payload_len, payload, tx_hash, v, r, s
```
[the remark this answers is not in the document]
ClementWalter: not really but the signature of `eth_call` takes this gas_price as input and it's actually how I figured out that we messed up with the OPCODE you fixed

tests/unit/src/kakarot/accounts/eoa/test_library.py
ClementWalter: And can't remember who introduced this convention in the project to use this 0XDEAD as a mean of "a random number we don't care about".

Could move this to a DEFAULT_PRIVATE_KEY in a constant file if you prefer, not sure it's much better though

jobez: ah, wasn't familiar with this!

if the `0xDEAD` usage is broader than just a default private key and just used as a random filler value, then maybe there's a place in testing/deployment docs where that can be mentioned

docs/general/kakarot_components.md
ClementWalter: there is actually a stalled issue for a diagram update! see #380 !

tests/unit/src/kakarot/accounts/eoa/test_library.py
ClementWalter: doing search for all "0xDEAD", it's used as a magic number for addresses & private keys in tests. I've removed all the usage apart from one explicit one `test__exec_extcodehash__should_handle_invalid_address` where `0xDEAD` is clearly used to be "the invalid address"

src/kakarot/execution_context.cairo
```diff
@@ -234,6 +138,25 @@ namespace ExecutionContext {
         );
     }
 
+    // @notice Finalizes the execution context.
```
jobez, line 141: is it worth linking here to some sort of documentation about the 'rules' of using dictionaries (and indirectly how memory and the stack are implemented as dictionaries) and how finalizing for the memory and stack involves squashing? 

https://www.cairo-lang.org/docs/reference/common_library.html#common-library-default-dict

From Eni's blogpost about sierra

>  Every Dict must be squashed at the end of a program by calling the dict_squash(Dict<K,V>) -> () function, which verifies the consistency of the sequence of key updates. Unsquashed dictionaries are dangerous, as a malicious prover could prove the correctness of inconsistent updates.

https://medium.com/nethermind-eth/under-the-hood-of-cairo-1-0-exploring-sierra-7f32808421f5

tests/integration/solidity_contracts/conftest.py
```diff
@@ -0,0 +1,118 @@
...
+
+from tests.utils.contracts import get_contract, use_kakarot_backend
+from tests.utils.helpers import hex_string_to_bytes_array
+from tests.utils.reporting import traceit
+
+logger = logging.getLogger()
+
+
+@pytest.fixture(scope="package")
+def get_starknet_address(account_proxy_class, kakarot):
```
jobez, line 18: so the broad refactoring across conftest files are to ensure that fixtures only get loaded for the specific types of tests that depend on them?

 a brief description of the motivation  will help ensure that I'm on the same page with you and can provide a more informed review

jobez (changes requested):
Could you please provide more details or clarification on the rationale for why `ExecutionContext.init` is refactored? 

It is the one major part that I am difficulty understanding in relationship to the PR description. I speculated the reason behind some of the conftest refactoring/test structure--an update to the PR description with your rationale for those changes would be appreciated as well :)

Aside from that, I left small comments about updating docstrings and markdown formating for docs! 

Great work!

docs/general/kakarot_components.md
```diff
@@ -9,54 +9,86 @@ The entire Kakarot protocol is composed of 4 different contracts:
...
+
+1. Declare the account proxy, the contract account and the externally owner
+   account contracts.
 
-1. Declare the *Contract Account*.
+This generates a class hash which will be used by the core Kakarot contract to
+deploy _accounts_.
 
-This generates a class hash which will be used by the core Kakarot contract to deploy *Contract Accounts*.
+1. Deploy Kakarot (core), with the following constructor arguments:
```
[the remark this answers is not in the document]
ClementWalter: indeed, markdown rendering turns this into a real numbered list, see the corresponding file (btw I had a indentation issue that this comment made me fix).
The advantage is that you can swap or introduces items without having to re-number everything
see result here https://github.com/sayajin-labs/kakarot/blob/dca8fd49c902ace28a088c821188d6c4c25e8c29/docs/general/kakarot_components.md#deploying-kakarot

docs/general/kakarot_components.md
```diff
@@ -9,54 +9,86 @@ The entire Kakarot protocol is composed of 4 different contracts:
...
+     Kakarot EVM)
+   - _Contract Account_ class hash.
+   - _Externally Owned Account_ class hash.
+   - _Account Proxy_ class hash.
 
-3. Deploy Account Registry
+1. Deploy Blockhash Registry
 
-4. Deploy Blockhash Registry
+1. Store the addresses of the account and blockhash registry contracts in
```
Eikix, line 89: `set_account_registry` there is no more account registry

src/kakarot/library.cairo
```diff
@@ -269,9 +282,118 @@ namespace Kakarot {
...
+    // @return return_data The returned bytes array
+    func eth_call{
+        syscall_ptr: felt*,
+        pedersen_ptr: HashBuiltin*,
+        range_check_ptr,
+        bitwise_ptr: BitwiseBuiltin*,
+    }(to: felt, gas_limit: felt, gas_price: felt, value: felt, data_len: felt, data: felt*) -> (
+        return_data_len: felt, return_data: felt*
+    ) {
+        let (success) = transfer(to, value);
```
Eikix, line 309: I don't think eth_call is allowed to transfer value, as there is no transaction posted on the block chain
image

src/kakarot/library.cairo
```diff
@@ -269,9 +282,118 @@ namespace Kakarot {
...
+    // @dev Raise if the declared corresponding evm address (retrieved with get_evm_address)
+    //      does not lead to the actual caller address
+    func assert_caller_is_kakarot_account{
+        syscall_ptr: felt*,
+        pedersen_ptr: HashBuiltin*,
+        range_check_ptr,
+        bitwise_ptr: BitwiseBuiltin*,
+    }() {
+        alloc_locals;
+        let (local actual_starknet_address) = get_caller_address();
```
Eikix, line 363: imho `actual_starknet_address` can be `caller_address`

src/kakarot/library.cairo
```diff
@@ -269,9 +282,118 @@ namespace Kakarot {
...
+    // @return return_data The returned bytes array
+    func eth_send_transaction{
+        syscall_ptr: felt*,
+        pedersen_ptr: HashBuiltin*,
+        range_check_ptr,
+        bitwise_ptr: BitwiseBuiltin*,
+    }(to: felt, gas_limit: felt, gas_price: felt, value: felt, data_len: felt, data: felt*) -> (
+        return_data_len: felt, return_data: felt*
+    ) {
+        assert_caller_is_kakarot_account();
```
Eikix, line 396: can you explain the logic here?

as I understand, 
- library function `Library.eth_call` can be called only as @view in kakarot contract -> won't trigger a tx
- `eth_send_transaction`  can be called as @external in kakarot contract. And here, we assert the caller is the Kakarot contract, and then call Library.eth_call -> that means we are in the context of send_transaction and as such there is a tx being posted?

src/kakarot/library.cairo
ClementWalter: actually the `eth_call` does have a `value`, "Integer of the value sent with this transaction", and I think that the comment in the doc regarding the gas: eth_call consumes zero gas, but this parameter may be needed by some executions applies to value as well. Even though this is only "view", the value may be required. 

if we consider the tagline of eth_call and eth_sendTransaction
- Executes a new message call immediately without creating a transaction on the block chain.
- Creates new message call transaction or a contract creation, if the data field contains code.

it makes sense to me that they do exactly the same *but* actually writing on the blockchain

src/kakarot/library.cairo
ClementWalter: > we assert the caller is the Kakarot contract

we don't assert that the caller is **the kakarot contract**, we assert that the calling starknet contract is **a kakarot account**, ie an EOA/CA deployed through kakarot. The problem here is that any starknet contract can implement `get_evm_address` and returns whatever they want. Without this check, such a contract could trigger a transaction to Kakarot pretending to be a given EVM address

src/kakarot/kakarot.cairo
```diff
@@ -204,12 +91,45 @@ func deploy_externally_owned_account{
...
+    syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr, bitwise_ptr: BitwiseBuiltin*
+}(to: felt, gas_limit: felt, gas_price: felt, value: felt, data_len: felt, data: felt*) -> (
+    return_data_len: felt, return_data: felt*
+) {
+    return Kakarot.eth_call(to, gas_limit, gas_price, value, data_len, data);
+}
+
+// @notice The eth_send_transaction function as described in the spec,
+//         see https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_sendtransaction
+// @dev "from" parameter is taken from get_caller_address syscall
```
jobez, line 117: another great contextualizing comment ☄️

jobez (approved):
great stuff. went through this slowly and the documentation and doc strings feel strong!

PR #556, opened 2023-03-24
Fix EthTransaction assumes that s and r are 32 bytes long

ClementWalter:
Time spent on this PR: 0.5

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

The EthTransaction.decode function assumes that s and r are 32 bytes long (uses `bytes32_to_uint256` helper) while it's not always the case.

## What is the new behavior?

Make use of `bytes_i_to_uint256` and `item_len`

## Other information

Added a transaction in `TRANSACTIONS` and fixed seeds for random private keys to test this special case


jobez (approved):
LGTM! 

Do you have the output of what it looked like when this issue was failing? Might be good to document for future diagnosing.

ClementWalter:
> LGTM!
> 
> Do you have the output of what it looked like when this issue was failing? Might be good to document for future diagnosing.

actually it just tries to access an unknown memory location, the data segment is, in the example I've added, 31 bytes long for so it was failing in the `bytes32_to_uint256`

Issue #568, opened 2023-04-12
bug: deploy_and_fund_evm_address fails on devnet when EVM_ADDRESS is set


bajpai244:
# Bug Report

**Kakarot version: 0.1.0**

**Current behavior:**

When we are on devnet, and set the **EVM_ADDRESS** variable, we get the following error.

```shell
Cannot mint token to 3162353902975168151163452425762749780210716163200473565734733406023685411743: {"code":"StarkErrorCode.MALFORMED_REQUEST","message":"Hash should be a hexadecimal string starting with 0x, or 'null'; got: '3162353902975168151163452425762749780210716163200473565734733406023685411743'."}
```

This happens after a call to **deploy_and_fund_evm_address** is made in [deploy_kakarot.py](https://github.com/sayajin-labs/kakarot/blob/main/scripts/deploy_kakarot.py).

**Expected behavior:**

The deployment should produced, and the EOA should get funded.

**Steps to reproduce:**

- set the EVM_ADDRESS environment variable 
- set network to devnet
- run starknet-devnet
- run deploy_kakarot.py { make deploy }

PR #569, opened 2023-04-12
fix(fund_address): fix address to be hex when network is devnet

bajpai244:
Fix the address to be hex inside the fund_address function when the input is of type int and the network is devnet.

## Pull request type: fix

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Upon investigating #568 , I found that the following function calls are involved:
- deploy_and_fund_evm_address 
- fund_address

When we investigate fund_address, the following code snippet is where the error comes from:

```python
 address = int(address, 16) if isinstance(address, str) else address
    amount = amount * 1e18
    if NETWORK == "devnet":
        response = requests.post(
            f"{GATEWAY_CLIENT.net}/mint", json={"address": address, "amount": amount}
        )
        if response.status_code != 200:
            logger.error(f"Cannot mint token to {address}: {response.text}")
        logger.info(f"{amount / 1e18} ETH minted to {address}")
```

the address is expected to be a hex string, and since it is not we get the error described in #568.

Resolves <https://github.com/sayajin-labs/kakarot/issues/568>

## What is the new behavior?

In the new behavior, if the type is *int*, then we call **hex()** to it, to convert it into a hex string.

### DIFF
--- a/scripts/utils/starknet.py
+++ b/scripts/utils/starknet.py
@@ -202,7 +202,7 @@ async def fund_address(address: Union[int, str], amount: float):
     amount = amount * 1e18
     if NETWORK == "devnet":
         response = requests.post(
-            f"{GATEWAY_CLIENT.net}/mint", json={"address": address, "amount": amount}
+            f"{GATEWAY_CLIENT.net}/mint", json={"address": hex(address), "amount": amount}
         )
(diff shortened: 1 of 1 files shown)


ClementWalter (approved):
thanks!

PR #572, opened 2023-04-13
fix(#571): fix errors in readme and .env.example, issue #571

### LINKED ISSUE #571: bug: documentation and .env.example has errors
# Bug Report

**Kakarot version:0.1.0**

**Current behavior:**

The documentation has errors that pop up when you try to run the project, listing them individually.

### `PRIVATE_KEY` should be hexadecimal

The documentation says that this environment variable should be decimal but it should actually be hexadecimal.

image

### Default value of `STARKNET_NETWORK` should be devnet

If you look at the current [.env.example](https://github.com/sayajin-labs/kakarot/blob/main/.env_example) file, you will notice that the default value for `STARKNET_NETWORK `is **testnet**, but according to the documentation it should be **devnet**.

image

### Specify the need for `GITHUB_TOKEN`

The current documentation can include that the user must have a GitHub token with read access to build the system, and hence needs to populate `GITHUB_TOKEN` environment variable.

**Expected behavior:**

These things should be corrected so that a new user can find it easy to deploy Kakarot.

bajpai244:
## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

There are inconsistencies in the **README** and the **.env.example** file as described in #571.

Resolves #571 .

## What is the new behavior?

It resolves all issues of #571 .

## Note

The updated keypair in the documentation is a valid keypair automatically generated from starknet-devnet.

### DIFF
--- a/.env_example
+++ b/.env_example
@@ -2,7 +2,7 @@
 GITHUB_TOKEN=
 
 # Network configuration
-STARKNET_NETWORK=testnet
+STARKNET_NETWORK=devnet
 
 # All addresses are hex strings
 MAINNET_ACCOUNT_ADDRESS=
--- a/README.md
+++ b/README.md
@@ -234,6 +234,14 @@ Rename the copied file to `.env`.
 The file holds the following content:
 
 ```text
+GITHUB_TOKEN=your_github_token
+```
+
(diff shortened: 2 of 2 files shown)


Eikix (approved):
lgtm, ✅
thanks for the PR 🤙

PR #591, opened 2023-06-15
Updates to work with madara

ClementWalter:
Time spent on this PR:

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Not working on madara:

- the default account at 0x1 doesn't implement the standard `__execute__`
- `addDeclare` doesn't "return" anything when the class already exist (Transaction not found)

## What is the new behavior?

Deploy script works with `STARKNET_NETWORK=madara`.

## Other information

Madara has the same issue as Katana:

- events not emitted
- not able to deploy ERC20 (probably out of resources error ? raises with `An ASSERT_EQ instruction failed: 340282366920938463463374607431768211424 != 340282366920938463463374607431768211455.`)


scripts/utils/starknet.py
```diff
@@ -149,12 +158,40 @@ async def fund_address(address: Union[int, str], amount: float):
             raise ValueError(
                 f"Cannot send {amount / 1e18} ETH from default account with current balance {balance / 1e18} ETH"
             )
-        tx = await eth_contract.functions["transfer"].invoke(
-            address, int_to_uint256(amount), max_fee=int(1e17)
+        prepared = eth_contract.functions["transfer"].prepare(
+            address, int_to_uint256(amount)
         )
+        if NETWORK == "madara" and account.address == 1:
```
Eikix, line 164: add todo to remove for when Madara processes standard calldata & call array?

scripts/utils/starknet.py
```diff
@@ -246,12 +283,63 @@ def _convert_offset_to_hex(obj):
...
+    logger.info(f"ℹ️  Deploying account")
+    res = await Account.deploy_account(
+        address=address,
+        class_hash=class_hash,
+        salt=salt,
+        key_pair=key_pair,
+        client=RPC_CLIENT,
+        chain=CHAIN_ID,
+        constructor_calldata=constructor_calldata,
+        max_fee=int(1e17),
```
Eikix, line 318: should we once and for all create a MAX_FEE const? and use it everywhere?

scripts/utils/starknet.py
```diff
@@ -246,12 +283,63 @@ def _convert_offset_to_hex(obj):
...
+
 async def declare(contract_name):
     logger.info(f"ℹ️  Declaring {contract_name}")
-    account = await get_starknet_account()
     artifact = get_artifact(contract_name)
     compiled_contract = Path(artifact).read_text()
     contract_class = create_compiled_contract(compiled_contract=compiled_contract)
+    class_hash = compute_class_hash(contract_class=deepcopy(contract_class))
+    try:
+        await RPC_CLIENT.get_class_by_hash(class_hash)
```
Eikix, line 337: it is in this moment that I miss Result<> type to handle this, but anyway^^
i'm starting to hate the try catch style of javascript and python

src/kakarot/accounts/library.cairo
```diff
@@ -25,6 +25,14 @@ from kakarot.interfaces.interfaces import IAccount
 func nonce() -> (nonce: felt) {
 }
 
+@storage_var
```
Eikix, line 28: I understand this is to work around events, should we add TODOs to delete them after event fix?

scripts/deploy_kakarot.py
```diff
@@ -32,6 +34,8 @@ async def main():
         f"ℹ️  Connected to CHAIN_ID {CHAIN_ID.value.to_bytes(ceil(log(CHAIN_ID.value, 256)), 'big')} "
         f"with RPC {RPC_CLIENT.url}"
     )
+    if NETWORK == "madara":
```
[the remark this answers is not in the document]
ClementWalter: it's because the only pre-deployed account is non standard, this may change in the future (actually I could make a PR to madara), but for now it's the only way I've found

scripts/utils/starknet.py
ClementWalter: possibly. tbh I don't remember why we don't use the fee estimation step and set max fee instead, I can try

src/kakarot/accounts/library.cairo
ClementWalter: we should! I was pretty sure to have some added, but apparently not

scripts/utils/starknet.py
ClementWalter: but also I had a tx failing with max_fee = 0.1 an not with 0.5, while the actual fee was much lower so the fee estimation is not really accurate. We'll need to clarify this when we launch mainnet (wen mainnet) but up until this point, these random hard coded values look ok to me we just to make sure that it passes

Eikix: alright

Eikix (approved):
lgtm

Issue #575, opened 2023-04-18
feat: Build a multi-arch devnet image in the CI


ClementWalter:
# Feature Request

**Describe the Feature Request**

Currently, the devnet image is not working well on mac m1, due to its x86 arch.

**Describe Preferred Solution**

Use Docker manifest to create a multi-arch image, see:
- https://docs.docker.com/engine/reference/commandline/manifest/#manifest-annotate
- https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry#pushing-container-images
- https://docs.docker.com/build/ci/github-actions/multi-platform/

**Related Code**

- [deployments.yml](../blob/main/.github/workflows/deployments.yml)
- [devnet](../blob/main/docker/devnet)

**Additional Context**

**If the feature request is approved, would you be willing to submit a PR?**
Yes / No _(Help can be provided if you need assistance submitting a PR)_

bajpai244:
@Eikix I guess we can close this, as now we do have a multi-arch image for our devnet.

PR #594, opened 2023-06-19
Feat/add katana docker container

### LINKED ISSUE #589: feat: Use katana instead of the starknet-devnet in the kakarot/devnet image
# Feature Request

Currently, the [kakarot devnet image](https://github.com/ClementWalter/kakarot/blob/f08d5ba045c40388af5bc70012eb4c99a117447c/docker/devnet/Dockerfile#L1) is using the starknet-devnet.

We'd like to use katana instead, and to make it multi-arch

In the Dockerfile, install katana instead, and find a way to store state or run deployment script in the command.
Make it multi-arch

bajpai244:
The PR adds a Dockerfile to build a katana image with Kakarot already deployed on it. The image is multi-arch and the pr also provides integration of its build and publishing in the current github workflow.

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

We don't have a docker image of Katana with Kakarot already deployed on it.

Resolves #589 #575 

## What is the new behavior?

provides a multi-arch docker image of katana with kakarot already deployed on it.

## Other information

The containers of these images will also have a healthcheck, the containers will become healthy, hence for testing, or building on top of these containers, the health can be inspected.

### DIFF
--- a/docker/katana/Dockerfile
+++ b/docker/katana/Dockerfile
@@ -1,0 +1,46 @@
+FROM --platform=$BUILDPLATFORM rust:1.70 as katana
+
+RUN apt-get update && apt-get install -y git
+
+WORKDIR /app
+
+RUN git clone https://github.com/dojoengine/dojo.git
+
+WORKDIR /app/dojo
+
+RUN cargo build --package katana --release
+
+FROM python:3.9.13-slim-bullseye 
+
+HEALTHCHECK --interval=30s --timeout=30s --start-period=5s --retries=3 \
+    CMD if [ "$(cat /root/.kakarot_deploy_status)" -eq 0 ]; then exit 0; else exit 1; fi
+
+COPY --from=katana /app/dojo/target/release/katana /app/katana
+COPY ./docker/katana/scripts /app/scripts
+
+WORKDIR /app
+
+RUN apt-get update && apt-get install -y build-essential coreutils curl libgmp3-dev
+
+RUN chmod +x scripts/run.sh
+RUN chmod +x scripts/wait_and_deploy.sh
+
+COPY ./docker/katana/cairo/requirements.txt ./requirements.txt
+
+RUN pip3 install -r requirements.txt
+
+# install poetry
+RUN curl -sSL https://install.python-poetry.org | python3 -
+
+ENV PATH="$PATH:/root/.local/bin"
+
+COPY . ./kakarot
+WORKDIR /app/kakarot
+
+RUN poetry install
+RUN make build
+
+EXPOSE 5050
+
+WORKDIR /app/scripts
+CMD [ "./run.sh" ]
--- a/docker/katana/Dockerfile
+++ b/docker/katana/Dockerfile
@@ -28,7 +28,6 @@ HEALTHCHECK --interval=30s --timeout=30s --start-period=5s --retries=3 \
 COPY --from=katana-builder /app/dojo/target/x86_64-unknown-linux-gnu/release/katana  /app/katana/x86_64-unknown-linux-gnu/release/katana 
 COPY --from=katana-builder /app/dojo/target/aarch64-unknown-linux-gnu/release/katana /app/katana/aarch64-unknown-linux-gnu/release/katana 
 COPY ./docker/katana/scripts /app/scripts
-COPY . /app/kakarot
 
 WORKDIR /app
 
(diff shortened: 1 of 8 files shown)


docker/katana/Dockerfile
```diff
@@ -0,0 +1,58 @@
...
+WORKDIR /app
+
+RUN apt-get update && apt-get install -y build-essential coreutils curl libgmp3-dev
+
+RUN chmod +x scripts/run.sh
+RUN chmod +x scripts/wait_and_deploy.sh
+
+COPY ./docker/katana/cairo/requirements.txt ./requirements.txt
+
+RUN pip3 install -r requirements.txt
```
ClementWalter, line 42: what's the point of this `requirements.txt` file? how has it been generated?

docker/katana/Dockerfile
```diff
@@ -0,0 +1,58 @@
...
+
+# install poetry
+RUN curl -sSL https://install.python-poetry.org | python3 -
+
+ENV PATH="$PATH:/root/.local/bin"
+
+COPY . ./kakarot
```
ClementWalter, line 49: it's usually better to first copy the requirements only (here the pyproject.toml and the poetry.lock) to leverage caching in `docker build`.

Then, copy the whole project, otherwise each modified file will invalidate the (usually long) install deps step.

docker/katana/cairo/requirements.txt
ClementWalter, line 1: what is this file for?

docker/katana/scripts/wait_and_deploy.sh
```diff
@@ -0,0 +1,40 @@
...
+  done
+
+  echo "curl request failed after retries. Exiting..."
+  return 1
+}
+
+check_katana_status || exit 1
+
+cd ../kakarot
+poetry shell
```
ClementWalter, line 29: You don't need `poetry shell` , `poetry run ...` is enough

kkrt.log
ClementWalter, line 1: what is this file?

ClementWalter (changes requested):
thanks! few minor things plz

docker/katana/Dockerfile
bajpai244: I was looking for inspiration for a cairo container, checking if someone has already worked on it before, and found one implementation [here](https://github.com/shramee/starknet-dapp-with-docker/tree/master/python-backend), took a look at the packages being installed there.

Then I double-checked it by looking into the Starknet book [here](https://book.starknet.io/chapter_1/environment_setup.html#starknet_cli_installation), and only kept the ones I saw were necessary, which are the current present ones in the file.

Adding an image here too for more context:
image

docker/katana/Dockerfile
bajpai244: Makes sense! Thanks for this info.

I have made the change [here.]([feat: group all COPY togther in the Dockerfile](https://github.com/kkrt-labs/kakarot/pull/594/commits/9bc4e11c247626cac56198c657247b7b0994934d))

docker/katana/cairo/requirements.txt
bajpai244: Answered [here].(https://github.com/kkrt-labs/kakarot/pull/594#discussion_r1234149546)

docker/katana/scripts/wait_and_deploy.sh
bajpai244: Gotcha, done [here](https://github.com/kkrt-labs/kakarot/pull/594/commits/b344db259ffdf543e7f36319b4efb9374197dd30).

kkrt.log
bajpai244: committed by mistake, thanks for catching this, removed it [here](https://github.com/kkrt-labs/kakarot/pull/594/commits/8120c9195edfef187928575c86851e90113f54ee).

docker/katana/Dockerfile
ClementWalter: if we install the whole project (what we do), we don't need extra lib

docker/katana/Dockerfile
ClementWalter: maybe it wasn't clear. You need to 

```
# first only the deps files
COPY poetry.lock poetry.lock
COPY pyproject.toml pyproject.toml

poetry install

# then the whole project
COPY . .
```

ClementWalter (changes requested):
plz remove the `requirements.txt`, should not be necessary

bajpai244:
> plz remove the `requirements.txt`, should not be necessary

Done [here.](https://github.com/kkrt-labs/kakarot/pull/594/commits/c8c960eba543cf861124467f8ee5fbdb124c0ab5)

docker/katana/cairo/requirements.txt
bajpai244: done [here](https://github.com/kkrt-labs/kakarot/pull/594/commits/c8c960eba543cf861124467f8ee5fbdb124c0ab5).

docker/katana/Dockerfile
bajpai244: Gotcha, thanks!

have removed [here](https://github.com/kkrt-labs/kakarot/pull/594/commits/c8c960eba543cf861124467f8ee5fbdb124c0ab5).

docker/katana/Dockerfile
bajpai244: Got it! 

Thanks, I have implemented [here](https://github.com/kkrt-labs/kakarot/pull/594/commits/c8c960eba543cf861124467f8ee5fbdb124c0ab5).

ClementWalter (dismissed):

ClementWalter (approved):

PR #598, opened 2023-06-21
Refacto chain definition

ClementWalter:
Time spent on this PR: 0.5 days

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

In the scripts/constants.py file, there are a bunch of parameters that need to be updated whenever we need to add a new chain.

## What is the new behavior?

The const definition has been updated to instead be a list of chain, each with all their parameters.

### DIFF
--- a/scripts/constants.py
+++ b/scripts/constants.py
@@ -69,10 +69,19 @@ NETWORKS = {
 NETWORK = NETWORKS[os.getenv("STARKNET_NETWORK", "devnet")]
 NETWORK["account_address"] = os.environ.get(
     f"{NETWORK['name'].upper()}_ACCOUNT_ADDRESS"
-) or os.environ.get("ACCOUNT_ADDRESS")
-NETWORK["private_key"] = os.environ.get(
-    f"{NETWORK['name'].upper()}_PRIVATE_KEY"
-) or os.environ.get("PRIVATE_KEY")
+)
+if NETWORK["account_address"] is None:
+    logger.warning(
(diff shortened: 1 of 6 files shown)


scripts/constants.py
```diff
@@ -1,97 +1,106 @@
...
-ACCOUNT_ADDRESS = os.environ.get(
-    f"{NETWORK.upper()}_ACCOUNT_ADDRESS"
+NETWORK = NETWORKS[os.getenv("STARKNET_NETWORK", "devnet")]
+NETWORK["account_address"] = os.environ.get(
+    f"{NETWORK['name'].upper()}_ACCOUNT_ADDRESS"
 ) or os.environ.get("ACCOUNT_ADDRESS")
-PRIVATE_KEY = os.environ.get(f"{NETWORK.upper()}_PRIVATE_KEY") or os.environ.get(
-    "PRIVATE_KEY"
-)
+NETWORK["private_key"] = os.environ.get(
```
jobez, line 73: might be good to add some logging here "no specific network private key set, defaulting to private key" 

this would handle the user error that might happen when someone wants to try deploying to a new network and is in an environment with a previous private key meant for a different network

also another option is to be strict and to throw an error if they arent specifically setting a private key specific to the network in an env file.

jobez (changes requested):

scripts/constants.py
ClementWalter: I'd go with warning only and non breaking because it's possible to use the same params for different chains

PR #601, opened 2023-06-23
event key is appended w/ `evm_contract_address`

### LINKED ISSUE #600: bug: eth event emission should be connectable to the eth emitting contract
# Bug Report

**Kakarot version:**

**Current behavior:**

Event emission was changed for the revert opcode: https://github.com/kkrt-labs/kakarot/blob/c3c8d802aeec98d064cb2ef855e1247933bfe9fa/src/kakarot/execution_context.cairo#L317

The reason for this change was to ensure that events are only fired in an execution context that ran successfully.

**Expected behavior:**

From the kkrt-rpc pov, we want the ability to relate the event to the contract that emits them, see: https://github.com/kkrt-labs/kakarot-rpc/issues/160.

Usually this would be via mapping the starknet event `from_address`.
This would be the starknet address that implements the ethereum contract that fires the event. The ethereum address would be derivable from `get_evm_address` of the implementation.

**Other information:**

- we can achieve a fix either via
   - adding an `emit_events` method to the contract interface that kakarot calls upon the finalization of an execution context
   - adding the evm address in the event keys/data of the event itself
   - see https://github.com/kkrt-labs/kakarot-rpc/pull/174#discussion_r1237868617 for the discussion

jobez:
Time spent on this PR: .5

## Pull request type

- [ X] Bugfix
- [ ] Feature
- [ ] Build related changes

Event emission is consolidated in the kkrt contract. This gives us an ability to filter out kkrt events, but we also want to be able to associate the kkrt event w/ kkrt contract.

Resolves #600 

## What is the new behavior?

- the `evm_contract_address` is appended to the event keys of every emitted event
- we test that each event log's address is equal to the address that a contract is mutated w/ at deploy time

### DIFF
--- a/src/kakarot/execution_context.cairo
+++ b/src/kakarot/execution_context.cairo
@@ -312,6 +312,7 @@ namespace ExecutionContext {
     }
 
     // @notice Iterates through a list of events and emits them on the case that a context ran successfully (stopped and not reverted).
+    // @param evm_contract_address The execution context's evm contract address.
     // @param events_len The length of the events array.
     // @param events The array of Event structs that are emitted via the `emit_event` syscall.
     func emit_events{
@@ -319,7 +320,7 @@ namespace ExecutionContext {
         pedersen_ptr: HashBuiltin*,
         range_check_ptr,
         bitwise_ptr: BitwiseBuiltin*,
-    }(events_len: felt, events: model.Event*) {
+    }(evm_contract_address: felt, events_len: felt, events: model.Event*) {
         alloc_locals;
 
         if (events_len == 0) {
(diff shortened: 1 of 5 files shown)


src/kakarot/execution_context.cairo
```diff
@@ -319,20 +319,30 @@ namespace ExecutionContext {
         pedersen_ptr: HashBuiltin*,
         range_check_ptr,
         bitwise_ptr: BitwiseBuiltin*,
-    }(events_len: felt, events: model.Event*) {
+    }(evm_contract_address: felt, events_len: felt, events: model.Event*) {
```
ClementWalter, line 322: update docstring stp

ClementWalter (approved):

ClementWalter (approved):

tests/integration/test_kakarot.py
```diff
@@ -87,10 +87,11 @@ async def test_execute(
         assert memory_result == hex_string_to_bytes_array(params["memory"])
 
         events = params.get("events")
+        event_keys_without_evm_contract_address = events.keys[:-1]
```
ClementWalter, line 90: just use `events.keys[:-1]` in the `assert`

src/kakarot/execution_context.cairo
```diff
@@ -364,7 +375,7 @@ namespace ExecutionContext {
             // this is called after a top level check that a given context is stopped
             // so this is the case of a stopped, non reverted context
             // meaning events should be fired off!
-            emit_events(self.events_len, self.events);
+            emit_events(self.evm_contract_address, self.events_len, self.events);
```
Eikix, line 378: this makes me wonder abt smth, let me give an example:
context A has subcontext B,

context A starts...,  then subcontext B is entered,

when context B terminates successfully at time t,

then context A resumes, and then reverts.

are subcontext B events emitted at time t?

or never emitted, since the overall call reverted?

jobez: i think in this form they would still be emitted. 

can add 
- a check where emit_event doesnt get called unless its the finalized root calling context
   - if root, go through events and fire them
   - if not, push the event w/ the current execution context to the events of the calling stack

Eikix: I think it'll help, right?

Eikix (approved):
lgtm

PR #602, opened 2023-06-23
Minor updates to deploy scripts after latest sharingan release

ClementWalter:
Time spent on this PR: 0.25

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Two unrelated topics:

1. the chain_id enum has some hard-coded values that needs to be maintained
1. the deploy scripts first deploy an OZ account in madara and sharingan. This was done because no standard account was available on these networks

## What is the new behavior?

Two unrelated updates:

1. the chain_id is now taken from the network itself so it's always the right one
1. no more initial deployment of an account in madara and sharingan because an OZ account has been added in the genesis state. The util is not removed though because it is useful to have it ready in case we need to create on the fly several accounts.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -8,7 +8,7 @@ SHARINGAN_RPC_URL=
 
 # Account used by default in all scripts
 # Prefixed accounts {NETWORK}_ will take precedeance over the default one
-MADARA_ACCOUNT_ADDRESS=0x1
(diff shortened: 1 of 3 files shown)


jobez (approved):
lgtm

PR #603, opened 2023-06-27
Bug/events should be linkable to emitting contract

Eikix:
copy of #601

### DIFF
--- a/scripts/deploy_kakarot.py
+++ b/scripts/deploy_kakarot.py
@@ -2,11 +2,7 @@
 import logging
 from asyncio import run
 
-from scripts.constants import (
-    COMPILED_CONTRACTS,
-    ETH_TOKEN_ADDRESS,
-    EVM_ADDRESS,
-)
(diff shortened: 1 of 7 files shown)


scripts/deploy_kakarot.py
```diff
@@ -2,11 +2,7 @@
 import logging
 from asyncio import run
 
-from scripts.constants import (
-    COMPILED_CONTRACTS,
-    ETH_TOKEN_ADDRESS,
-    EVM_ADDRESS,
-)
+from scripts.constants import COMPILED_CONTRACTS, ETH_TOKEN_ADDRESS, EVM_ADDRESS
```
jobez, line 5: is there a reason why this formatting is only picked up now? just curious

jobez (approved):

scripts/deploy_kakarot.py
Eikix: i think it's the commit I added, the push hook wasn't working due to us changing the name of the org

PR #627, opened 2023-07-07
feat: add bytecode test for kakarot

bajpai244:
# Draft preview

Currently, on devnet, not all instances of these tests are passing. We should investigate them.

## How to run this?
- `cd integration-tests`
- `cp .env.example` `.env`
-  add your infura key 
-  `cargo test -- --nocapture`

You will see logs on which of these test cases are passing and which are failing.
image

Note: 

- Currently, we are just checking for the result to be Ok() as a metric of success for these tests.
- The kakarot_address is the address of the fixture that exposes the `execute` view, [here](https://github.com/kkrt-labs/kakarot/blob/main/tests/fixtures/EVM.cairo) is the source file.


integration-tests/src/main.rs
```diff
@@ -0,0 +1,3 @@
+fn main () {
```
danilowhk, line 1: Do we need this function?

integration-tests/tests/bytecode_test.rs
```diff
@@ -0,0 +1,66 @@
...
+    let starknet_config = StarknetConfig::from_env().unwrap();
+    let starknet_provider = JsonRpcClient::new(HttpTransport::new(starknet_config.starknet_rpc));
+
+    let kakarot_client = KakarotClient::new(
+        starknet_provider,
+        starknet_config.kakarot_address,
+        starknet_config.proxy_account_class_hash
+    );
+
+    let file = match File::open("test_cases.json") {
```
danilowhk, line 25: Maybe we would like to have the file directory as an .env variable, instead of hard coding?

integration-tests/tests/helpers/client.rs
```diff
@@ -0,0 +1,67 @@
...
+        kakarot_calldata: Vec<FieldElement>,
+    ) -> Result<Vec<FieldElement>, Box<dyn Error>> {
+        let kakarot_calldata_len = FieldElement::from(kakarot_calldata.len());
+        let bytecode_len = FieldElement::from(bytecode.len());
+
+        let mut execute_arguments = vec![];
+        execute_arguments.push(value); //value
+
+        execute_arguments.push(bytecode_len); //bytecode_len
+        for bytecode_bytes in bytecode {
```
danilowhk, line 40: in this case could be "bytecode_byte" as each element in the vector would be 1 byte and not several

integration-tests/src/main.rs
bajpai244: nope, removed.

integration-tests/tests/bytecode_test.rs
bajpai244: I think this file should be organized in a `data` directory and not be passed via the environment, as we might have other json files in the future.

I am putting them in a `data` directory inside `tests`, but let me know if you have other thoughts.

integration-tests/tests/helpers/client.rs
bajpai244: nice catch! corrected :)

Eikix:
What is the status of this PR, should we close it @bajpai244 ?

bajpai244:
> 

@Eikix we should close this, I don't think after the ef-tests this is relevant.

PR #638, opened 2023-07-14
line up deterministic addressing for the zero nonce case

### LINKED ISSUE #637: bug: Deployed addresses are not the expected one
# Bug Report

**Current behavior:**

When deploying a contract to kakarot, the returned address is not the expected one.

See this output from `forge`:

```
// Expected addresses
[984601] → new Bytes@0x5FbDB2315678afecb367f032d93F642f64180aa3
[146071] → new MultiSendCallOnly@0xe7f1725E7734CE288F8367e1Bb143E90bb3F0512
[2495543] → new BasicCellRenderer@0x9fE46736679d2D9a65F0992F2272dE9f3c7fa6e0
[208382] → new Math@0xCf7Ed3AccA5a467e9e704C703E8D87F634fB0Fc9
[2659170] → new Evmsheet@0xDc64a140Aa3E981100a9becA4E685f962f0cF6C9

// Real addresses
Contract Address: 0xe7f1725E7734CE288F8367e1Bb143E90bb3F0512
Contract Address: 0x9fE46736679d2D9a65F0992F2272dE9f3c7fa6e0
Contract Address: 0xCf7Ed3AccA5a467e9e704C703E8D87F634fB0Fc9
Contract Address: 0xDc64a140Aa3E981100a9becA4E685f962f0cF6C9
Contract Address: 0x5FC8d32690cc91D4c39d9d3abcBD16989F875707
```

In other words, the addresses are just shifted by one, which probably means that we use increase the nonce before instead of after the tx.

**Expected behavior:**

Output = expected.

**Steps to reproduce:**

This is the log when deploying kakasheet

jobez:
Time spent on this PR:

## Pull request type

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes

We currently increment account nonce before create / execution. 

I also added the first fixture address to line up with anvil, to make it easier to diff results.

Resolves #637 

## What is the new behavior?

- Now we do so after, along with fixing a subtle bug in how we were encoding the case of the nonce being zero to rlp, see: https://ethereum.stackexchange.com/posts/61021/revisions
-
-


ClementWalter (approved):
so we never add nonce=0 before, and never saw that rlp encoding was wrong for 0, interesting

PR #641, opened 2023-07-15
Minor devX improvements

ClementWalter:
Time spent on this PR: 0.1

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

`chain_id` is fetch from the given RPC endpoint. Even though we now use the Gateway for mainnet/testnet/testnet2,
we then still need an RPC endpoint just for this.

`fetch_deployments` is called when network is not `devnet` and raises if no deployment is found. This is not
really necessary and was done so when we used only devnet/testnet/mainnet.

## What is the new behavior?

`chain_id` is prefilled for testnet/testnet2/mainnet
`try/catch` around `fetch_deployments`


ftupas (approved):
LGTM!

PR #649, opened 2023-07-20
feat: remove nonce management, use starknet  nonce

### LINKED ISSUE #648: feat: remove nonce management from contracts, and rely on starknet nonce
# Feature Request

We as of now are [managing nonce](https://github.com/kkrt-labs/kakarot/blob/dcc93a78d9126c0101af739d6aebf4476514a62e/src/kakarot/accounts/library.cairo#L25) ourselves for EOAs and Contracts, this causes a lot of issues, for example if you flood the mempool of the sequencer with a lot of transactions, only 1 of them will pass and all following will fail on validation, as we check the nonce for validation.

We have to remove management of nonce, and rely on starknet nonce which we should be able to obtain via [`get_tx_info`](https://docs.starknet.io/documentation/architecture_and_concepts/Contracts/system-calls-cairo0/#get_transaction_info) syscall. 

Yes

bajpai244:
This pr removes nonce management and uses starknet nonce wherever nonce is required.

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

We currently manage nonce ourselves which causes multiple issues, migrating to starknet nonce is a better solution and the PR implements the required changes for that.

Resolves #648 

## What is the new behavior?

- nonce management has been removed
- we now use starknet nonce wherever required

## Other information

- The rpc is already [using starknet nonce](https://github.com/kkrt-labs/kakarot-rpc/blob/d6e3676de6fc9e804dc751aa3493810c95957bba/crates/core/src/client/mod.rs#L338), which means that the current transactions by tools like metamask are already being signed with starknet nonce and not our own nonce when RPC is being used.

### DIFF
--- a/tests/src/kakarot/instructions/test_system_operations.py
+++ b/tests/src/kakarot/instructions/test_system_operations.py
@@ -87,8 +87,8 @@ class TestSystemOperations:
             system_operations.contract_address
         )
 
-    @pytest.mark.parametrize("salt", [0, 127, 256, 2**55 - 1])
-    async def test_create(self, system_operations, salt):
+    async def test_create(self, system_operations):
+        salt = 0
         # given we start with the first anvil test account
         evm_caller_address_int = 0xF39FD6E51AAD88F6F4CE6AB8827279CFFFB92266
         evm_caller_address_bytes = evm_caller_address_int.to_bytes(20, byteorder="big")
--- a/src/kakarot/accounts/contract/contract_account.cairo
+++ b/src/kakarot/accounts/contract/contract_account.cairo
@@ -88,21 +88,3 @@ func is_initialized{
 }() -> (is_initialized: felt) {
     return ContractAccount.is_initialized();
 }
-
-// @notice This function is used to read the nonce from storage
-// @return nonce: The current nonce of the contract account
-@view
-func get_nonce{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr}() -> (nonce: felt) {
-    return Accounts.get_nonce();
-}
-
-// @notice This function increases the contract accounts nonce by 1
-// @return nonce: The new nonce of the contract account
-@external
-func increment_nonce{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr}() -> (
-    nonce: felt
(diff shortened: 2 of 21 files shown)


ClementWalter (dismissed):

ClementWalter (approved):

tests/src/kakarot/instructions/test_system_operations.py
```diff
@@ -87,7 +87,6 @@ async def test_call__should_transfer_value(self, system_operations, mint):
             system_operations.contract_address
         )
 
-    @pytest.mark.parametrize("salt", [0, 127, 256, 2**55 - 1])
     async def test_create(self, system_operations, salt):
```
ClementWalter, line 90: why don't you keep the `salt`?

ClementWalter: and it's become undefined in the body I think

ClementWalter (commented):
If you feel so, you can create fixtures to name used addresses once for all, and avoid magic numbers, you can put them at the top of the file

tests/src/kakarot/instructions/test_system_operations.py
bajpai244: @ClementWalter I can keep this parameterized as well, for now I just replaced the salt with 0, but I guess I can bump it back to being parameterized if you want.

More about the issue can be found here -> https://github.com/kkrt-labs/kakarot-rpc/blob/b9fcd0c574ef4aa06dd41dbc1150903154ff6a7a/crates/core/src/client/mod.rs#L412

Note: as mentioned in the above issues, @jobez will later on refactor these tests after this PR merge.

bajpai244:
> If you feel so, you can create fixtures to name used addresses once for all, and avoid magic numbers, you can put them at the top of the file

Makes sense, lemme do that!

tests/src/kakarot/instructions/test_system_operations.py
ClementWalter: just need to remove the arg I guess.

ClementWalter (dismissed):

ClementWalter (approved):

ClementWalter (approved):

PR #659, opened 2023-07-25
Update starknet utils to latest

ClementWalter:
Time spent on this PR: 0.1

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

The project relies on a `scripts/utils/starknet.py` set of tools for interacting with Starknet(-like) network(s).

This module is indeed a "hidden package" originated from Starksheet devX and shared in different projects.

## What is the new behavior?

This PR just updates it to the latest improvements brought elsewhere.

## Other information

I've never considered making this a real python packaging because it always introduces some burden
when it comes to maintainability and easy adaption to this ever changing dev environment.
Furthermore, the gold standard will soon be to use `starkli` so I'm not sure it's worth doing it rn.

That being said, if reviewers feels like it's a good idea, I can consider it!

### DIFF
--- a/scripts/utils/starknet.py
+++ b/scripts/utils/starknet.py
@@ -434,7 +434,14 @@ async def invoke_contract(
     return await account.execute(call, max_fee=_max_fee)
 
 
-async def invoke(contract, *args, **kwargs):
+async def invoke(contract: Union[str, int], *args, **kwargs):
+    """
+    Invoke a contract specified:
+     - either with a name (expect that a matching ABIs is to be found in the project artifacts)
+       `invoke("MyContract", "foo")`
+     - or with a plain address (in this later case, no parsing is done on the calldata)
+       `invoke(0x1234, "foo")`
+    """
     response = await (
         invoke_address(contract, *args, **kwargs)
         if isinstance(contract, int)
@@ -472,7 +479,14 @@ async def call_contract(contract_name, function_name, *inputs, address=None):
     return await contract.functions[function_name].call(*inputs)
 
 
-async def call(contract, *args, **kwargs):
+async def call(contract: Union[str, int], *args, **kwargs):
+    """
+    Call a contract specified:
+     - either with a name (expect that a matching ABIs is to be found in the project artifacts)
+     `call("MyContract", "foo")`
+     - or with a plain address (in this later case, no parsing is done on the calldata)
+     `call(0x1234, "foo")`
+    """
     return await (
         call_address(contract, *args, **kwargs)
         if isinstance(contract, int)
--- a/scripts/utils/starknet.py
+++ b/scripts/utils/starknet.py
@@ -7,7 +7,7 @@ import time
 from copy import deepcopy
 from datetime import datetime
 from pathlib import Path
-from typing import Union, cast
+from typing import List, Union, cast
 
 import requests
 from caseconverter import snakecase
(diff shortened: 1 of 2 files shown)


Eikix:
I don't understand in which case invoke does not target a contract, since everything is a contract in starknet 

is it: call invoke contract if contract is an instantiated class?
otherwise invoke address?

if else statement may be reversed then?

ClementWalter:
> I don't understand in which case invoke does not target a contract, since everything is a contract in starknet
> 
> is it: call invoke contract if contract is an instantiated class? otherwise invoke address?
> 
> if else statement may be reversed then?

I don't understand this!

you do 

```
invoke("my_contract", "foo")
# => invoke_contract (use contract abi)
# raise if it doesn't find a deployed "my_contract" contract in `deployments.json`

invoke("my_contract", "foo", address=0x1234)
# => invoke_contract (use contract abi)
# use the provided address

invoke(0x1234, "foo")
# => invoke_address (raw call)
```

same for `call`

scripts/utils/starknet.py
```diff
@@ -201,7 +206,10 @@ def dump_deployments(deployments):
 
 
 def get_deployments():
-    return json.load(open(DEPLOYMENTS_DIR / "deployments.json", "r"))
+    try:
+        return json.load(open(DEPLOYMENTS_DIR / "deployments.json", "r"))
+    except FileNotFoundError:
+        return {}
```
ftupas, line 212: Is this for when the devnet is dockerized?

scripts/utils/starknet.py
```diff
@@ -424,6 +472,14 @@ async def call(contract_name, function_name, *inputs, address=None):
     return await contract.functions[function_name].call(*inputs)
 
 
+async def call(contract, *args, **kwargs):
+    return await (
+        call_address(contract, *args, **kwargs)
+        if isinstance(contract, int)
+        else call_contract(contract, *args, **kwargs)
+    )
```
ftupas, line 480: Just to see if I understand this right, if the deployments exist, we load the abi and call from `Contract` object, otherwise we take the address and do a raw call using `call_contract`?

Eikix: I think it's the opposite:
```
 call_address(contract, *args, **kwargs)
        if isinstance(contract, int)
```

means -> do a raw call if `contract` is an integer, i.e. not a contract name (string), as integers represent addresses in cairo
-> else do a contract call

Eikix (dismissed):
lgtm!

scripts/utils/starknet.py
Eikix: the naming is a bit misleading -> 
could be commented above that contract (the first var) is either a contract name or a contract address

ftupas: Ah sorry I meant a lower level `call_contract` https://github.com/kkrt-labs/kakarot/pull/659/files#diff-0da6276ff678d86aaa4359903ad0e57bbd931de7cda2d03ba561a58592bc6236R455, but yeah agree with the flow you illustrated 👍

scripts/utils/starknet.py
ClementWalter: in some cases, you may want to have in your deployment script a check regarding previously deployed contracts. However, when there are no previous deployments, this would raise because no file exist.

Eikix (approved):
lgtm

PR #664, opened 2023-07-28
Fix Caller Opcode

danilowhk:
Adding the `origin` arg to the `eth_call` and the `ExecutionContext`.
This solves the `from` (`msg.sender`) being 0 when in `@view` mode.

Time spent on this PR: 0.5

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

## What is the new behavior?

The `ExecutionContext` stores the ORIGIN of the calls, given as argument.


src/kakarot/precompiles/precompiles.cairo
```diff
@@ -76,6 +76,7 @@ namespace Precompiles {
             gas_price=0,
             starknet_contract_address=0,
             evm_contract_address=address,
+            origin=0,
```
Eikix, line 79: why is the origin 0 here?

Eikix (approved):
lgtm

src/kakarot/precompiles/precompiles.cairo
ClementWalter: no good reason but doesn't have any impact because this is a stopped context, so this value cannot be used. I've put the true origin instead though for clarity

Eikix (approved):

PR #669, opened 2023-08-02
feat: return eoa deployment fee to deployer

### LINKED ISSUE #668: feat: update EOA management, so that a deployer of an EOA account can recover the fees spent on it
# Feature Request

Update EOA deployment management to allow a deployer of an account to recover the fee spent for deploying this account.

We would need a storage variable for this, that can be updated periodically later on to keep track of latest account deployment fees. 

Then, inside `[deploy_externally_owned_account](https://github.com/kkrt-labs/kakarot/blob/03f3686780f6b0145b575de3168e7ebd1c18593f/src/kakarot/library.cairo#L282)` the fees can be paid back to the caller of the function.
```[tasklist]
### Tasks
- [x] update EOA management to so that deployer can recover fee
- [x] add tests for the same
```

bajpai244:
Improves EOA deployment management, with this any Starknet account should be able to deploy a Kakarot Account, and they should be able to redeem back the fees spent for this operation.

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Currently if someone deploys an account via [Kakarot.deploy_externally_owned_account](https://github.com/kkrt-labs/kakarot/blob/03f3686780f6b0145b575de3168e7ebd1c18593f/src/kakarot/library.cairo#L282), they would have to spend fees for that and they won't be able to redeem it.

Resolves #668

## What is the new behavior?

- Anyone can deploy a Kakarot Account for an eoa
- Address needs to have a certain amount of eth already at their address to get a deployment
- The deployer can recover the amount spent for this deployment


scripts/deploy_kakarot.py
```diff
@@ -2,7 +2,12 @@
 import logging
 from asyncio import run
 
-from scripts.constants import COMPILED_CONTRACTS, ETH_TOKEN_ADDRESS, EVM_ADDRESS
+from scripts.constants import (
+    COMPILED_CONTRACTS,
+    ETH_TOKEN_ADDRESS,
+    DEPLOY_FEE,
```
ClementWalter, line 8: run `make format` I guess that this doesn't pass

src/kakarot/library.cairo
```diff
@@ -192,6 +195,27 @@ namespace Kakarot {
         return (native_token_address_,);
     }
 
+     // @notice Set the deploy fee for deploying EOA on Kakarot.
```
ClementWalter, line 198: format

scripts/deploy_kakarot.py
bajpai244: done!

src/kakarot/library.cairo
```diff
@@ -285,8 +309,23 @@ namespace Kakarot {
...
+        let (caller_address) = get_caller_address();
         let (class_hash) = externally_owned_account_class_hash.read();
         let (starknet_contract_address) = Accounts.create(class_hash, evm_contract_address);
+
+        let (local native_token_address) = get_native_token();
+        let (local deploy_fee) = get_deploy_fee();
+
+        let amount = Helpers.to_uint256(deploy_fee);
+        let (success) = IERC20.transferFrom(
+            contract_address=native_token_address,
```
ClementWalter, line 323: format

src/kakarot/library.cairo
bajpai244: done!

src/kakarot/kakarot.cairo
```diff
@@ -71,6 +73,26 @@ func get_native_token{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_chec
     return Kakarot.get_native_token();
 }
 
+// @notice Set the deploy fee for deploying EOA on Kakarot.
```
ClementWalter, line 76: format

scripts/utils/kakarot.py
```diff
@@ -228,6 +228,9 @@ async def deploy_and_fund_evm_address(evm_address: str, amount: float):
     """
     Deploy an EOA linked to the given EVM address and fund it with amount ETH
     """
+
```
bajpai244, line 231: We are funding the account before, so that it has eth to pay back to the deployer.

tests/fixtures/ERC20.cairo
```diff
@@ -6,7 +6,7 @@
 from starkware.cairo.common.cairo_builtins import HashBuiltin
 from starkware.cairo.common.uint256 import Uint256
 
-from openzeppelin.token.erc20.library import ERC20
+from tests.fixtures.lib.erc20 import ERC20
```
ClementWalter, line 9: why this change?

src/kakarot/library.cairo
bajpai244: done [here](https://github.com/kkrt-labs/kakarot/pull/669/commits/e6c6e89d056885547c5df8e2f357d197bbfd944c).

tests/utils/accounts.py
```diff
@@ -0,0 +1,12 @@
+from starkware.starknet.testing.starknet import StarknetContract
+
+
+async def fund_evm_address(
```
ClementWalter, line 4: make it a fixture to have `kakarot` and `eth` already existing for the "end user" (the dev writing a test)

tests/fixtures/ERC20.cairo
bajpai244: the openzepplin contracts have an [assert](https://github.com/OpenZeppelin/cairo-contracts/blob/4dd04250c55ae8a5bbcb72663c989bb204e8d998/src/openzeppelin/token/erc20/library.cairo#L188), which reverts if the transfer is being done to the 0 address, in our test runner, what I can see is regardless of whatever I send the caller_address, inside the test it was 0, hence this was reverting, so I have changed the library to disable it.

tests/utils/constants.py
```diff
@@ -1,5 +1,7 @@
 CHAIN_ID = 1263227476  # KKRT (0x4b4b5254) in ASCII
 
+DEPLOY_FEE = 10000
```
bajpai244, line 3: I am not importing the constant from scripts here because if I do so it will ask for STARKNET_NETWORK environment variable.

src/kakarot/kakarot.cairo
bajpai244: done.

tests/integration/accounts/test_contract_account.py
```diff
@@ -91,3 +94,19 @@ async def test_should_give_infinite_allowance_to_kakarot(
...
+        async def test_should_take_deployment_fees(self, kakarot, eth):
+            evm_address = int(generate_random_evm_address(), 16)
+            computed_starknet_address = (
+                await kakarot.compute_starknet_address(evm_address).call()
+            ).result[0]
+
+            amount = 100000
+            await fund_evm_address(evm_address, kakarot, eth, amount)
+
+            await kakarot.deploy_externally_owned_account(evm_address).execute()
```
ClementWalter, line 107: could we use a non trivial `caller_address` so as to make sure that funds are not just sent to address 0

ClementWalter (changes requested):
plz review in github yourself first

otherwise, some python modif and should be good

tests/integration/accounts/test_contract_account.py
bajpai244: It makes no difference from what I could see, the caller address is 0 regardless of whatever caller address you choose.

ClementWalter: there is plenty of example, even in this file, where we use `.execute(caller_address=xxx)` that wouldn't pass otherwise so I think it's working. can you double check?

tests/fixtures/ERC20.cairo
ClementWalter: why would we mint to address `0`? I think that it's not necessary

tests/integration/accounts/test_contract_account.py
bajpai244: Yeah, I have tried it, the reason how I explored this issue was when I started seeing the error, cannot transfer to 0 address from the ERC20 library!

I tried changing the value multiple times, but it looked like the caller address from tx.info was always 0! But, I am fixing few tests, and lemme recheck again and get back!

tests/integration/solidity_contracts/PlainOpcodes/test_safe.py
```diff
@@ -1,38 +1,32 @@
 import pytest
+from tests.utils.constants import ACCOUNT_BALANCE
 
 
 @pytest.mark.asyncio
 @pytest.mark.Counter
```
ClementWalter, line 6: can you fix this copy/pasted mark also plz?

tests/integration/solidity_contracts/PlainOpcodes/test_safe.py
```diff
@@ -1,38 +1,32 @@
...
 class TestSafe:
     class TestReceive:
-        async def test_should_receive_eth(self, safe, owner, eth):
-            await eth.mint(owner.starknet_address, (int(1e9), 0)).execute(
-                caller_address=owner.starknet_address
-            )
-            await safe.deposit(value=int(1e9), caller_address=owner.starknet_address)
-            assert await safe.balance() == 1e9
+        async def test_should_receive_eth(self, safe, owner):
+            await safe.deposit(value= ACCOUNT_BALANCE, caller_address=owner.starknet_address)
```
ClementWalter, line 11: you definitely **need** to run `make format` (black here) on save

tests/integration/solidity_contracts/PlainOpcodes/test_safe.py
```diff
@@ -1,38 +1,39 @@
...
+
 
 @pytest.mark.asyncio
-@pytest.mark.Counter
+@pytest.mark.Safe
 @pytest.mark.usefixtures("starknet_snapshot")
 class TestSafe:
     class TestReceive:
-        async def test_should_receive_eth(self, safe, owner, eth):
-            await eth.mint(owner.starknet_address, (int(1e9), 0)).execute(
```
bajpai244, line 10: no need to mint here, as the account is already funded with (1e9) amount of tokens so that it could pay for its deployment fee.

Hence, the balance of token for `owner` is `1e9 - DEPLOY_FEE`,  which is represented by the constant `ACCOUNT_BALANCE` { this constant is later on used in other places }

tests/fixtures/ERC20.cairo
bajpai244: resolved, I think probably some human error on my end because of which I got confused.

tests/utils/accounts.py
bajpai244: done.

tests/integration/solidity_contracts/PlainOpcodes/test_safe.py
bajpai244: done.

tests/integration/solidity_contracts/PlainOpcodes/test_safe.py
bajpai244: gotcha!

Might be we can also add git hooks just like how we have at the RPC? Lemme know, I can create an issue for it.

tests/integration/accounts/test_contract_account.py
```diff
@@ -91,3 +93,25 @@ async def test_should_give_infinite_allowance_to_kakarot(
...
+            await kakarot.deploy_externally_owned_account(evm_address).execute(
+                caller_address=4
+            )
+
+            # asserting that the balance of the account is the amount minus the deployment fee
+            assert (
+                await eth.balanceOf(computed_starknet_address).call()
+            ).result.balance.low == ACCOUNT_BALANCE
+
+            # asserting that the deployer gets the Deployment Fee back { initial balance of the deployer is 0 }
```
ClementWalter, line 116: you should compare balance of deployer before and after and not assume that it was 0

ClementWalter (commented):
why did you have to put different caller_address everywhere?

bajpai244:
> why did you have to put different caller_address everywhere?

Initially it was to make sure that there is no other caller_address=4, because I make a check on balance of this address in a test after it does a deployment of an EOA.

So, I made caller_address=5, then I thought that for the other two caller_addresses as well I can make them 6 and 7 so that if someone writes a test in future for caller_address=5  just like I did for 4, they won't have to count the number of deployments 5 has made.

But, I think might be I have over optimized here, I also think I can make 5,6,7 and 2, so pushing with these changes.

tests/integration/accounts/test_contract_account.py
bajpai244: Done.

tests/integration/accounts/test_contract_account.py
```diff
@@ -91,3 +93,27 @@ async def test_should_give_infinite_allowance_to_kakarot(
...
+            self, kakarot, eth, fund_evm_address
+        ):
+            evm_address = int(generate_random_evm_address(), 16)
+            computed_starknet_address = (
+                await kakarot.compute_starknet_address(evm_address).call()
+            ).result[0]
+
+            await fund_evm_address(evm_address)
+
+            assert (await eth.balanceOf(4).call()).result.balance.low == 0
```
ClementWalter, line 107: please assign magic number to a variable before usage, and use the variable instead

tests/integration/accounts/test_contract_account.py
```diff
@@ -91,3 +93,27 @@ async def test_should_give_infinite_allowance_to_kakarot(
                 )
                 == "Uint256(low=340282366920938463463374607431768211455, high=340282366920938463463374607431768211455)"
             )
+
+        async def test_should_take_deployment_fees(
```
ClementWalter, line 97: forgot to say but this test is ill placed. This tests a kakarot entrypoint and should consequently be in the test_kakarot.py file, in a `TestDeployExternallyOwnedAccount` class

ClementWalter (commented):
I'd like us to clean these `caller_address`, plus move the test to the `test_kakarot.py` file, and eventually add some other tests for this `deploy_externally_owned_account` function

bajpai244:
> I'd like us to clean these `caller_address`, plus move the test to the `test_kakarot.py` file, and eventually add some other tests for this `deploy_externally_owned_account` function

- I have moved the test to `test_kakarot.py`
- I have create a fixture `deployer_address`, which is 2, and have used it in all places where I was using `caller_address=2`
- I am leaving the hardcoded value `caller_address=4` as it is, because it is only used in 1 test, and is supposed to be unique and not shared with other tests, hence not making a fixture for it now, but if need be to extend testing with this fixture we can make one in future for it.

tests/integration/accounts/test_contract_account.py
bajpai244: Can you explain the difference between these two tests:
- `def test_should_deploy_starknet_contract_at_corresponding_address`
- `def test_should_deploy_starknet_contract_with_correct_implementation`

tests/utils/helpers.py
```diff
@@ -145,11 +145,6 @@ def generate_random_private_key(seed=0):
     return keys.PrivateKey(int.to_bytes(random.getrandbits(256), 32, "big"))
 
 
-def generate_random_evm_address(seed=0):
```
bajpai244, line 148: It has now been made a fixutre, you can check #671 for more context.

tests/integration/accounts/test_contract_account.py
bajpai244: resolved.

tests/integration/accounts/test_contract_account.py
ClementWalter: I was thinking about the class hash, to make sure that it actually deploys an EOA. But let' not block for this

ClementWalter (approved):

PR #706, opened 2023-08-28
dev: use explicit name for custom network

### LINKED ISSUE #705: dev: rename custom RPC configuration NETWORK_NAME to `custom_rpc`
```[tasklist]
### Tasks
- [x] in https://github.com/kkrt-labs/kakarot/blob/c2cceb899f57f3e33988204559309749178ca7c5/scripts/constants.py#L87
- [x] replace the name variable with an explicit name such as `custom_rpc`
- [x] update accordingly with regards to https://github.com/kkrt-labs/kakarot/pull/686
```

Related code:
```python
else:
    NETWORK = {
        "name": "",
        "rpc_url": os.getenv("RPC_URL"),
        "explorer_url": "",
        "devnet": False,
        "check_interval": float(os.getenv("CHECK_INTERVAL", 6)),
        "max_wait": float(os.getenv("MAX_WAIT", 30)),
    }
```

ftupas:
Time spent on this PR: 0

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [X] Other (please describe): Fix code smell

Resolves #705 

## What is the new behavior?

- Use `custom_rpc` network name for custom RPC configuration i.e. when `STARKNET_NETWORK` is not set
- Fix code smell on the networks which deployer is being deployed on

### DIFF
--- a/scripts/constants.py
+++ b/scripts/constants.py
@@ -86,7 +86,7 @@ if os.getenv("STARKNET_NETWORK") is not None:
         )
 else:
     NETWORK = {
-        "name": "custom_rpc",
+        "name": os.getenv("RPC_NAME", "custom-rpc"),
         "rpc_url": os.getenv("RPC_URL"),
         "explorer_url": "",
         "devnet": False,
--- a/scripts/constants.py
+++ b/scripts/constants.py
@@ -86,7 +86,7 @@ if os.getenv("STARKNET_NETWORK") is not None:
         )
 else:
     NETWORK = {
-        "name": "",
+        "name": "custom_rpc",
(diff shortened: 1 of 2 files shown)


scripts/constants.py
```diff
@@ -86,7 +86,7 @@
         )
 else:
     NETWORK = {
-        "name": "",
+        "name": "custom_rpc",
```
ClementWalter, line 89: we have already `starknet-devnet` as a name so better not to mix cases. Can you make this `custum-rpc`? and why not adding an env var to possible set it, like `os.getenv("RPC_NAME", "custom-rpc")`?

ClementWalter (approved):

PR #708, opened 2023-08-29
initialize return_data to its length in zeroes when initializing call context

jobez:
Time spent on this PR:

## Pull request type

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes

Resolves #<Issue number>

## What is the new behavior?

`return_data` is prepped to be able to be finalized in the case where it is empty [here](https://gist.github.com/jobez/b21a84a35e7c97eb0de98424506cd4ef#file-resultant_traceback-txt-L132), where the finalized calling context attempts to write an empty return array to memory

-
-
-

### DIFF
--- a/src/kakarot/instructions/system_operations.cairo
+++ b/src/kakarot/instructions/system_operations.cairo
@@ -452,6 +452,7 @@ namespace CallHelper {
         // During teardown we update the memory using this offset
         let return_data: felt* = alloc();
         assert [return_data] = ret_offset;
+        Helpers.fill(ret_size, return_data+1, 0);
 
         // Load calldata from Memory
         let (calldata: felt*) = alloc();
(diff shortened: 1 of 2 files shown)


src/kakarot/instructions/system_operations.cairo
```diff
@@ -452,6 +452,7 @@ namespace CallHelper {
         // During teardown we update the memory using this offset
         let return_data: felt* = alloc();
         assert [return_data] = ret_offset;
+        Helpers.fill(ret_size, return_data+1, 0);
```
ClementWalter, line 455: you don't format automatically in your workflow

src/kakarot/instructions/system_operations.cairo
```diff
@@ -584,12 +585,16 @@ namespace CallHelper {
 
         let stack = Stack.push(ctx.stack, status);
         let ctx = ExecutionContext.update_stack(ctx, stack);
+        // in the case that the return data is empty, we use `slice_data` 
+        // to give us a zero padded array
+        let return_data_len = ctx.sub_context.return_data_len;
+        let return_data = Helpers.slice_data(return_data_len, ctx.sub_context.return_data, 0, return_data_len);
```
ClementWalter, line 591: how can this solve the case where `return_data_len` is not consistent with `return_data` anymore?

src/kakarot/instructions/system_operations.cairo
ClementWalter: since we append data to `return_data` in the `exec_return`, how do this work because the segment is already written?

Issue #474, opened 2023-02-02
feat: Handle non deployed contract in EXTCODE_SIZE/HASH/COPY


ClementWalter:
# Feature Request

**Describe the Feature Request**

These opcodes need to check wether there actually exists a contract deployed at the required address.
Before the Accounts refacto, this was done by using the Registry returning 0 for non deployed EVM address.

**Describe Preferred Solution**

There should be a way (a syscall, something) to check if there is a contract at a given felt

Eikix:
Can @karasakalmt work on this?

Eikix:
These opcodes give you an input: an EVM address := 160 bits

You will then get `starknet_address := Accounts.compute_starknet_address(evm_address)`

starknet_address <- felt 

and you need to know if at this specific starknet_address, there is a smart contract or not

karasakalmt:
I can be assigned

hurrikaanig:
#479 is referring to the same issue so I am closing it. I think there is no way to solve this issue for now, we need to wait for a forthcoming update

Eikix:
Thanks @hurrikaanig for your input, 
Quick update on this issue: it is still not possible to do

ClementWalter:
see https://github.com/kkrt-labs/kakarot/issues/719

Eikix:
> see #719

From this issue and PR, is this issue now:
- finished and should be closed
- re-enabled since before we couldn't check if wether there actually existed a contract deployed at an evm address

ClementWalter:
This has been fixed in https://github.com/kkrt-labs/kakarot/pull/720

PR #720, opened 2023-09-01
Add evm to starknet address mapping

### LINKED ISSUE #719: bug: cannot send ETH to a non deployed Account
# Bug Report

**Current behavior:**

When sending eth (meaning, kakaort native token), to a non-deployed account, Kakarot raises.

**Expected behavior:**

Possible to send eth to a non deployed account

**Steps to reproduce:**

Create a tx with `to` being set to a non deployed account, with `value > 0`.

**Related code:**

This bug has been introduced with the removal of this early return:

https://github.com/kkrt-labs/kakarot/commit/0c6e7a62a38e973ad6ceebfbf648841749ffe0ee#diff-93b1174fee1f2f813ece54d2954f5afe8a6813579979cb484752948514b51761L363-L368

This early return has been removed because it was a hack (with a todo) since it's a 100% valid to send an EVM tx (to 0 or to an address) with no payload.

However, removing this, when calling an address `to != 0`, enters the `call` branch, and assumes that the `to` is an existing contract (calling `IAccount` on it).

Since it's not possible to check on chain that a given account exists or not, the transaction fails trying to call `get_evm_address` onto a non existing account.

The different solutions envisioned to solve this issues were:

1. do nothing and wait for the `syscall` to be created
1. make the RPC check if the `to` exists and send this info to the EOA, which in turns would handle both cases accordingly
  - PROs:
    - doesn't change the current main Kakarot architecture
    - leverage AA
  - CONs:
    - only able to address a first level non existing `to`
    - split logic that would be in Kakarot when the syscall exists in between the RPC and the EOA
    - make the EOA diverge from standard Starknet Account
3. add the Account registry
  - PROs:
    - easy to implement
    - in place replacement of the missing syscall
    - leave all the logic in the Kakarot code base, and allow for bringing some of the RPC back into Kakarot
  - CONs:
    - we had it before and removed it, so we may have forgotten the reason why!
    - creates an ever growing mapping expected to store on the long run millions of addresses
      - though SW insures that there is no limitation in the size of a storage_var

After a ~1h long tech refinement with @greged93 and @bajpai244 we decided to go with the registry to move forward with the EF tests and unlock the RPC situation.

ClementWalter:
Time spent on this PR: 0.3

## Pull request type

- [x] Bugfix
- [x] Feature
- [ ] Build related changes

The Kakarot contract cannot know whether a starknet address is deployed or not.

Resolves #719 #474

## What is the new behavior?

A mapping evm_address -> starknet_address has been added:

- returns the starknet_address when it's deployed
- returns 0 otherwise

## Note

I didn't replace all the `compute_starknet_address`, and did actually the minimal changes,
since it is still possible in some case (e.g. transfer) to send ETH to a non deployed account.

However, now that we have introduced a mapping, we could move away from this deterministic mapping
to allow any starknet contract to claim ownership of an EVM address.

This is let to further dev, probably in cairo1.

### DIFF
--- a/scripts/utils/kakarot.py
+++ b/scripts/utils/kakarot.py
@@ -177,7 +177,7 @@ async def _contract_exists(address: int) -> bool:
 async def get_eoa(private_key=None, amount=0.1) -> Account:
     private_key = private_key or keys.PrivateKey(bytes.fromhex(EVM_PRIVATE_KEY[2:]))
 
-    starknet_address = await _get_starknet_address(
+    starknet_address = await _compute_starknet_address(
         private_key.public_key.to_checksum_address()
     )
     if not await _contract_exists(starknet_address):
@@ -230,7 +230,7 @@ async def eth_send_transaction(
     return await CLIENT.get_transaction_receipt(response.transaction_hash)
 
 
-async def _get_starknet_address(address: Union[str, int]):
+async def _compute_starknet_address(address: Union[str, int]):
     evm_address = int(address, 16) if isinstance(address, str) else address
     kakarot_contract = await _get_starknet_contract("kakarot")
     return (
@@ -257,7 +257,7 @@ async def deploy_and_fund_evm_address(evm_address: str, amount: float):
 
 
 async def fund_address(address: Union[str, int], amount: float):
-    starknet_address = await _get_starknet_address(address)
+    starknet_address = await _compute_starknet_address(address)
     logger.info(
         f"ℹ️  Funding EVM address {address} at Starknet address {hex(starknet_address)}"
     )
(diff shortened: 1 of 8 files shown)


bajpai244:
@ClementWalter what about the `deploy_externally_owned_account`? 

If someone is deploying EOAs through it, then they should be registered as well to the registry right?

bajpai244:
Another question is, do you think this will require some changes from the RPC, or will it work fine when we merge it?

ClementWalter:
> Another question is, do you think this will require some changes from the RPC, or will it work fine when we merge it?

I think that it will work directly, but that we could simplify the RPC by adding some more logic in kakarot (like the deployment of not deployed `origin` EOA in a tx)

greged93 (approved):
lgtm

PR #741, opened 2023-09-09
Add an example forge script for easy test to anvil

ClementWalter:
Time spent on this PR: 0.1

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

No way to easily run some forge script to deploy into a real EVM

## What is the new behavior?

Updated config to target anvil by default and an example script.

Note: once the end to end refacto is done, I think that we'll update the structure of the solidity project

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -43,3 +43,4 @@ docs/callgraphs/*
 .hypothesis/*
 node_modules
 tmp
+broadcast
--- a/foundry.toml
+++ b/foundry.toml
@@ -3,5 +3,8 @@ src = 'tests/integration/solidity_contracts'
 test = 'tests/integration/solidity_contracts'
 out = 'tests/integration/solidity_contracts/build'
 libs = ['lib']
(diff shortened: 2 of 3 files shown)


Eikix (commented):
How does one run this script?
Should we add comments / a make command?

ClementWalter:
> How does one run this script? Should we add comments / a make command?

tbh I'm not a huge fan of wrapping commands in makefile, it creates a later of complexity/abstraction and prevent people from learning how to use the few basic tool of the stack. It's like always rewriting a custom cli instead of using the standard one.

Eikix:
I think because in our case we have specific flags we need to pass everytime to make it work, it's worth it? Unless the plane command just works

ClementWalter:
> I think because in our case we have specific flags we need to pass everytime to make it work, it's worth it? Unless the plane command just works

used with anvil, the plain command `forge scripts path/to/script` works

PR #744, opened 2023-09-12
Feat/push-katana-dumps-on-push

### LINKED ISSUE #743: feat: dump katana on every PR merge
# Feature Request

Katana now seems to be providing the option to dump katana state, and then be able to load later on from it, this makes the option of deploying every PR merged to this repo to a katana instance, and then publishing the dump as an artifact to be consumed by other repos and CI jobs.

On every PR merge, run a github action, which does the following:
- deploys the current version to Katana and creates a dump
- publishes the dump, along with address of various deployments as artifacts.

Yes

bajpai244:
## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Resolves #743 

## What is the new behavior?

- We are uploading katana dump and deployments on Katana on every push to main.


ClementWalter:
can't we use the same job for the dump and the end-to-end ? with this PR, we'll `make deploy` 2 times

bajpai244:
> 

@ClementWalter so, I think the end-to-end tests are required to be run `per pull request` and the dumps are supposed to be done per `push`.

Both these actions are required to be run on different events, I don't thinking building twice will be a big issue as well, because Kakarot compilation is quite quick.

Lemme know what you think about it.

PR #748, opened 2023-09-18
Add trunk

ClementWalter:
Time spent on this PR: 0.4

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Miscellaneous linter configurations and tools.

## What is the new behavior?

Use [trunk](https://docs.trunk.io)


.github/workflows/ci.yml
```diff
@@ -40,22 +46,8 @@ jobs:
         run: make setup
 
       # Build artifacts
-      - name: Event name
```
Eikix, line 43: Why do we delete this job?

ClementWalter: I could put it in the `trunk` workflow actually, but I wanted to remove it from the build/tests CI. That being said, I'm not sure it's a good idea to have this since it doesn't enforce dev to have a proper local setup. With the trunk setup being one app/one command for all, I've guesses that it was better to enforce this than to patch at the CI level

Eikix: okay makes sense

Eikix (approved):
lgtm

we should add
```
concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: true
```
to everyone job in our repos?

PR #750, opened 2023-09-20
feat/add ef blockchain format test runner

jobez:
Time spent on this PR: 5 days

## Pull request type

Please check the type of change your PR introduces:

- [ ] Bugfix
- [X] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe):

## What is the current behavior?

Devs are required to do bespoke recreations for each failing case in the ef-test flow.

Resolves #738 

## What is the new behavior?

This PR enhances the debugging experience for ef-tests by offering multiple options for running test cases, each catering to different needs and scenarios.

EF tests are loaded as fixtures using the `pytest_generate_tests` hook. Devs can use the standard pytest keyword interface to select a subset of all GeneralStateTransition tests in Blockchain Test format for the Shanghai network.

Examples:

### Multiple suites
```
poetry run pytest ./tests/integration/ef_tests/ -k "mulmod or swap or calldatasize or push or random or calldatacopy" -vvv -rA
```

### Singular case
```
poetry run pytest ./tests/integration/ef_tests/ -k "sha3_d0g0v0_Shanghai" -vvv -rA 
```

PR #752, opened 2023-09-20
Feat/refactor return info

### LINKED ISSUE #718: bug: Finalization Issue with Mismatched return_data_length and Empty return_data
# Bug Report

**Kakarot version:**

**Current behavior:**

The sha ef tests presented a flow where a called context is finalized without ever executing the `exec_return` opcode

- in this case, the return_data of the execution context is empty, but the return_data_length is nonzero (in this tests case, it is 64
- when we finalize, we attempt to store the return data in the calling contexts memory
- we hit an error because the length of the return data does not match the empty return data array

**Expected behavior:**

- execution context finalization should be able to handle the case where an executing context has no return data, even when called with a caller with a retSize 
- we need to differentiate the caller's retSize and the called's return_data_length

**Steps to reproduce/related code:**

https://github.com/kkrt-labs/kakarot/blob/f00acce92296b4352169e58d33b87b558faa31f4/tests/integration/solidity_contracts/EFTests/test_sha3.py#L13-L41

**Other information:**

jobez:
I am working on top of #750 and will be simpler to review once that is reviewed and merged. Currently I am pushing here to to make sure changes are passing 

- all the python tests 
- all the e2e tests
- the particular failing ef-test case that inspired this refactor: sha3_d0g0v0_Shanghai

Time spent on this PR:

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Resolves #718 

## What is the new behavior?

-
-
-


src/kakarot/execution_context.cairo
```diff
@@ -258,13 +254,18 @@ namespace ExecutionContext {
     func revert(
         self: model.ExecutionContext*, revert_reason: felt*, size: felt
     ) -> model.ExecutionContext* {
-        memcpy(self.return_data, revert_reason, size);
+        alloc_locals;
+        let return_data: felt* = alloc();
+        memcpy(return_data, revert_reason, size);
+        tempvar return_info: model.ReturnInfo* = new model.ReturnInfo(
+            len=size, data=return_data, size=0, offset=0
```
Eikix, line 261: here we have many concepts: len, size, offset,
could we maybe find more explicit names for it?

we got:
len=size, size=0 my brain cannot comprehend

Eikix: here I don't understand why we set size and offset to 0

Eikix: maybe adding comments to walk the reader through the process can help

src/kakarot/execution_context.cairo
```diff
@@ -529,12 +528,15 @@ namespace ExecutionContext {
     }(
         self: model.ExecutionContext*, new_return_data_len: felt, new_return_data: felt*
     ) -> model.ExecutionContext* {
+        tempvar new_return_info: model.ReturnInfo* = new model.ReturnInfo(
+            len=new_return_data_len, data=new_return_data, size=0, offset=0
```
Eikix, line 532: same questions here,
can we add comments of what we're doing?

src/kakarot/instructions/environmental_information.cairo
```diff
@@ -549,7 +549,7 @@ namespace EnvironmentalInformation {
         bitwise_ptr: BitwiseBuiltin*,
     }(ctx: model.ExecutionContext*) -> model.ExecutionContext* {
         // Get return data size.
-        let return_data_size = Helpers.to_uint256(ctx.sub_context.return_data_len);
+        let return_data_size = Helpers.to_uint256(ctx.sub_context.return_info.len);
```
Eikix, line 552: here again, i face the confusion between the words size and len, and i don't really know what is what
though i'm playing a bit dumb, but we may be able to find ubiquitous language to guide us through this hardship

src/kakarot/instructions/system_operations.cairo
```diff
@@ -516,6 +517,12 @@ namespace CallHelper {
                 value=call_args.value,
             );
             let (starknet_contract_address) = Accounts.compute_starknet_address(call_args.address);
+            tempvar return_info: model.ReturnInfo* = new model.ReturnInfo(
```
Eikix, line 520: ok here i understand,
we init a returninfo struct with ->
len = 0, 
return_data = empty array
size = requested return size
offset = offset in memory to store the data

Eikix: can this declaration live before the if branch? to avoid duplication below?

src/kakarot/precompiles/precompiles.cairo
```diff
@@ -47,28 +47,31 @@ namespace Precompiles {
...
     ) -> model.ExecutionContext* {
         alloc_locals;
 
         // Execute the precompile at a given address
         let (output_len, output, gas_used) = _exec_precompile(address, calldata_len, calldata);
 
         // Copy results of precompile to return data
-        memcpy(return_data, output, output_len);
+
+        memcpy(return_info.data, output, output_len);
```
Eikix, line 59: here how do we know return_info.data is empty?
meaning that if `return_info.data` already has some values, this would fail?

src/kakarot/execution_context.cairo
jobez: its still a draft because i was focused on the structural changes and passing the test suite before making it reader friendly :)

src/kakarot/instructions/environmental_information.cairo
jobez: i started with just using the terms as they are used in the call opcodes and was planning on clarifying the names once everything compiled, passed the pytest stuff, and the inciting ef-test case that inspired this ask for a refactor

PR #755, opened 2023-09-25
feat: add ef-tests to CI

### LINKED ISSUE #730: feat: Integrate ef-test to the CI
# Feature Request

We are currently doing ef-tests in parallel, we some tests passing, and we as we finding bugs, we are updating the Kakarot source, we would not want any new changes being made to Kakarot to break any compatibility with any passing EF tests, hence we should integrate ef-tests in the CI.

The flow looks like the following:
- on every PR start a CI job
- download the [dump-katana](https://github.com/kkrt-labs/kakarot-rpc/blob/main/crates/core/src/test_utils/bin/dump-katana.rs) binary in the CI
- take the builds of the current Kakarot contracts
- make dump-katana create a dump from these builds
- download the binary of the ef-test runner
- execute the runner with this binary

```mermaid
flowchart TD
    A[on every PR] -->| start a CI job | B( CI JOB ) 
    B -->|1| C[ download katana-dump ]
    B -->|2| D[ build kakarot ]
    B -->|3| E[ build dumps ]
    B -->|4| F[ download ef-test runner ]
    B -->|5| G[ run tests with runner ]
```

Yes  🥕 
```[tasklist]
### Tasks
- [ ] https://github.com/kkrt-labs/ef-tests/issues/75
- [ ] https://github.com/kkrt-labs/kakarot-rpc/issues/530
```

bajpai244:
## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

The current kakarot changes that are pushed are not tested against the current set of passing ef-tests.

Resolves #730 

## What is the new behavior?

- All new Kakarot changes are tested against ef-tests

We have the following flow:
- build kakarot for the current proposed change
- create a katana dump for the change via RPC
- run ef-tests against this dump


ClementWalter (commented):
couldn't you 
- reuse the dump-katana workflow?
- reuse builds from other workflows?

bajpai244:
> couldn't you
> 
> * reuse the dump-katana workflow?
> * reuse builds from other workflows?

So, the goal of the feature is to:
- take a new change that someone is trying to push { hence we are building }
- make a dump out of this new change { we need a katana dump for this specific new change that someone is trying to push }

Lemme know if this makes sense, or you have some other comments.

.github/workflows/ef-tests.yml
```diff
@@ -0,0 +1,118 @@
...
+        with:
+          toolchain: stable
+      - name: Rust cache
+        uses: Swatinem/rust-cache@v2
+        with:
+          cache-on-failure: "true"
+      - name: make environment
+        run: ./scripts/make_with_env.sh
+        working-directory: kakarot-rpc
+      - name: pull kakarot submodule
```
ClementWalter, line 76: why do you pull the submodule if you build from kakarot directly? also, using this you'll have the latest kakarot@main, and not the one built just before

ClementWalter (commented):
I'm horrified by the complexity of this job, but if it passes, why not

.github/workflows/ef-tests.yml
bajpai244: The reason is that the dump-katana binary will look for two file-paths:
- kakarot-rpc/lib/kakarot
- kakarot-rpc/lib/kakarot/tests/integration/solidity_contracts

I am pulling the submodule just to get the directory structure, and then I am copying the builds and not building it { hence I am still using the latest builds! }

image

If you want, then I can create these directories manually as well!

bajpai244:
> I'm horrified by the complexity of this job, but if it passes, why not

Yeah, this should heavily reduced once we start running the ef-tests on our own seqeuncer!

ClementWalter (dismissed):
And let’s update this when the sequencer is reasy

.github/workflows/ef-tests.yml
```diff
@@ -0,0 +1,118 @@
...
+        run: |
+          git submodule update --init --remote ./lib/kakarot
+        working-directory: kakarot-rpc
+      - name: copy kakarot builds
+        run: |
+          cp -r ./kakarot/build ./kakarot-rpc/lib/kakarot
+          cp -r ./kakarot/tests/integration/solidity_contracts/build ./kakarot-rpc/lib/kakarot/tests/integration/solidity_contracts
```
Eikix, line 83:
```suggestion
          cp -r ./kakarot/solidity_contracts/build ./kakarot-rpc/lib/kakarot/solidity_contracts
```

Eikix: By the way should it be `cp -r ./kakarot/solidity_contracts/build ./kakarot-rpc/lib/kakarot/solidity_contracts` or 
`cp -r ./kakarot/solidity_contracts/build ./kakarot-rpc/lib/kakarot/solidity_contracts/build`

Eikix:
Is the custom sequencer already in production? if so, does this job use it?

bajpai244:
@Eikix the job has been updated work with the new changes of the ef-tests runner.

.github/workflows/ef-tests.yml
```diff
@@ -0,0 +1,78 @@
...
+on:
+  pull_request:
+    branches:
+      - main
+
+permissions: read-all
+
+jobs:
+  ef-tests:
+    runs-on: ubuntu-latest
```
Eikix, line 12: how long does this take to run with a little machine such as this?

.github/workflows/ef-tests.yml
```diff
@@ -0,0 +1,78 @@
...
+      - name: Load cached venv
+        id: cached-poetry-dependencies
+        uses: actions/cache@v3
+        with:
+          path: .venv
+          key: venv-${{ runner.os }}-${{ hashFiles('**/poetry.lock') }}
+      - name: Install dependencies
+        # TODO: investigate why cache is not working and remove this
+        # if: steps.cached-poetry-dependencies.outputs.cache-hit != 'true'
+        run: make setup
```
Eikix, line 52: why do we need to run make setup?

bajpai244: make setup does `poetry install`.

.github/workflows/ef-tests.yml
Eikix: yes for sure

.github/workflows/ef-tests.yml
```diff
@@ -0,0 +1,78 @@
...
+        uses: actions/checkout@v3
+        with:
+          repository: kkrt-labs/ef-tests
+          path: ef-test
+      - name: Rust cache
+        uses: Swatinem/rust-cache@v2
+        with:
+          cache-on-failure: "true"
+      - name: setup
+        run: make setup && make pull-kakarot
```
Eikix, line 70: I think it's best to do `setup-kakarot` and all the build phase above

.github/workflows/ef-tests.yml
bajpai244: bumped to 32 cores, which is what ef-tests is using for its CI.

.github/workflows/ef-tests.yml
bajpai244: I am not doing `setup-kakarot` on purpose, as it will build Kakarot as well, which we don't need as we already have the build with us on which we want to do this testing.

Eikix: I see, I just thought that since we do pull-kakarot, it'll pull the kakarot submodule from and use that specific submodule for the action

Eikix:
You'll need to add a trunk ignore as it doesn't recognize the 32 cores runner (weird)

.github/workflows/ef-tests.yml
```diff
@@ -0,0 +1,79 @@
...
+on:
+  pull_request:
+    branches:
+      - main
+
+permissions: read-all
+
+jobs:
+  ef-tests:
+    # trunk-ignore(actionlint/runner-label)
```
ClementWalter, line 12: you need to add a file `.github/actionlint.yaml` with

```
self-hosted-runner:
  labels:
    - ubuntu-latest-16-cores
```

instead

.github/workflows/ef-tests.yml
```diff
@@ -0,0 +1,79 @@
...
+    runs-on: ubuntu-latest-32-cores
+    steps:
+      # Install packages
+      - name: install packages
+        run: sudo apt-get install curl
+
+      - name: Install Foundry
```
ClementWalter, line 19: why do we need foundry?

.github/workflows/ef-tests.yml
ClementWalter, line 1: we should get the builds from an other workflow instead of building again

.github/workflows/ef-tests.yml
```diff
@@ -0,0 +1,79 @@
...
+
+on:
+  pull_request:
+    branches:
+      - main
+
+permissions: read-all
+
+jobs:
+  ef-tests:
```
ClementWalter, line 11: I think that you could just put this job in the ci.yml file, after `tests`, you make it `needs: tests` and reuse the build from the previous job

.github/workflows/ef-tests.yml
bajpai244: resolved.

.github/workflows/ef-tests.yml
bajpai244: Nice catch, we needed it with last approach, we no more need it.

bajpai244: resolved.

.github/workflows/ef-tests.yml
bajpai244: resolved, now we take them from the `build` job.

.github/workflows/ef-tests.yml
bajpai244: done.

.github/workflows/ci.yml
```diff
@@ -109,6 +116,35 @@ jobs:
...
+      - name: Rust cache
+        uses: Swatinem/rust-cache@v2
+        with:
+          cache-on-failure: "true"
+      - name: setup
+        run: make setup && make pull-kakarot
+      - uses: taiki-e/install-action@nextest
+
+      - name: make build directory
+        run: cd ./lib/kakarot && mkdir build
```
ClementWalter, line 136: do we really need this? maybe the actions/download-artifacts does it if the path doesn't exist

.github/workflows/ci.yml
```diff
@@ -109,6 +116,35 @@ jobs:
...
+      - name: Checkout ef-tests
+        uses: actions/checkout@v3
+        with:
+          repository: kkrt-labs/ef-tests
+      - name: Rust cache
+        uses: Swatinem/rust-cache@v2
+        with:
+          cache-on-failure: "true"
+      - name: setup
+        run: make setup && make pull-kakarot
```
ClementWalter, line 132: we don't need to pull kakarot anymore, do we?

.github/workflows/ci.yml
bajpai244: I can try! 

Lemme push somthing and see if it can work without it.

bajpai244: Seems to be working, removed and resolved.

.github/workflows/ci.yml
bajpai244: removed `make pull-kakarot`, resolved.

PR #765, opened 2023-10-09
Fix OutOfBounds for return_data

### LINKED ISSUE #718: bug: Finalization Issue with Mismatched return_data_length and Empty return_data

ClementWalter:
Time spent on this PR: 0.4

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Fails to finalize a calling context when `ret_size` is greater than `return_data_len`.

Resolves #718 

## What is the new behavior?

Use `Helpers.slice` to make sure that the stored `return_data` is big enough.

Note: we previously stored the `ret_offset` in the `sub_context.return_data[0]`. I have put both `ret_offset` and `ret_size` in the calling_context return data, as:

- this is more logical: they belong to the calling context and not the sub context
- it is safe because return_data is only written in RETURN, which stops the context, so no data could have been put before a call
- actually I've seen that geth and the execution spec reads RETURNDATASIZE and RETURNDATACOPY directly from the execution context and don't keep a ref to the previous sub_context. Might be an idea for a future refacto

### DIFF
--- a/scripts/compile_kakarot.py
+++ b/scripts/compile_kakarot.py
@@ -1,9 +1,12 @@
 # %% Imports
 import logging
+import multiprocessing as mp
 from datetime import datetime
 
 from scripts.constants import COMPILED_CONTRACTS, NETWORK
 from scripts.utils.starknet import compile_contract
+
+mp.set_start_method("fork")
 
 logging.basicConfig()
 logger = logging.getLogger(__name__)
@@ -15,8 +18,8 @@ def main():
     # %% Compile
     logger.info(f"ℹ️  Compiling contracts for network {NETWORK['name']}")
     initial_time = datetime.now()
-    for contract in COMPILED_CONTRACTS:
-        compile_contract(contract)
+    with mp.Pool() as pool:
+        pool.map(compile_contract, COMPILED_CONTRACTS)
 
     logger.info(
         f"✅ Compiled all in {(datetime.now() - initial_time).total_seconds():.2f}s"
--- a/src/kakarot/execution_context.cairo
+++ b/src/kakarot/execution_context.cairo
@@ -518,8 +518,8 @@ namespace ExecutionContext {
     // @notice Update the return data of the current execution context.
     // @dev The memory is updated with the given memory.
     // @param self The pointer to the execution context.
-    // @param new_return_data_len The length of the return data array.
-    // @param new_return_data The return data array.
+    // @param return_data_len The length of the return data array.
+    // @param return_data The return data array.
     // @return ExecutionContext The pointer to the updated execution context.
     func update_return_data{
         syscall_ptr: felt*,
@@ -527,14 +527,14 @@ namespace ExecutionContext {
         range_check_ptr,
         bitwise_ptr: BitwiseBuiltin*,
     }(
-        self: model.ExecutionContext*, new_return_data_len: felt, new_return_data: felt*
+        self: model.ExecutionContext*, return_data_len: felt, return_data: felt*
     ) -> model.ExecutionContext* {
         return new model.ExecutionContext(
             call_context=self.call_context,
             program_counter=self.program_counter,
             stopped=self.stopped,
-            return_data=new_return_data,
-            return_data_len=new_return_data_len,
+            return_data=return_data,
+            return_data_len=return_data_len,
             stack=self.stack,
             memory=self.memory,
             gas_used=self.gas_used,
--- a/src/kakarot/execution_context.cairo
+++ b/src/kakarot/execution_context.cairo
@@ -60,7 +60,6 @@ namespace ExecutionContext {
         dw 0;  // evm_contract_address
         dw 0;  // origin
         dw 0;  // calling_context
-        dw 0;  // sub_context
         dw 0;  // destroy_contracts_len
         dw 0;  // destroy_contracts
         dw 0;  // events_len
@@ -113,7 +112,6 @@ namespace ExecutionContext {
 
         let stack: model.Stack* = Stack.init();
         let memory: model.Memory* = Memory.init();
-        let sub_context = init_empty();
 
         return new model.ExecutionContext(
             call_context=call_context,
@@ -130,7 +128,6 @@ namespace ExecutionContext {
             evm_contract_address=evm_contract_address,
             origin=origin,
             calling_context=calling_context,
-            sub_context=sub_context,
             destroy_contracts_len=0,
             destroy_contracts=empty_destroy_contracts,
             events_len=0,
@@ -197,17 +194,6 @@ namespace ExecutionContext {
         return FALSE;
     }
 
-    // @notice Return whether the current execution context is a leaf.
-    // @dev A leaf context is a context without sub context.
-    // @param self The pointer to the execution context.
-    // @return is_leaf TRUE if the execution context is a leaf, FALSE otherwise.
-    func is_leaf(self: model.ExecutionContext*) -> felt {
-        if (cast(self.sub_context.call_context, felt) == 0) {
-            return TRUE;
-        }
-        return FALSE;
-    }
-
     // @notice Stop the current execution context.
     // @dev When the execution context is stopped, no more instructions can be executed.
     // @param self The pointer to the execution context.
@@ -228,7 +214,6 @@ namespace ExecutionContext {
             evm_contract_address=self.evm_contract_address,
             origin=self.origin,
             calling_context=self.calling_context,
-            sub_context=self.sub_context,
             destroy_contracts_len=self.destroy_contracts_len,
             destroy_contracts=self.destroy_contracts,
             events_len=self.events_len,
@@ -258,12 +243,11 @@ namespace ExecutionContext {
     func revert(
         self: model.ExecutionContext*, revert_reason: felt*, size: felt
     ) -> model.ExecutionContext* {
-        memcpy(self.return_data, revert_reason, size);
         return new model.ExecutionContext(
             call_context=self.call_context,
             program_counter=self.program_counter,
             stopped=TRUE,
-            return_data=self.return_data,
+            return_data=revert_reason,
             return_data_len=size,
             stack=self.stack,
             memory=self.memory,
@@ -274,7 +258,6 @@ namespace ExecutionContext {
             evm_contract_address=self.evm_contract_address,
             origin=self.origin,
             calling_context=self.calling_context,
-            sub_context=self.sub_context,
             destroy_contracts_len=self.destroy_contracts_len,
             destroy_contracts=self.destroy_contracts,
             events_len=self.events_len,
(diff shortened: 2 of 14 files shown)


Eikix:
> * actually I've seen that geth and the execution spec reads RETURNDATASIZE and RETURNDATACOPY directly from the execution context and don't keep a ref to the previous sub_context. Might be an idea for a future refacto

This is what we're doing in SSJ IIRC,
We're storing inside the execution context a child_return_data field, that's an Option<Span<u8>>
It's None until the child context RETURNS and then stores the child return data, 

That way, we can: access its size -> `Span::len`, and copy it into memory

scripts/compile_kakarot.py
```diff
@@ -1,10 +1,13 @@
 # %% Imports
 import logging
+import multiprocessing as mp
```
Eikix, line 3: what is this for?

src/kakarot/instructions/system_operations.cairo
```diff
@@ -581,15 +573,23 @@ namespace CallHelper {
...
         let ctx = ExecutionContext.update_stack(ctx, stack);
 
-        // ret_offset, see prepare_args
-        let memory = Memory.store_n(
-            ctx.memory,
-            ctx.sub_context.return_data_len,
-            ctx.sub_context.return_data,
-            [ctx.sub_context.return_data - 1],
+        // see prepare_args
+        let ret_offset = [ctx.return_data];
```
Eikix, line 577: I wonder why we store the `ret_offset` in the first element of the return data,
It is known at the time the CALL is made, meaning that it could be a call argument just like `return_data_len`

src/kakarot/instructions/system_operations.cairo
```diff
@@ -581,15 +573,23 @@ namespace CallHelper {
...
+        let ret_size = ctx.return_data_len;
+        let return_data = Helpers.slice_data(
```
Eikix, line 579: does this cover the case where ret_size > ctx.sub_context.return_data_len
-> meaning does it padd with zeroes the return_data of the subcontext?

src/kakarot/instructions/system_operations.cairo
```diff
@@ -581,15 +573,23 @@ namespace CallHelper {
...
+            data_len=ctx.sub_context.return_data_len,
+            data=ctx.sub_context.return_data,
+            data_offset=0,
+            slice_len=ret_size,
         );
-
+        let memory = Memory.store_n(ctx.memory, ret_size, return_data, ret_offset);
         let ctx = ExecutionContext.update_memory(ctx, memory);
+
+        // Clear sub_context's return_data args
```
Eikix, line 588: Why do we need to clear it?

Eikix:
> it is safe because return_data is only written in RETURN, which stops the context, so no data could have been put before a call

What happens if:
-> A Context CALLs into a subcontext, let's call them resp. ctx and sub_ctx
-> ctx.return_data is written to, same for ctx.return_data_len as you explained,
-> sub_ctx RETURNs, we use those values to properly handle the sub_ctx.return_data and sub_ctx.return_data_len
-> now, ctx RETURNs, but we'll try to overwrite return_data that's already written to! -> cairo error


UPDATE: as i'm writing this, I see https://github.com/kkrt-labs/kakarot/pull/765#discussion_r1351349801
And realize we're clearing it.
Ok! I think https://github.com/kkrt-labs/kakarot/pull/765/files/3fee6647c0636659307d460ac27e6134a2f79e8a#diff-c44c003822c0bac680d33a3719a1255689022e7dbdbe51c30093300e8067bd26R588 this line needs more commenting for context

scripts/compile_kakarot.py
ClementWalter: for compiling the contracts in parallel

src/kakarot/instructions/system_operations.cairo
ClementWalter: I don't understand this. When a CALL is made, we create a sub context, which is run in place of the calling context. When the sub context is stopped (possibly after hundreds of opcodes called), then we go back to the calling context. How could we pass this info to have it in the finalize?

src/kakarot/instructions/system_operations.cairo
ClementWalter: the doc says

```
        // slice's len = min(slice_len, data_len-offset, 0)
        // which corresponds to full, partial or empty overlap with data
        // The result is zero-padded in case of partial or empty overlap.
```

src/kakarot/instructions/system_operations.cairo
ClementWalter: I think that I get it: we could remove the link to the `child_context` in the `ExecutionContext` and stores in the ctx.return_data the return_data of the last call. Works for me, it will save steps

Eikix: > I think that I get it: we could remove the link to the `child_context` in the `ExecutionContext` and stores in the ctx.return_data the return_data of the last call. Works for me, it will save steps

afaik this is an optimisation that is not related to my point

i meant it as a question: would it be cleaner to have `ret_offset` be its own variable, rather than being collocated at the first index of return_data

ClementWalter: ok so I did the non requested optim, still don't understand what you mean by "be its own variable". There is no difference with `ret_offset` and `ret_size`, both could be put in the `return_data` segment. `ret_size` is **not** the `return_data_len`.

The pb here is that we need to access at the end of the sub context data retrieved in the beginning of the sub context, and with the current implementation there is no way to easily pass these args. So we need to store them somewhere. It could be in the stack, in the memory, or as in the current implementation in the `return_data` (somehow hacky because it's not actually the return data but the args concerning the handling of the `return_data`).

The stack may make more sense, but would be less efficient

ClementWalter: ok, actually the CALL args are such that `ret_offset` and `ret_size` are the last items popped, so I'll update to pop them in the `finalize_calling_context` instead of directly in the `prepared_args`

src/kakarot/execution_context.cairo
```diff
@@ -258,12 +243,11 @@ namespace ExecutionContext {
     func revert(
         self: model.ExecutionContext*, revert_reason: felt*, size: felt
     ) -> model.ExecutionContext* {
-        memcpy(self.return_data, revert_reason, size);
         return new model.ExecutionContext(
             call_context=self.call_context,
             program_counter=self.program_counter,
             stopped=TRUE,
-            return_data=self.return_data,
+            return_data=revert_reason,
```
Eikix, line 250: nice fix

src/kakarot/instructions/stop_and_arithmetic_operations.cairo
```diff
@@ -50,8 +51,16 @@ namespace StopAndArithmeticOperations {
...
         range_check_ptr,
         bitwise_ptr: BitwiseBuiltin*,
-    }(ctx_ptr: model.ExecutionContext*) -> model.ExecutionContext* {
-        let ctx = ExecutionContext.stop(ctx_ptr);
+    }(ctx: model.ExecutionContext*) -> model.ExecutionContext* {
+        // return_data stored the return_data for the last executed sub context
+        // see CALLs opcodes. When we run the STOP opcode, we stop the current
+        // execution context with *no* return data (unlike RETURN and REVERT).
+        // hence we just clear the return_data and stop.
+        let (return_data: felt*) = alloc();
```
Eikix, line 59: nice documentation

src/kakarot/instructions/system_operations.cairo
```diff
@@ -555,42 +535,57 @@ namespace CallHelper {
...
-            // reverted contexts put a zero on the calling context's stack,
-            // so the calling context can handle failing cases
-            assert status = Uint256(low=0, high=0);
-        } else {
-            // otherwise, we put 1
-            assert status = Uint256(low=1, high=0);
-        }
+        // Pop ret_offset and ret_size
+        let (stack, popped) = Stack.pop_n(self=ctx.calling_context.stack, n=2);
+        let ret_offset = 2 ** 128 * popped[0].high + popped[0].low;
```
Eikix, line 540: thought i know the offset in memory can't probably be higher than a u64::max (eni calculated that the resulting memory expansion would cost billions), should we add some checks for overflow?

otherwise it could panic and revert the cairo vm?

src/kakarot/instructions/environmental_information.cairo
```diff
@@ -549,7 +549,7 @@ namespace EnvironmentalInformation {
         bitwise_ptr: BitwiseBuiltin*,
     }(ctx: model.ExecutionContext*) -> model.ExecutionContext* {
         // Get return data size.
-        let return_data_size = Helpers.to_uint256(ctx.sub_context.return_data_len);
+        let return_data_size = Helpers.to_uint256(ctx.return_data_len);
```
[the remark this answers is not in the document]
ClementWalter: my updated understanding of the topic is that there is only one RETURN_DATA storage, and that the previous implementation was somehow a mistake. Since the current behavior is the same as in geth and in the execution specs, i think that it's fine like that

src/kakarot/instructions/system_operations.cairo
```diff
@@ -435,7 +427,9 @@ namespace CallHelper {
         ctx: model.ExecutionContext*, call_args: CallArgs
     ) {
         alloc_locals;
-        let (stack, popped) = Stack.pop_n(self=ctx.stack, n=6 + with_value);
+        // Note: We don't pop ret_offset and ret_size here but at the end of the sub context
+        // See finalize_calling_context
+        let (stack, popped) = Stack.pop_n(self=ctx.stack, n=4 + with_value);
```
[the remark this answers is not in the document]
ClementWalter: this is not the same stack

src/kakarot/instructions/system_operations.cairo
ClementWalter: this does seem to be in the spec. If this brings an OOG issue, we should raise this in the corresponding component (probably the `Memory.store_n` where we would compute in advance the gas cost associated)

src/kakarot/instructions/system_operations.cairo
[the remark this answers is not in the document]
Eikix: so calling_ctx encounters CALL at pc X

we pop 4 from the stack instead of 6, 

we start the child execution context (different stack, different code)

we finalize the child context, come back up to calling_context, 
and then that calling_context still has 2 items of the stack to pop

Eikix (approved):
lgtm!! lfg bugs are getting rekt

PR #780, opened 2023-10-31
Use generic function for arithmetic ops

ClementWalter:
Time spent on this PR: 0.3

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

All the opcodes share the same pre and post processing (assert stack size, pop, push, increment gas) but the code doing it
is duplicated in each opcode.

## What is the new behavior?

Created a `exec_arithmetic_operation` that does the main logic and is used in each opcode.
Note that this is the first step of a refacto of the main jump table (evm.cairo) to jump
opcode families (ArithmeticOperations in this context) and not opcode directly.

This will allow opcodes of the same group (that often share the same logic) to reuse more code
(and hopefully save on steps).

### DIFF
--- a/src/kakarot/instructions/stop_and_arithmetic_operations.cairo
+++ b/src/kakarot/instructions/stop_and_arithmetic_operations.cairo
@@ -259,7 +259,6 @@ namespace Internals {
         bitwise_ptr: BitwiseBuiltin*,
     }(ctx: model.ExecutionContext*, opcode: model.Opcode*) -> model.ExecutionContext* {
         alloc_locals;
-        local stack: model.Stack*;
 
         let stack_underflow = is_le(ctx.stack.size, opcode.stack_input - 1);
         if (stack_underflow != 0) {
@@ -268,9 +267,10 @@ namespace Internals {
             return ctx;
         }
 
-        let (_stack, popped) = Stack.pop_n(ctx.stack, opcode.stack_input);
+        let (local stack, popped) = Stack.pop_n(ctx.stack, opcode.stack_input);
 
-        assert stack = _stack;
+        // offset is 1 (new line) + 4 (call + op + jmp + end) per opcode
+        // opcode is offset from by 0x1 (index of the first opcode)
         tempvar offset = 1 + 4 * (opcode.number - 0x1);
 
         // Prepare arguments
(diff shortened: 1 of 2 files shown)


src/kakarot/instructions/stop_and_arithmetic_operations.cairo
```diff
@@ -63,468 +44,374 @@ namespace StopAndArithmeticOperations {
...
         }
 
-        // Stack input:
-        // 0 - a: number.
-        // 1 - b: exponent.
-        let stack = ctx.stack;
-        let (stack, popped) = Stack.pop_n(self=stack, n=2);
+        let (_stack, popped) = Stack.pop_n(ctx.stack, opcode.stack_input);
+
+        assert stack = _stack;
```
Eikix, line 273: can we add comments for all the dark cairo magic we do for readability and maintanability

src/kakarot/instructions/stop_and_arithmetic_operations.cairo
```diff
@@ -63,468 +44,374 @@ namespace StopAndArithmeticOperations {
...
+        let opcode = cast([fp - 3], model.Opcode*);
+
+        // Finalize opcode
+        let stack = Stack.push(stack, result);
+        let ctx = ExecutionContext.update_stack(ctx, stack);
+        let ctx = ExecutionContext.increment_gas_used(self=ctx, inc_value=opcode.gas);
+        return ctx;
+    }
+
+    func add{range_check_ptr}(popped: Uint256*) -> Uint256 {
```
Eikix, line 327: is there an argument for using Uint256 or Uint256* in the return signature?

Eikix (commented):
I like where this is going^^

src/kakarot/instructions/stop_and_arithmetic_operations.cairo
ClementWalter: This is the starkware’s default, I’m just keeping the same. Furthermore a pointer cost one more (the size of the pointer)

ClementWalter: eventually, in the main body I had to create a pointer with `new` otherwise the VM was raising "unsafe pointer used" when doing 

```
let result = cast([ap - 2], Uint256*);
```

instead of

```
tempvar result = new Uint256([ap - 2], [ap - 1]);
```

Eikix: ok!

Eikix (approved):
lgtm

PR #802, opened 2023-11-07
Commit account iif it has code or nonce

### LINKED ISSUE #793: bug: commit account should not always deploy a new account
### Current behavior

When committing the state, when committing accounts, we should not deploy accounts that don't need to be deployed
Example:
- A tx towards an undeployed account, that justs transfers money.

Cases where things already work:
- create opcodes
- deploy tx
- call tx

This edge case where:
- we touch an undeployed account during a tx,
- but dont want to deploy it at the end, is not covered

Solution:
- add a new check on top of is_registered, check if nonce!=0 or has code

ClementWalter:
Time spent on this PR: 0.1

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

When an account is not registered, it's deployed as a CA in `commit` in any cases.

Resolves #793

## What is the new behavior?

A contract is deployed only if it has code or nonce.

### DIFF
--- a/src/kakarot/account.cairo
+++ b/src/kakarot/account.cairo
@@ -125,16 +125,27 @@ namespace Account {
                 return ();
             }
 
-            // Deploy accounts
-            let (class_hash) = contract_account_class_hash.read();
-            deploy(class_hash, self.address);
-            // Write bytecode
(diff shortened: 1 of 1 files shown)


Eikix (approved):
lgtm

PR #805, opened 2023-11-09
Fix ERC20 error due to newly deployed SELFDESTRUCT contracts

ClementWalter:
Time spent on this PR: 0.2

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

When a newly deployed contract calls SELFDESTRUCT, it's never deployed on Starknet and consequently
Kakarot is not allowed to `transferFrom` its contract.

The Cairo VM hence raises in `state.commit`.

## What is the new behavior?

In `State.commit`, when executing transfer to a newly deployed selfdestructed contract, uses
the Kakarot contract instead as temporary ETH holders.

Note: I have eventually realized that separating the balance from the rest of the Account "to save on loading bytecode just to send ETH" is indeed a nonsense because there is only two ways to send ETH without requiring the bytecode:
- the initial `value` of a tx, hence the sender is an EOA and loading its bytecode is just loading empty code so cheap
- the recipient of SELFDESTRUCT

Consequently, I think that the balance would be better re-integrated into the Account, but I don't know if the refacto is worth it per se. Maybe if we remove completely the ERC20, then we can do it in the meantime.

### DIFF
--- a/src/kakarot/state.cairo
+++ b/src/kakarot/state.cairo
@@ -5,7 +5,7 @@
 from starkware.cairo.common.alloc import alloc
 from starkware.cairo.common.cairo_builtins import HashBuiltin, BitwiseBuiltin
 from starkware.cairo.common.default_dict import default_dict_new, default_dict_finalize
-from starkware.cairo.common.dict import dict_read, dict_write, dict_squash
+from starkware.cairo.common.dict import dict_read, dict_write
 from starkware.cairo.common.dict_access import DictAccess
 from starkware.cairo.common.hash_state import hash_finalize, hash_init, hash_update, hash_felts
 from starkware.cairo.common.math import assert_not_zero
@@ -13,7 +13,8 @@ from starkware.cairo.common.memcpy import memcpy
 from starkware.cairo.common.registers import get_fp_and_pc
 from starkware.cairo.common.uint256 import Uint256, uint256_add, uint256_sub, uint256_le
 from starkware.starknet.common.storage import normalize_address
-from starkware.starknet.common.syscalls import call_contract, emit_event, get_contract_address
+from starkware.starknet.common.syscalls import call_contract
+from starkware.starknet.common.syscalls import emit_event
 
 from kakarot.account import Account
 from kakarot.storages import native_token_address, contract_account_class_hash
@@ -121,7 +122,6 @@ namespace State {
         range_check_ptr,
         bitwise_ptr: BitwiseBuiltin*,
     }(self: Summary*) {
-        alloc_locals;
         // Accounts
         Internals._commit_accounts(self.accounts_start, self.accounts);
 
@@ -130,11 +130,7 @@ namespace State {
 
         // Transfers
         let (native_token_address_) = native_token_address.read();
-        let accounts = self.accounts;
-        Internals._transfer_eth{accounts=accounts}(
-            native_token_address_, self.transfers_len, self.transfers
-        );
-        dict_squash(self.accounts_start, accounts);
+        Internals._transfer_eth(native_token_address_, self.transfers_len, self.transfers);
 
         return ();
     }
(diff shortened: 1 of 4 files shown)


src/kakarot/state.cairo
```diff
@@ -130,7 +130,11 @@ namespace State {
 
         // Transfers
         let (native_token_address_) = native_token_address.read();
-        Internals._transfer_eth(native_token_address_, self.transfers_len, self.transfers);
+        let accounts = self.accounts;
+        Internals._transfer_eth{accounts=accounts}(
+            native_token_address_, self.transfers_len, self.transfers
+        );
+        dict_squash(self.accounts_start, accounts);
```
Eikix, line 137: we weren't squashing the account dict before?

src/kakarot/state.cairo
```diff
@@ -435,17 +439,41 @@ namespace Internals {
...
 
         let transfer = [transfers];
-        IERC20.transferFrom(
-            token_address, transfer.sender.starknet, transfer.recipient.starknet, transfer.amount
-        );
+
+        let (kakarot_address) = get_contract_address();
+
+        let (pointer) = dict_read{dict_ptr=accounts}(key=transfer.recipient.starknet);
+        let recipient = cast(pointer, model.Account*);
```
Eikix, line 454: we should probably add some extensive comments to explain what is being done here,
just like the PR description

src/kakarot/state.cairo
```diff
@@ -435,17 +439,41 @@ namespace Internals {
...
+        let is_sender_registered = Account.is_registered(sender.address);
+        let use_kakarot_for_sender = sender.selfdestruct * (1 - is_sender_registered);
+        let sender_address = use_kakarot_for_sender * kakarot_address + (
+            1 - use_kakarot_for_sender
+        ) * transfer.sender.starknet;
+
+        // The default ERC20.transferFrom implementation raises even if sender = caller
+        // when there is no prior approval
+        if (sender_address == kakarot_address) {
+            IERC20.transfer(token_address, recipient_address, transfer.amount);
```
Eikix, line 472: should we add an invariant check here? something like:
transfer.amount < sender.balance (though this doesn't exist in the current model)

since this performs a direct transfer from Kakarot to some address, this is one of the attack vectors of our architecture:
deploy a contract with tx.value > 0 -> schedule for self destruct in same tx -> contract performs a lot of transfers -> in commit, kakarot will perform the transfers in its place

in the case of some weird wrap around, or failed check when creating a transfer, we might have a delicate situation

that being said, i know that we check sufficient balance before adding to the transfer dict

Eikix (commented):
smart workaround for this trap!
there is no better way to do this afaik,

unless we have a tricount hint^^ that is able to "flatten & clear" all transfers and minimize exchange, like a clearing house in tradfi

src/kakarot/state.cairo
ClementWalter: yes, but we read again, so we need to squash again

src/kakarot/state.cairo
ClementWalter: I agree it's not the best. Different solutions I see now are:
- deploying another "escrow" contract instead just for this purpose
- deploying the SELFDESTRUCTed account in any case, but without writing bytecode nor nonce

AFAIU SELFDESTRUCT, it seems like deploying the account and leaving it empty (no code nor nonce) will work for later user

src/kakarot/state.cairo
Eikix: is it "expensive" to squash twice?

src/kakarot/state.cairo
Eikix: > I agree it's not the best. Different solutions I see now are:
> 
> * deploying another "escrow" contract instead just for this purpose
> * deploying the SELFDESTRUCTed account in any case, but without writing bytecode nor nonce
> 
> AFAIU SELFDESTRUCT, it seems like deploying the account and leaving it empty (no code nor nonce) will work for later user
> 
> The benefit or deploying the selfdestructed Starknet account is that it's a one line change and everything else can be kept the same

But it introduces a weird state of 'ghost accounts' where the account is deployed
-> we need to be careful not to try and re-trigger a deploy syscall on the same address in the case of a later deployment, 
-> the account will be registered and "void", so we need to make sure we never use is_registered as a condition for "existing" but rather, hasCodeOrNonce, as in Kakarot SSJ, we extensively (maybe wrongly?) use the is_registered as a condition to be void.

ClementWalter: `has code or nonce` is the only valid check, and the EF tests eventually the only important things.

as a matter of fact, in this codebase there was only one place where I used `is_registered`, in the beginning of a tx

```
        // Check collision
        let is_registered = Account.is_registered(address.evm);
        let is_collision = is_registered * is_deploy_tx;
```

but I've just replaced with

```
        // Check collision
        let account = Account.fetch_or_create(address);
        let code_or_nonce = Account.has_code_or_nonce(account);
        let is_collision = code_or_nonce * is_deploy_tx;
```

src/kakarot/state.cairo
ClementWalter: I don't know the cost of squashing, but it's required in any case, and not optimizable so much other than optimizing the number of read. In any case, for this PR I've removed all this

ClementWalter:
ok let's see if the ef tests pass all, then I think that this final solution is much better

greged93 (approved):
just needs a rebase

PR #807, opened 2023-11-11
Put balance in Account instead of State

ClementWalter:
Time spent on this PR: 0.2

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

The `model.State` maintains a dict for accounts balances.
This is a mis conception due the the false belief that "we don't want
to fetch an account just to send ETH". But there is no way to send ETH
without making a CALL (apart in SELFDESTRUCT).

## What is the new behavior?

The balance belongs to the Account instead (as in the ef tests json description for example).

Note: this is a preliminary work to run Kakarot in standalone (no syscall, ie. no Starknet) mode.

### DIFF
--- a/src/kakarot/account.cairo
+++ b/src/kakarot/account.cairo
@@ -12,8 +12,6 @@ from starkware.cairo.common.hash import hash2
 from starkware.cairo.common.math_cmp import is_not_zero
 from starkware.cairo.common.memcpy import memcpy
 from starkware.cairo.common.uint256 import Uint256
-from starkware.starknet.common.storage import normalize_address
-from starkware.starknet.common.syscalls import deploy as deploy_syscall, get_contract_address
 from starkware.cairo.common.hash_state import (
     hash_finalize,
     hash_init,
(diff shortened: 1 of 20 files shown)


src/kakarot/account.cairo
```diff
@@ -177,10 +114,15 @@ namespace Account {
         alloc_locals;
         let starknet_account_exists = is_registered(address.evm);
 
+        let balance = read_balance(address);
+        tempvar balance_ptr = new Uint256(balance.low, balance.high);
```
Eikix, line 118: why do we need a `new` keyword here?

ClementWalter: this is the easy way to make it a pointer. Otherwise I need to use the `__fp__` variable to do `&balance`, which I can do if you prefer

Eikix: no preference! was just curious

Eikix (approved):
looks good to me

PR #810, opened 2023-11-14
fix: fix early return for create

### LINKED ISSUE #795: bug: collision in CREATE and CREATE2 should not revert
### Current behavior

In create and create2, https://github.com/kkrt-labs/kakarot/blob/6265d2b1b6fdac032f5cfeed34dbcc36d55aaf44/src/kakarot/instructions/system_operations.cairo#L836

We revert if there is a collision. We should increment the nonce of the caller (reverted create) and push 0 to the stack

https://github.com/ethereum/execution-specs/blob/3fe6514f2d9d234e760d11af883a47c1263eff51/src/ethereum/berlin/vm/instructions/system.py#L59

Eikix:
Time spent on this PR: 0.2 day

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Resolves #795 

## What is the new behavior?

- add early return for collision in create
- add early return for unsufficient balance in create (value > 0)

### DIFF
--- a/src/kakarot/instructions/system_operations.cairo
+++ b/src/kakarot/instructions/system_operations.cairo
@@ -705,7 +705,6 @@ namespace CreateHelper {
     }
 
     // @notice Pre-compute the evm address of a contract account before deploying it.
-    // @dev Side-effect: increment the nonce of the deployer account.
     func get_evm_address{
         syscall_ptr: felt*,
         pedersen_ptr: HashBuiltin*,
@@ -722,8 +721,6 @@ namespace CreateHelper {
         alloc_locals;
         let (state, account) = State.get_account(state, address);
         let nonce = account.nonce;
-        let account = Account.set_nonce(account, nonce + 1);
-        let state = State.set_account(state, address, account);
 
         // create2 context pops 4 off the stack, create pops 3
         // so we use popped_len to derive the way we should handle
@@ -754,15 +751,16 @@ namespace CreateHelper {
     }(ctx: model.ExecutionContext*, popped_len: felt, popped: Uint256*) -> model.ExecutionContext* {
         alloc_locals;
 
-        // Load CallContext bytecode code from memory
         let value = popped[0];
         let offset = popped[1];
         let size = popped[2];
 
+        // Load CallContext bytecode code from memory
         let (bytecode: felt*) = alloc();
         let (memory, gas_cost) = Memory.load_n(
             self=ctx.memory, element_len=size.low, element=bytecode, offset=offset.low
         );
+        let ctx = ExecutionContext.update_memory(ctx, memory);
 
         // Get target account
         let (state, evm_contract_address) = get_evm_address(
@@ -770,20 +768,6 @@ namespace CreateHelper {
         );
         let (starknet_contract_address) = Account.compute_starknet_address(evm_contract_address);
         tempvar address = new model.Address(starknet_contract_address, evm_contract_address);
-
-        // Create Account with empty bytecode
-        let (state, account) = State.get_account(state, address);
-        let is_collision = Account.has_code_or_nonce(account);
-
-        // Update calling context before creating sub context
-        let ctx = ExecutionContext.update_memory(ctx, memory);
-        let ctx = ExecutionContext.update_state(ctx, state);
-
-        if (is_collision != 0) {
-            let stack = Stack.push_uint128(ctx.stack, 0);
-            let ctx = ExecutionContext.update_stack(ctx, stack);
-            return ctx;
-        }
 
         let transfer = model.Transfer(
             sender=ctx.call_context.address, recipient=address, amount=value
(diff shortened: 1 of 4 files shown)


src/kakarot/instructions/system_operations.cairo
```diff
@@ -729,6 +731,7 @@ namespace CreateHelper {
         alloc_locals;
         let (state, account) = State.get_account(state, address);
         let nonce = account.nonce;
+        // Increment nonce of deployer account
```
ClementWalter, line 734: I don't really like this kind of doc really saying what can be read from the code

src/kakarot/instructions/system_operations.cairo
```diff
@@ -771,31 +774,46 @@ namespace CreateHelper {
...
         );
 
-        // Get new account address
+        // Get new account address and increment nonce of deployer account
         let (state, evm_contract_address) = get_evm_address(
             ctx.state, ctx.call_context.address, popped_len, popped, size.low, bytecode
         );
         let (starknet_contract_address) = Account.compute_starknet_address(evm_contract_address);
-        tempvar address = new model.Address(starknet_contract_address, evm_contract_address);
+        tempvar target_address = new model.Address(starknet_contract_address, evm_contract_address);
```
ClementWalter, line 782: not sure about the renaming from `address` to `target_address`

src/kakarot/state.cairo
```diff
@@ -272,9 +272,19 @@ namespace State {
         let fp_and_pc = get_fp_and_pc();
         local __fp__: felt* = fp_and_pc.fp_val;
 
+        // Early return if the sender is the recipient
```
ClementWalter, line 275: how is this comment informative?

src/kakarot/state.cairo
```diff
@@ -272,9 +272,19 @@ namespace State {
...
+        if (transfer.sender == transfer.recipient) {
+            return (self, 1);
+        }
+        // Early return if the amount is 0
```
ClementWalter, line 279: how is this comment informative?

src/kakarot/state.cairo
```diff
@@ -272,9 +272,19 @@ namespace State {
...
+        let (null_transfer) = uint256_eq(transfer.amount, Uint256(0, 0));
+        if (null_transfer != 0) {
+            return (self, 1);
+        }
+
         let (self, sender_balance_prev) = read_balance(self, transfer.sender);
         let (success) = uint256_le(transfer.amount, sender_balance_prev);
-
+        // Early return if the sender doesn't have enough balance with success = 0
```
ClementWalter, line 287: how is this comment informative?

ClementWalter (commented):
how many ef tests does this solve?
Also, I think that there is a place where the value is used to do early return and not `add_transfer`, that should be fixed

you add a lot of comments I don't fell bring any value to the codebase.

src/kakarot/instructions/system_operations.cairo
Eikix: I added this comment because it is surprising (but "efficient") that the `get_evm_address` method does mutate account of caller, 
I wanted to bring "visibility" to this surprising behaviour, though I will remove the comment.

src/kakarot/instructions/system_operations.cairo
Eikix: alright

Eikix:
> Also, I think that there is a place where the value is used to do early return and not `add_transfer`, that should be fixed

In system operations, line 800, we handle transfer in CREATE, but adding an "if amount == 0" would probably disturb most of my references. Not sure how to handle this without an early return inside the add_transfer

ClementWalter:
> > Also, I think that there is a place where the value is used to do early return and not `add_transfer`, that should be fixed
> 
> In system operations, line 800, we handle transfer in CREATE, but adding an "if amount == 0" would probably disturb the references and result in revoked reference errors. Not sure how to handle this without an early return inside the add_transfer. But I agree this early return should probably not be used

early return in `add_transfer` is fine for me. I was referring to this

https://github.com/ClementWalter/kakarot/blob/bff112a701d00e9e6991b9780ba66a2fad389f54/src/kakarot/instructions/system_operations.cairo#L238

that should then probably be removed since the `add_transfer` integrates it.

ClementWalter (commented):
I'm ok with the proposed changes in terms of code (though need to fix 2 tests). However I'm wondering if the proposed change implements the right behavior as it makes 30 ef-tests failing

Eikix:
Blocked by an update on the ef-test runner

tests/src/kakarot/instructions/test_system_operations.cairo
```diff
@@ -709,7 +709,7 @@ func test__exec_create2__should_return_a_new_context_with_bytecode_from_memory_a
 
     // Fill the stack with exec_create2 args
     let stack: model.Stack* = Stack.init();
-    tempvar value = new Uint256(1, 0);
+    tempvar value = new Uint256(0, 0);
```
ClementWalter, line 712: mint and leave value as 1

tests/src/kakarot/instructions/test_system_operations.py
```diff
@@ -105,14 +105,24 @@ async def test_should_return_a_new_context_based_on_calling_ctx_stack(
             await mint(ZERO_ACCOUNT, 2)
             await system_operations.test__exec_delegatecall__should_return_a_new_context_based_on_calling_ctx_stack().call()
 
-    async def test_create(self, system_operations):
+    async def test_create(self, system_operations, eth):
         salt = 0
         # given we start with the first anvil test account
+        evm_caller_address_int = int(0xF39FD6E51AAD88F6F4CE6AB8827279CFFFB92266)
```
ClementWalter, line 111: don't need the `int`

tests/src/kakarot/instructions/test_system_operations.py
```diff
@@ -105,14 +105,24 @@ async def test_should_return_a_new_context_based_on_calling_ctx_stack(
...
-            0xF39FD6E51AAD88F6F4CE6AB8827279CFFFB92266
+            evm_caller_address_int.to_bytes(20, byteorder="big")
         )
         expected_create_addr = get_create_address(evm_caller_address, salt)
 
+        # Mint to evm_caller_address
+        # Test runner starknet address is by default 0 so we fund address 0
+        starknet_contract_address = (
+            await system_operations.compute_starknet_address(
+                evm_caller_address_int
```
ClementWalter, line 121: `int(evm_caller_addres, 16)`

ClementWalter (approved):

PR #822, opened 2023-11-23
Fix bytecode is not written when starknet contract already exists

### LINKED ISSUE #821: bug: accounts created at previously SELFDESTRUCT don't write bytecode
### Current behavior

In `Account.commit`, one uses the `is_register` to check whether one needs to deploy and write bytecode.
However, a previously SELFDESTRUCTed contract is registered but still need to commit its bytecode.

### Expected behavior

Write bytecode

### Related code

Add here a condition:

```
let (registered_len) = IAccount.bytecode_len();
if (registered_len == 0) {
    IContractAccount.write_bytecode(starknet_address, self.code_len, self.code);
}
```

See
https://github.com/ClementWalter/kakarot/blob/4c108731aa7adae1c73853a2ba0e612b9dcf6a51/src/data_availability/starknet.cairo#L170

ClementWalter:
Time spent on this PR: 0.3

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Resolves #821

## What is the new behavior?

Overwrite bytecode if stored bytecode len doesn't match account.code_len

### DIFF
--- a/solidity_contracts/tests/PlainOpcodes.t.sol
+++ b/solidity_contracts/tests/PlainOpcodes.t.sol
@@ -79,7 +79,7 @@ contract PlainOpcodesTest is Test {
         assertEq(_address_1, address(0));
     }
 
-    function testSelfDestructAndCreateAgainCollision() public {
+    function testCreate2CollisionReturnsZeroAddress() public {
         bytes memory bytecode = type(ContractWithSelfdestructMethod).creationCode;
         uint256 salt = 1234;
         address _address_0;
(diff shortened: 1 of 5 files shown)


solidity_contracts/tests/PlainOpcodes.t.sol
```diff
@@ -63,17 +63,35 @@ contract PlainOpcodesTest is Test {
...
-        uint256 count = contract_.count();
-        assertEq(count, 2);
+        address _address_1;
+        assembly {
+            _address_1 := create2(0, add(bytecode, 32), mload(bytecode), salt)
+        }
+        assertEq(_address_1, address(0));
+    }
+
+    function testSelfDestructAndCreateAgainCollision() public {
```
enitrat, line 82: wrong title

enitrat (approved):

PR #824, opened 2023-11-23
Fix safe uint256 sub

### LINKED ISSUE #683: bug: decrementing 0 should fail, but it's getting successful transaction
# Bug Report

**Kakarot version:**

when decrementing 0 it normally should get a failed transaction, but actually it is getting a successful transaction.

**Current behavior:**

while testing that decrement function with a `count` state `0`:

```
    // Function to decrement count by 1
    function dec() public {
        // This function will fail if count = 0
        count -= 1;
    }
```

when trying to decrementing from the 0 value, although the state value still remains 0, it is getting a successful transaction, it should fail.

**Expected behavior:**

when trying to decrementing from the 0 value, the txn should revert with an error like that:

```
transact to Counter.dec errored: VM error: revert.

revert
	The transaction has been reverted to the initial state.
Note: The called function should be payable if you send value and the value you send should be less than your current balance.
Debug the transaction to get more information.
```

**Steps to reproduce:**

I used that contract:

```
// SPDX-License-Identifier: MIT
pragma solidity ^0.8.17;

contract Counter {
    uint public count;

    // Function to get the current count
    function get() public view returns (uint) {
        return count;
    }

    // Function to increment count by 1
    function inc() public {
        count += 1;
    }

    // Function to decrement count by 1
    function dec() public {
        // This function will fail if count = 0
        count -= 1;
    }
}

```

at the initial state, with count at 0, I tried to call `dec()`, it got a successful transaction.

**Other information:**

EVM environment: Kakarot ZKEVM blockchain (Testnet)
Solidity version: 0.8.17
No balance was sent to the contract upon deployment.

ClementWalter:
Time spent on this PR: 0.1

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

A test is marked as failed while it has the expected behavior
No test for safe sub.

Resolves #683

## What is the new behavior?

Fixed tests.
Behavior was already correct.

### DIFF
--- a/solidity_contracts/src/PlainOpcodes/Counter.sol
+++ b/solidity_contracts/src/PlainOpcodes/Counter.sol
@@ -17,17 +17,21 @@ contract Counter {
         count += 1;
     }
 
-    function decUnchecked() public {
-        unchecked {
-            count -= 1;
-        }
+    function dec() public greaterThanZero {
+        count -= 1;
     }
 
     function decInPlace() public greaterThanZero {
         count--;
     }
 
-    function dec() public greaterThanZero {
+    function decUnchecked() public {
+        unchecked {
+            count -= 1;
+        }
+    }
+
+    function decUnsafe() public {
         count -= 1;
     }
 
--- a/solidity_contracts/src/PlainOpcodes/Counter.sol
+++ b/solidity_contracts/src/PlainOpcodes/Counter.sol
@@ -17,22 +17,22 @@ contract Counter {
         count += 1;
     }
 
-    function dec() public greaterThanZero {
+    function dec() public {
         count -= 1;
     }
 
-    function decInPlace() public greaterThanZero {
+    function decInPlace() public {
         count--;
+    }
+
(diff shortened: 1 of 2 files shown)


solidity_contracts/src/PlainOpcodes/Counter.sol
```diff
@@ -17,17 +17,21 @@ contract Counter {
...
     }
 
-    function dec() public greaterThanZero {
+    function decUnchecked() public {
+        unchecked {
+            count -= 1;
+        }
+    }
+
+    function decUnsafe() public {
```
Eikix, line 34: what do we mean by unsafe? that it wraps around?
i never really looked into, i just remember SafeMath was integrated into Solidity 0.8 or smth and now integers are just "safe"

Eikix, line 34: Idk if unsafe is the right name for it, since it's behaving exaclty like "safemath.sol"

Eikix (approved):
lgtm

PR #828, opened 2023-11-24
Add memory expansion gas accounting

### LINKED ISSUE #827: memory expansion gas

ClementWalter:
Time spent on this PR: 0.5

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Only the static gas is taken into account

Resolves #827

## What is the new behavior?

Taken into accounting.


ClementWalter:
just realizing that I could use a another jmp label for CALLs and CREATEs instead of making this `if` guess condition, which also leverage the pc and consequently prevents from fixing it (ie incrementing after and not before).

I'll update ASAP though it doesn't change the logic of the PR

src/kakarot/instructions/system_operations.cairo
```diff
@@ -480,7 +478,7 @@ namespace CallHelper {
             read_only=read_only,
             is_create=FALSE,
         );
-        let sub_ctx = ExecutionContext.init(call_context);
+        let sub_ctx = ExecutionContext.init(call_context, 0);
```
Eikix, line 481: should we check above (in prepare_call?) that call_args.gas < ctx.gas_limit

ClementWalter: Indeed, even though not related to memory expansion, better do it rn. I’ll fix fist the tests and then just add this

src/kakarot/execution_context.cairo
```diff
@@ -248,9 +233,28 @@ namespace ExecutionContext {
     // @param self The pointer to the execution context.
     // @param inc_value The value to increment the gas used with.
     // @return ExecutionContext The pointer to the updated execution context.
-    func increment_gas_used(
+    func increment_gas_used{range_check_ptr}(
```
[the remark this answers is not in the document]
ClementWalter: I've copied the behavior from the execution-specs, where they have `charge_gas` that reverts if there is not enough gas, see https://github.com/ethereum/execution-specs/blob/3fe6514f2d9d234e760d11af883a47c1263eff51/src/ethereum/shanghai/vm/gas.py#L99

I don't really like these "checked" or "safe" prefix/suffix as what they mean is never 100% clear without reading the code. And you need to read the code, well, you don't need the prefix/suffix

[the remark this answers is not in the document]
Eikix: Sure but increment_gas_used when reading the name of the function, I can't imagine it has the ability to revert, because the word "increment" does not carry that meaning
"charge" makes sense because it carries the meaning of paying something to someone, and if one's balance is too low, charge should fail

src/kakarot/evm.cairo
```diff
@@ -700,6 +711,13 @@ namespace EVM {
     ) -> Summary* {
         alloc_locals;
 
+        // Compute intrinsic gas usage
+        // See https://www.evm.codes/about#gascosts
+        let count = Helpers.count_nonzeroes(nonzeroes=0, idx=0, arr_len=calldata_len, arr=calldata);
```
Eikix, line 716: why do we need to provide a nonzeroes argument to count_nonzeroes?
i think the function pre-existed this PR but can be good to simply erase this param?

src/kakarot/evm.cairo
```diff
@@ -709,9 +727,11 @@ namespace EVM {
             let (empty: felt*) = alloc();
             tempvar bytecode = calldata;
             tempvar calldata = empty;
+            tempvar intrinsic_gas = intrinsic_gas + 32000;
```
Eikix, line 730: Let's check on [revm](https://github.com/bluealloy/revm/blob/32e0edff78765eb9433b53946fc0d76bee8e08b3/crates/interpreter/src/gas/calc.rs#L340C65-L340C65) if we're changing enough, or [execution spec](https://github.com/ethereum/execution-specs/blob/master/src/ethereum/shanghai/fork.py#L689)

On second thought, we should add another issue since it's outside the scope of this PR

src/kakarot/instructions/system_operations.cairo
```diff
@@ -406,81 +418,78 @@ namespace CallHelper {
         let value = with_value * stack_value + (1 - with_value) * ctx.call_context.value;
         let args_offset = 2 ** 128 * popped[2 + with_value].high + popped[2 + with_value].low;
         let args_size = 2 ** 128 * popped[3 + with_value].high + popped[3 + with_value].low;
+        let ret_offset = 2 ** 128 * ret_offset_uint256.high + ret_offset_uint256.low;
+        let ret_size = 2 ** 128 * ret_size_uint256.high + ret_size_uint256.low;
+
+        // 2. Gas
+        let available_gas = ctx.call_context.gas_limit - ctx.gas_used;
```
Eikix, line 425: we should group this with the gas operations below for the code to be easier to read

src/kakarot/execution_context.cairo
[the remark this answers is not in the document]
ClementWalter: ok I go for `ctx.charge_gas` then

src/kakarot/evm.cairo
ClementWalter: I read the spec, and this is correct afaiu

Eikix: Yes it feels like it, but I keep seeing "initcode stipend" for create operations since shanghai, idk if it applies here
https://github.com/kkrt-labs/kakarot/pull/828
https://github.com/bluealloy/revm/blob/32e0edff78765eb9433b53946fc0d76bee8e08b3/crates/interpreter/src/gas/calc.rs#L379

ClementWalter: indeed possibly missing some parts. I just happened to (partially) fix the tx cost, but the goal of this pr is memory expansion

Eikix: Fine with me! Let's fix this in another PR

pyproject.toml
```diff
@@ -13,6 +13,7 @@ amarna = "^0.1.5"
 starknet-py = "0.17.1"
 py-evm = "^0.7.0a1"
 openzeppelin-cairo-contracts = "^0.6.1"
+ethereum = { git = "https://github.com/ethereum/execution-specs.git" }
```
Eikix, line 16: ahh i didnt know you could do this!

Eikix (approved):
lgtm

PR #839, opened 2023-12-01
Refactor/eoa new param

### LINKED ISSUE #829: refactor: EOA validate and execute receives a separate encoded tx calldata and signature (in tx_info)

- Currently the EOA receives a payload of eth_sendRawTransaction -> it has to: RLP::decode, extract sig, RLP::encode the calldata without sig, hash initial encoded payload, check signature, then validate.

### Describe Preferred Solution

```[tasklist]
### Tasks
- [ ] Pretend that the payload received in KKT0 is the RLP-encoded calldata of an Ethereum transaction (without the signature)
- [ ] Hash that RLP-encoded calldata, and use the tx_info.signature to do the ecdsa check
- [ ] Decode the RLP-encoded calldata, perform checks (nonce and chain_id)
- [ ] Validate
- [ ] In execute, RLP-decode and call Kakarot eth_send_transaction
``` 

### Additional Context

Check out Kakarot SSJ implementation:
https://github.com/kkrt-labs/kakarot-ssj/blob/main/crates/contracts/src/eoa.cairo

### If the feature request is approved, would you be willing to submit a PR? Yes

/ No _(Help can be provided if you need assistance submitting a PR)_

Eikix:
Time spent on this PR:1d

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Resolves #829 

## What is the new behavior?

- change eoa to now take an unsigned payload as calldata and signature in tx_info

### DIFF
--- a/scripts/utils/kakarot.py
+++ b/scripts/utils/kakarot.py
@@ -38,10 +38,11 @@ from scripts.constants import (
 from scripts.utils.starknet import call as _call_starknet
 from scripts.utils.starknet import fund_address as _fund_starknet_address
 from scripts.utils.starknet import get_contract as _get_starknet_contract
-from scripts.utils.starknet import get_deployments, int_to_uint256
+from scripts.utils.starknet import get_deployments
 from scripts.utils.starknet import invoke as _invoke_starknet
 from scripts.utils.starknet import wait_for_transaction
-from tests.utils.helpers import rlp_encode_tx
+from tests.utils.helpers import rlp_encode_signed_data
+from tests.utils.uint256 import int_to_uint256
 
 logging.basicConfig()
 logger = logging.getLogger(__name__)
@@ -310,12 +311,7 @@ async def eth_send_transaction(
         hex(evm_account.signer.private_key),
     )
 
-    # We invoke eth_send_transaction for clarity but in reality,
-    # The eoa.__execute__ will always call eth_send_transaction of Kakarot
-    r = int_to_uint256(evm_tx.r)
-    s = int_to_uint256(evm_tx.s)
-
-    encoded_unsigned_tx = rlp_encode_tx(typed_transaction.as_dict())
+    encoded_unsigned_tx = rlp_encode_signed_data(typed_transaction.as_dict())
 
     prepared_invoke = await evm_account._prepare_invoke(
         calls=[
@@ -332,7 +328,7 @@ async def eth_send_transaction(
     prepared_invoke = Invoke(
         version=prepared_invoke.version,
         max_fee=prepared_invoke.max_fee,
-        signature=[r["low"], r["high"], s["low"], s["high"], evm_tx.v],
+        signature=[*int_to_uint256(evm_tx.r), *int_to_uint256(evm_tx.s), evm_tx.v],
         nonce=prepared_invoke.nonce,
         sender_address=prepared_invoke.sender_address,
         calldata=prepared_invoke.calldata,
@@ -351,6 +347,8 @@ async def eth_send_transaction(
     if len(transaction_events) != 1:
         raise ValueError("Cannot locate the single event giving the actual tx status")
     (
+        _msg_hash_low,
+        _msg_hash_high,
         response_len,
         *response,
         success,
--- a/scripts/utils/kakarot.py
+++ b/scripts/utils/kakarot.py
@@ -15,6 +15,7 @@ from hexbytes import HexBytes
 from starknet_py.net.account.account import Account
 from starknet_py.net.client_errors import ClientError
 from starknet_py.net.client_models import Call, Event
+from starknet_py.net.models.transaction import Invoke
 from starknet_py.net.signer.stark_curve_signer import KeyPair
 from starkware.starknet.public.abi import starknet_keccak
 from web3 import Web3
@@ -37,9 +38,10 @@ from scripts.constants import (
 from scripts.utils.starknet import call as _call_starknet
 from scripts.utils.starknet import fund_address as _fund_starknet_address
 from scripts.utils.starknet import get_contract as _get_starknet_contract
-from scripts.utils.starknet import get_deployments
+from scripts.utils.starknet import get_deployments, int_to_uint256
 from scripts.utils.starknet import invoke as _invoke_starknet
 from scripts.utils.starknet import wait_for_transaction
+from tests.utils.helpers import rlp_encode_tx
 
 logging.basicConfig()
 logger = logging.getLogger(__name__)
@@ -287,32 +289,57 @@ async def eth_send_transaction(
 ):
     """Execute the data at the EVM contract to on Kakarot."""
     evm_account = caller_eoa or await get_eoa()
+    nonce = await evm_account.get_nonce()
 
-    typed_transaction = TypedTransaction.from_dict(
-        {
-            "type": 0x2,
-            "chainId": KAKAROT_CHAIN_ID,
-            "nonce": await evm_account.get_nonce(),
-            "gas": gas,
-            "maxPriorityFeePerGas": int(1e19),
-            "maxFeePerGas": int(1e19),
-            "to": to_checksum_address(to) if to else None,
-            "value": value,
-            "data": data,
-        }
-    )
+    payload = {
+        "type": 0x2,
+        "chainId": KAKAROT_CHAIN_ID,
+        "nonce": nonce,
+        "gas": gas,
+        "maxPriorityFeePerGas": int(1e19),
+        "maxFeePerGas": int(1e19),
+        "to": to_checksum_address(to) if to else None,
+        "value": value,
+        "data": data,
+    }
+
+    typed_transaction = TypedTransaction.from_dict(payload)
+
     evm_tx = EvmAccount.sign_transaction(
         typed_transaction.as_dict(),
         hex(evm_account.signer.private_key),
     )
-    response = await evm_account.execute(
-        calls=Call(
-            to_addr=0xDEAD,  # unused in current EOA implementation
-            selector=0xDEAD,  # unused in current EOA implementation
-            calldata=evm_tx.rawTransaction,
-        ),
+
+    # We invoke eth_send_transaction for clarity but in reality,
+    # The eoa.__execute__ will always call eth_send_transaction of Kakarot
+    r = int_to_uint256(evm_tx.r)
+    s = int_to_uint256(evm_tx.s)
+
+    encoded_unsigned_tx = rlp_encode_tx(typed_transaction.as_dict())
+
+    prepared_invoke = await evm_account._prepare_invoke(
+        calls=[
+            Call(
+                to_addr=0xDEAD,  # unused in current EOA implementation
+                selector=0xDEAD,  # unused in current EOA implementation
+                calldata=encoded_unsigned_tx,
+            )
+        ],
         max_fee=int(5e17) if max_fee is None else max_fee,
     )
+    # We need to reconstruct the prepared_invoke with the new signature
+    # And Invoke.signature is Frozen
+    prepared_invoke = Invoke(
+        version=prepared_invoke.version,
+        max_fee=prepared_invoke.max_fee,
+        signature=[r["low"], r["high"], s["low"], s["high"], evm_tx.v],
+        nonce=prepared_invoke.nonce,
+        sender_address=prepared_invoke.sender_address,
+        calldata=prepared_invoke.calldata,
+    )
+
+    response = await evm_account.client.send_transaction(prepared_invoke)
+
     await wait_for_transaction(tx_hash=response.transaction_hash)
     receipt = await CLIENT.get_transaction_receipt(response.transaction_hash)
     transaction_events = [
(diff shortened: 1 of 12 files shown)


scripts/utils/kakarot.py
```diff
@@ -287,32 +289,63 @@ async def eth_send_transaction(
...
+        "nonce": nonce,
+        "gas": gas,
+        "maxPriorityFeePerGas": int(1e19),
+        "maxFeePerGas": int(1e19),
+        "to": to_checksum_address(to) if to else None,
+        "value": value,
+        "data": data,
+    }
+
+    typed_transaction = DynamicFeeTransaction.from_dict(payload)
```
ClementWalter, line 306: why did you change the tx type?

scripts/utils/kakarot.py
```diff
@@ -287,32 +289,63 @@ async def eth_send_transaction(
...
-            "gas": gas,
-            "maxPriorityFeePerGas": int(1e19),
-            "maxFeePerGas": int(1e19),
-            "to": to_checksum_address(to) if to else None,
-            "value": value,
-            "data": data,
-        }
-    )
     evm_tx = EvmAccount.sign_transaction(
         typed_transaction.as_dict(),
```
ClementWalter, line 309: you could just put the payload here, eventually we don't **need** this Transaction object

scripts/utils/kakarot.py
```diff
@@ -287,32 +289,63 @@ async def eth_send_transaction(
...
+    # We need to reconstruct the prepared_invoke with the new signature
+    # And Invoke.signature is Frozen
+    prepared_invoke = prepared_invoke.__class__(
+        version=prepared_invoke.version,
+        max_fee=prepared_invoke.max_fee,
+        signature=[r["low"], r["high"], s["low"], s["high"], evm_tx.v],
+        nonce=prepared_invoke.nonce,
+        sender_address=prepared_invoke.sender_address,
+        calldata=prepared_invoke.calldata,
+    )
```
ClementWalter, line 346: I'd just directly create the correct `Invoke` object, see for example `declare` [in ](https://github.com/Eikix/kakarot/blob/c084833739665ab746b6a833e01c37ef71318fd1/scripts/utils/starknet.py#L360)

scripts/utils/kakarot.py
```diff
@@ -287,32 +289,63 @@ async def eth_send_transaction(
...
+        "type": 0x2,
+        "chainId": KAKAROT_CHAIN_ID,
+        "nonce": nonce,
+        "gas": gas,
+        "maxPriorityFeePerGas": int(1e19),
+        "maxFeePerGas": int(1e19),
+        "to": to_checksum_address(to) if to else None,
+        "value": value,
+        "data": data,
+    }
```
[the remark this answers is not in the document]
Eikix: I wanted to build a util so that we don't have to care about the order of the keys and fields in the dict

scripts/utils/kakarot.py
Eikix: Yeah mb, I didn't change the tx type, but the constructor is more specialized, we don't need this though
TypedTransaction.from_dict, with tx type == 2 will create a DynamicFeeTransaction

scripts/utils/kakarot.py
Eikix: actually doing TypedTransaction.from_dict will pop the "type" key in the dict, I can also do it manually

scripts/utils/kakarot.py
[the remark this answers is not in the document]
Eikix: Updated the util to "just work"

tests/utils/helpers.py
```diff
@@ -17,6 +20,45 @@
...
+            typed_transaction.transaction.dictionary
+        )
+
+        # RPC-structured transaction to rlp-structured transaction
+        rlp_serializer = (
+            typed_transaction.transaction.__class__._unsigned_transaction_serializer
+        )
+        encoded_unsigned_tx = pipe(
+            rlp_serializer.from_dict(sanitized_transaction),  # type: ignore  # noqa: E501
+            lambda val: rlp.encode(val),  # rlp([...])
```
ClementWalter, line 37: you don't need to create a `lambda` here, but using just `rlp.encode` should work

tests/src/utils/test_eth_transaction.py
```diff
@@ -37,8 +42,27 @@ async def test_should_pass_all_transactions_types(
             private_key = generate_random_private_key()
             address = private_key.public_key.to_checksum_address()
             signed = Account.sign_transaction(transaction, private_key)
+
+            r = int_to_uint256(signed.r)
+            s = int_to_uint256(signed.s)
+
+            print(transaction)
```
ClementWalter, line 49: no print

tests/src/utils/test_eth_transaction.py
```diff
@@ -37,8 +42,27 @@ async def test_should_pass_all_transactions_types(
...
+                int(address, 16),
+                transaction["nonce"],
+                (
+                    r["low"],
+                    r["high"],
+                ),
+                (
+                    s["low"],
+                    s["high"],
+                ),
```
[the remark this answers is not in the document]
ClementWalter: I'd rather do

```
await eth_transaction.test__validate(
    int(address, 16),
    transaction["nonce"],
    int_to_uint256(signed.r),
    int_to_uint256(signed.s),
    signed.v,
    list(encoded_unsigned_tx),
)
```

to see the whole signature at once

tests/utils/helpers.py
```diff
@@ -17,6 +20,45 @@
...
+            lambda val: bytes(
+                [typed_transaction.transaction.__class__.transaction_type]
+            )
+            + val,  # (0x01 || rlp([...]))
+        )
+
```
ClementWalter, line 43: what's the diff with `transaction.payload()`? `serializable_unsigned_transaction_from_dict` from `eth_account/_utils/legacy_transactions.py` might also be just what you're looking for, handling all tx types

scripts/utils/kakarot.py
ClementWalter: let's use the object, but then use it 100%

src/utils/eth_transaction.cairo
```diff
@@ -324,26 +192,57 @@ namespace EthTransaction {
...
+            );
+            let (msg_hash) = cairo_keccak_bigend(inputs=words, n_bytes=tx_data_len);
+        }
+        finalize_keccak(keccak_ptr_start, keccak_ptr);
+
+        // Note: here, the validate process assumes an ECDSA signature, and r, s, v field
+        // Technically, the transaction type can determine the signature scheme.
+        let _is_legacy = is_legacy_tx(tx_data);
+        if (_is_legacy != FALSE) {
+            tempvar parity = (v - 2 * Constants.CHAIN_ID - 35);
```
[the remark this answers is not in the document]
Eikix: as discussed, it's fine this way

tests/utils/helpers.py
Eikix: the `payload` method includes the signature in it

```python
if not all(k in self.dictionary for k in "vrs"):
            raise ValueError("attempting to encode an unsigned transaction")
```


`serializable_unsigned_transaction_from_dict` isn't doing exactly what I want because in the case of a typed tx, it's simply returning an instance of the TypedTransaction object

scripts/utils/kakarot.py
```diff
@@ -287,32 +289,57 @@ async def eth_send_transaction(
...
+            )
+        ],
         max_fee=int(5e17) if max_fee is None else max_fee,
     )
+    # We need to reconstruct the prepared_invoke with the new signature
+    # And Invoke.signature is Frozen
+    prepared_invoke = Invoke(
+        version=prepared_invoke.version,
+        max_fee=prepared_invoke.max_fee,
+        signature=[r["low"], r["high"], s["low"], s["high"], evm_tx.v],
```
ClementWalter, line 335: `signature=[*int_to_uint256(evm_tx.r), *int_to_uint256(evm_tx.s), evm_tx.v]`

scripts/utils/kakarot.py
```diff
@@ -287,32 +289,57 @@ async def eth_send_transaction(
...
     )
-    response = await evm_account.execute(
-        calls=Call(
-            to_addr=0xDEAD,  # unused in current EOA implementation
-            selector=0xDEAD,  # unused in current EOA implementation
-            calldata=evm_tx.rawTransaction,
-        ),
+
+    # We invoke eth_send_transaction for clarity but in reality,
+    # The eoa.__execute__ will always call eth_send_transaction of Kakarot
```
ClementWalter, line 314: ? doesn't seem to be the case

src/kakarot/accounts/eoa/library.cairo
```diff
@@ -161,9 +163,8 @@ namespace ExternallyOwnedAccount {
 
         // See Argent account
         // https://github.com/argentlabs/argent-contracts-starknet/blob/c6d3ee5e05f0f4b8a5c707b4094446c3bc822427/contracts/account/ArgentAccount.cairo#L132
-        // Using here the EVM msg_hash instead of the Starknet tx.hash
         transaction_executed.emit(
-            msg_hash=msg_hash, response_len=return_data_len, response=return_data, success=success
+            response_len=return_data_len, response=return_data, success=success
```
ClementWalter, line 167: why did you remove the hash btw? it'd be necessary in case of a multicall

src/utils/eth_transaction.cairo
```diff
@@ -307,26 +193,43 @@ namespace EthTransaction {
...
+            // > Same as keccak, but outputs the hash in big endian representation.
+            // > Note that the input is still treated as little endian.
+            bytes_to_bytes8_little_endian(words, tx_data_len, tx_data);
+            let (msg_hash) = cairo_keccak_bigend(inputs=words, n_bytes=tx_data_len);
+
+            // Note: here, the validate process assumes an ECDSA signature, and r, s, v field
+            // Technically, the transaction type can determine the signature scheme.
+            let _is_legacy = is_legacy_tx(tx_data);
+            if (_is_legacy != FALSE) {
+                tempvar parity = (v - 2 * Constants.CHAIN_ID - 35);
```
ClementWalter, line 225: do this in `decode`, here the function should be version agnostic

tests/src/utils/test_eth_transaction.py
```diff
@@ -37,8 +42,16 @@ async def test_should_pass_all_transactions_types(
...
             address = private_key.public_key.to_checksum_address()
             signed = Account.sign_transaction(transaction, private_key)
+
+            encoded_unsigned_tx = rlp_encode_tx(transaction)
+
             await eth_transaction.test__validate(
-                int(address, 16), transaction["nonce"], list(signed["rawTransaction"])
+                int(address, 16),
+                transaction["nonce"],
+                (int_to_uint256(signed.r)["low"], int_to_uint256(signed.r)["high"]),
```
ClementWalter, line 51: just `int_to_uint256(signed.r)`

tests/src/utils/test_eth_transaction.py
```diff
@@ -5,9 +5,14 @@
 from eth_account.account import Account
 from starkware.starknet.testing.starknet import Starknet
 
+from scripts.utils.starknet import int_to_uint256
```
ClementWalter, line 8: use from tests utils

tests/utils/helpers.py
```diff
@@ -17,6 +17,45 @@
...
+            typed_transaction.transaction.__class__._unsigned_transaction_serializer
+        )
+        encoded_unsigned_tx = pipe(
+            rlp_serializer.from_dict(sanitized_transaction),
+            rlp.encode,
+            lambda val: bytes(
+                [typed_transaction.transaction.__class__.transaction_type]
+            )
+            + val,  # add the tx type right before the rlp[(...)]
+        )
```
ClementWalter, line 39: I'd rather do

```
encoded_unsigned_tx = [
    type_transaction.transaction_type,
    *rlp.encode(rlp_serializer.from_dict(sanitized_transaction))
]
```

tests/utils/helpers.py
```diff
@@ -17,6 +17,45 @@
 )
 
 
+def rlp_encode_tx(tx: dict) -> bytes:
```
ClementWalter, line 20: this is not really encoding the `tx` but rather the signed data

src/utils/eth_transaction.cairo
Eikix: If we do it in decode, it's less clean because, either:
- decode takes as an argument `v` and returns `y_parity`, then when we call decode in __execute__ we have to provide a `v` argument which is out of the scope of `execute`

src/kakarot/accounts/eoa/library.cairo
Eikix: it's not clear to me how we can leverage multicalls for now:
- we have only access to one signature now since it's in the tx_info (as opposed to one signature per call previously)
- `validate` isn't able to properly validate multiple signed data, so we would have to either:  
    - add many signatures in tx_info.signature field ([r,s,v, r,s,v, ..., r,s,v]), 1 sig per call, and `validate` would validate by jumping from 5 index to 5 index (very clunky imo)
    - sign once the array of calls (better), but we currently don't support this in wallets

ClementWalter (approved):

PR #840, opened 2023-12-05
Add casting utils

ClementWalter:
Time spent on this PR: 2

## Pull request type

Please check the type of change your PR introduces:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe):

## What is the current behavior?

Some casting utils in `utils.utils.Helpers`, sometimes redundant, sometimes with non consistant naming, sometime very expensive.

Resolves #835

## What is the new behavior?

New utils in dedicated files with EVM consistent naming and signature.

PR #841, opened 2023-12-05
Refacto casting utils

ClementWalter:
Following partial #840 merged after CI went ✅

### DIFF
--- a/src/utils/bytes.cairo
+++ b/src/utils/bytes.cairo
@@ -1,11 +1,12 @@
 from starkware.cairo.common.alloc import alloc
-from starkware.cairo.common.math import unsigned_div_rem, split_int, assert_nn
+from starkware.cairo.common.math import unsigned_div_rem
 from starkware.cairo.common.uint256 import Uint256
 from starkware.cairo.common.memcpy import memcpy
+from starkware.cairo.common.registers import get_label_location
 
 from utils.array import reverse
 
-func felt_to_ascii{range_check_ptr}(n: felt) -> (ascii_len: felt, ascii: felt*) {
+func felt_to_ascii{range_check_ptr}(dst: felt*, n: felt) -> felt {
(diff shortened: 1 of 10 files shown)


src/utils/bytes.cairo
```diff
@@ -32,8 +33,77 @@ func felt_to_ascii{range_check_ptr}(n: felt) -> (ascii_len: felt, ascii: felt*)
...
 
-    let (rev: felt*) = alloc();
-    reverse(rev, ascii_len, ascii);
+    reverse(dst, ascii_len, ascii);
 
-    return (ascii_len, rev);
+    return ascii_len;
+}
+
+func bytes_to_bytes8_little_endian(dst: felt*, bytes_len: felt, bytes: felt*) -> felt {
```
Eikix, line 41: why are we returning a felt, and then not using it?

src/utils/bytes.cairo
```diff
@@ -32,8 +33,77 @@ func felt_to_ascii{range_check_ptr}(n: felt) -> (ascii_len: felt, ascii: felt*)
...
+    }
+
+    let (local pow256) = get_label_location(pow256_table);
+
+    tempvar dst_index = 0;
+    tempvar bytes_index = bytes_len - 1;
+    tempvar bytes8 = 0;
+    tempvar bytes8_index = 7;
+
+    body:
```
Eikix, line 55: this is cairo 0 dark magic haha ;DDD
from the function definition point of view, the signature is way cleaner

but it's slighty harder to read than last version 👀

src/utils/bytes.cairo
ClementWalter: looks like it's not used all the time. This function is just used with keccak things requiring the intput to be encoded as such, but using the initial bytes_len

Eikix (approved):

PR #843, opened 2023-12-06
Refacto CREATEs opcodes

ClementWalter:


src/kakarot/instructions/system_operations.cairo
```diff
@@ -39,63 +39,124 @@ namespace SystemOperations {
...
+        );
+
+        if (ctx.reverted != FALSE) {
+            let ctx = ExecutionContext.update_stack(ctx, stack);
+            return ctx;
+        }
+
+        // Get message call gas
+        let available_gas = ctx.call_context.gas_limit - ctx.gas_used;
+        let (gas_limit, _) = unsigned_div_rem(available_gas, 64);
```
Eikix, line 76: why is this called gas limit ? it's 1/63th of the available gas right?

Eikix (commented):
first review i'll spend more time tomorrow

src/kakarot/instructions/system_operations.cairo
ClementWalter: it's just that I cannot put in the same line `available_gas - unsigned_div_rem(available_gas, 64).quotient` so I've found this convenient to just put the target name directly though it's still partial computation

PR #846, opened 2023-12-08
Use implicit args

### LINKED ISSUE #845: use implicit args

ClementWalter:
Time spent on this PR: 1

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

We don't leverage implicit args, which causes the code to be:

- heavier as we need to keep track of the stack, memory and state update manually
- error prone: `Wrong dict pointer` is raised when the pointer is not tracked successfully

Resolves #845

## What is the new behavior?

Use implicit args for Stack, Memory and State.
Update naming to be closed to the EELS.


greged93 (dismissed):
lgtm

PR #866, opened 2024-01-08
Remove RLP.encode and use optimized bytes_to_felt in decode

ClementWalter:
Time spent on this PR: 0.15

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

The EthTransaction library use a util `bytes_to_felt` which is indeed the same as
`load_word`, but ~8x less efficient.

The RLP library has unused encode functions

## What is the new behavior?

Just use the optimize version (renamed to `bytes_to_felt` to be more consistent with the project naming).


blockchain-tests-skip.yml
```diff
@@ -7131,8 +7131,9 @@ testname:
     - randomStatetest583_d0g0v0_Shanghai
   stQuadraticComplexityTest:
     - Call20KbytesContract50_1_d0g1v0_Shanghai
-    - Call20KbytesContract50_3_d0g1v0_Shanghai
     - Call20KbytesContract50_2_d0g1v0_Shanghai
+    - Call20KbytesContract50_3_d0g0v0_Shanghai
```
enitrat, line 7135: was this passing before and now no longer?

enitrat (approved):

blockchain-tests-skip.yml
ClementWalter: it was passing because of `RunResources out of steps` but we increased the step limit in the ef-test repo from 20% to 50M, and it's now running until the end, but failing

PR #867, opened 2024-01-09
feat: valid jumpdests

### LINKED ISSUE #693: bug: JUMPI Operation Allows Invalid Jump Destinations
# Bug Report

**Kakarot version:**

**Current behavior:**

The EVM implementation incorrectly allows jumps to destinations that should be invalid. Specifically, it allows jumps to a byte position where the value is `0x5B` but that `0x5B` is not actually an opcode, but rather part of the data of another opcode. This allows contracts to behave in unexpected ways, including potential security risks.

**Expected behavior:**

Jumps should only be allowed to valid `JUMPDEST` locations. A `JUMPDEST` is considered valid not just if the byte at that position is `0x5B`, but also if it is actually intended to be a `JUMPDEST` opcode, rather than being part of the data or argument of another opcode.

compare geth's `validJumpdest`: 
https://github.com/ethereum/go-ethereum/blob/60ec41ce73ee22452cc6f97a8e94797a42f1da00/core/vm/contract.go#L85-L97 

to ours: https://github.com/kkrt-labs/kakarot/blob/ee6458a7089f169334a7dba9665bcc5551507762/src/kakarot/execution_context.cairo#L938-L948

in particular the predicate `isCode` that enforces the case that the program counter must point to an opcode and not an argument for another opcode: https://github.com/ethereum/go-ethereum/blob/master/core/vm/contract.go#L101

**Steps to reproduce:**

- clone ef-tests: https://github.com/kkrt-labs/ef-tests
- generate katana dump of kakarot system and move the .katana folder to your ef-tests directory
- run 
```
RUN_SPECIFIC_TESTS=jumpi_d14g0v0_Shanghai cargo test -p ef-testing --features ef-tests -- --nocapture
```
in particular this case is failing: https://github.com/ethereum/tests/blob/develop/src/GeneralStateTestsFiller/VMTests/vmIOandFlowOperations/jumpiFiller.yml#L238

failing output:

```
[!] Case /Users/jmsb/exps/internet-computers/starknet/ef-tests/crates/ef-testing/ethereum-tests/BlockchainTests/GeneralStateTests/VMTests/vmIOandFlowOperations/jumpi.json failed (description: jumpi): Test failed: from address 0xcccccccccccccccccccccccccccccccccccccccc:

Test failed: failed test jumpi_d14g0v0_Shanghai: expected storage value 2989, got 0x000000000000000000000000000000000000000000000000000000000000600d

```

enitrat:
Time spent on this PR: 1d

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Resolves #693 

## What is the new behavior?

- JUMP and JUMPI can only jump to valid jumpdests
- valid jumpdests are stored in the message struct once the bytecode to execute is known.
-

### DIFF
--- a/src/kakarot/account.cairo
+++ b/src/kakarot/account.cairo
@@ -386,6 +386,9 @@ namespace Account {
         let (local valid_jumpdests_start: DictAccess*) = default_dict_new(0);
         tempvar valid_jumpdests = valid_jumpdests_start;
         tempvar i = 0;
+        jmp body if bytecode_len != 0;
+        tempvar range_check_ptr = range_check_ptr;
+        jmp end;
 
         body:
         let bytecode_len = [fp - 4];
@@ -419,7 +422,10 @@ namespace Account {
         static_assert valid_jumpdests == [ap - 2];
         static_assert i == [ap - 1];
         jmp body if check_bound != 0;
+        tempvar range_check_ptr = range_check_ptr;
 
+        end:
+        tempvar range_check_ptr = [ap - 1];
         return (valid_jumpdests_start=valid_jumpdests_start, valid_jumpdests=valid_jumpdests);
     }
 }
--- a/src/kakarot/account.cairo
+++ b/src/kakarot/account.cairo
@@ -401,27 +401,27 @@ namespace Account {
         let is_opcode_ge_0x5f = is_le(0x5f, opcode);
         let is_opcode_le_0x7f = is_le(opcode, 0x7f);
         let is_push_opcode = is_opcode_ge_0x5f * is_opcode_le_0x7f;
-        let i = i + 1 + is_push_opcode * (opcode - 0x5f);
+        let next_i = i + 1 + is_push_opcode * (opcode - 0x5f);  // 0x5f is the first PUSHN opcode, opcode - 0x5f is the number of arguments.
 
         if (opcode == 0x5b) {
             dict_write{dict_ptr=valid_jumpdests}(i, TRUE);
             tempvar valid_jumpdests = valid_jumpdests;
-            tempvar i = i;
+            tempvar next_i = next_i;
             tempvar range_check_ptr = range_check_ptr;
         } else {
             tempvar valid_jumpdests = valid_jumpdests;
-            tempvar i = i;
+            tempvar next_i = next_i;
             tempvar range_check_ptr = range_check_ptr;
         }
 
-        tempvar check_bound = is_le(bytecode_len + 1, i);
+        tempvar is_not_done = is_le(next_i + 1, bytecode_len);
         tempvar range_check_ptr = range_check_ptr;
         tempvar valid_jumpdests = valid_jumpdests;
-        tempvar i = i;
+        tempvar i = next_i;
         static_assert range_check_ptr == [ap - 3];
         static_assert valid_jumpdests == [ap - 2];
         static_assert i == [ap - 1];
-        jmp body if check_bound != 0;
+        jmp body if is_not_done != 0;
 
         end:
         tempvar range_check_ptr = [ap - 3];
--- a/src/kakarot/account.cairo
+++ b/src/kakarot/account.cairo
@@ -368,65 +368,64 @@ namespace Account {
     }
 
     // @notice Initializes a dictionary of valid jump destinations in EVM bytecode.
-    // @param i The current index in the bytecode.
     // @param bytecode_len The length of the bytecode.
     // @param bytecode The EVM bytecode to analyze.
-    // @param valid_jumpdests A dictionary to be filled with valid jump destinations.
-    // @return valid_jumpdests The updated dictionary with valid jump destinations.
+    // @return (valid_jumpdests_start, valid_jumpdests) The starting and ending pointers of the valid jump destinations.
     //
     // @dev This function iterates over the bytecode from the current index 'i'.
     // If the opcode at the current index is between 0x5f and 0x7f (PUSHN opcodes) (inclusive),
     // it skips the next 'n_args' opcodes, where 'n_args' is the opcode minus 0x5f.
     // If the opcode is 0x5b (JUMPDEST), it marks the current index as a valid jump destination.
-    // The function then increments the index and recursively calls itself until it has processed the entire bytecode.
-    func init_valid_jumpdests{range_check_ptr}(bytecode_len: felt, bytecode: felt*) -> (
+    // It continues by jumping back to the body flag until it has processed the entire bytecode.
+    func get_jumpdests{range_check_ptr}(bytecode_len: felt, bytecode: felt*) -> (
         valid_jumpdests_start: DictAccess*, valid_jumpdests: DictAccess*
     ) {
         alloc_locals;
         let (local valid_jumpdests_start: DictAccess*) = default_dict_new(0);
+        tempvar range_check_ptr = range_check_ptr;
         tempvar valid_jumpdests = valid_jumpdests_start;
         tempvar i = 0;
         jmp body if bytecode_len != 0;
-        tempvar range_check_ptr = range_check_ptr;
+
+        static_assert range_check_ptr == [ap - 3];
         jmp end;
 
         body:
         let bytecode_len = [fp - 4];
         let bytecode = cast([fp - 3], felt*);
+        let range_check_ptr = [ap - 3];
         let valid_jumpdests = cast([ap - 2], DictAccess*);
         let i = [ap - 1];
 
-        let opcode = [bytecode + i];
+        tempvar opcode = [bytecode + i];
         let is_opcode_ge_0x5f = is_le(0x5f, opcode);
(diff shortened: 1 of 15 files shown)


src/kakarot/evm.cairo
```diff
@@ -169,14 +170,41 @@ namespace EVM {
...
+        if (is_valid_jumpdest == FALSE) {
             let (revert_reason_len, revert_reason) = Errors.invalidJumpDestError();
+            tempvar self = new model.EVM(
+                message=updated_message,
+                return_data_len=self.return_data_len,
+                return_data=self.return_data,
+                program_counter=self.program_counter,
+                stopped=self.stopped,
+                gas_left=self.gas_left,
+                reverted=self.reverted,
```
ClementWalter, line 200: since we create a new EVM from scratch, I think that we can integrate directly here the updates from the `.stop` to avoid creating a new object two times

src/kakarot/evm.cairo
```diff
@@ -169,14 +170,41 @@ namespace EVM {
             return evm;
         }
 
-        if ([self.message.bytecode + new_pc_offset] != 0x5b) {
+        let valid_jumpdests = self.message.valid_jumpdests;
+        let (is_valid_jumpdest) = dict_read{dict_ptr=valid_jumpdests}(new_pc_offset);
+        tempvar updated_message = new model.Message(
```
ClementWalter, line 175: I'd call this only `message`, since the original message is `self.message`

src/kakarot/interpreter.cairo
```diff
@@ -734,21 +736,34 @@ namespace Interpreter {
...
         } else {
-            tempvar bytecode = bytecode;
-            tempvar calldata = calldata;
-            tempvar intrinsic_gas = intrinsic_gas;
+            assert bytecode = tmp_bytecode;
+            assert calldata = tmp_calldata;
+            assert intrinsic_gas = tmp_intrinsic_gas;
         }
 
+        let (valid_jumpdests_start: DictAccess*) = default_dict_new(0);
```
ClementWalter, line 757: it's not required afaik to add the type here in a `let`

src/utils/array.cairo
```diff
@@ -81,3 +81,33 @@ func slice{range_check_ptr}(dst: felt*, data_len: felt, data: felt*, offset: fel
...
+    tempvar i = 0;
+    tempvar is_found = FALSE;
+
+    body:
+    let arr_len = [fp - 5];
+    let arr = cast([fp - 4], felt*);
+    let value = [fp - 3];
+    let i = [ap - 1];
+
+    tempvar check_value = ([arr + i] - value);
```
ClementWalter, line 101: remove useless parenthesis

src/utils/utils.cairo
```diff
@@ -590,4 +591,43 @@ namespace Helpers {
...
+    // @param bytecode The EVM bytecode to analyze.
+    // @param valid_jumpdests A dictionary to be filled with valid jump destinations.
+    // @return valid_jumpdests The updated dictionary with valid jump destinations.
+    //
+    // @dev This function iterates over the bytecode from the current index 'i'.
+    // If the opcode at the current index is between 0x5f and 0x7f (PUSHN opcodes) (inclusive),
+    // it skips the next 'n_args' opcodes, where 'n_args' is the opcode minus 0x5f.
+    // If the opcode is 0x5b (JUMPDEST), it marks the current index as a valid jump destination.
+    // The function then increments the index and recursively calls itself until it has processed the entire bytecode.
+    func init_valid_jumpdests{range_check_ptr, valid_jumpdests: DictAccess*}(
```
ClementWalter, line 607: refactor this without tail recursion, and remove the `i` from the args. The signature should not get the destination dict as implicit arg, but the function should create the dict and return it.

This function could also be moved to accounts.cairo, there are already some `static` utils for accounts related functions

src/utils/utils.cairo
```diff
@@ -590,4 +591,43 @@ namespace Helpers {
...
+        i: felt, bytecode_len: felt, bytecode: felt*
+    ) {
+        let is_done = is_le(bytecode_len, i);
+        if (is_done != FALSE) {
+            return ();
+        }
+
+        let opcode = [bytecode + i];
+        let is_opcode_ge_0x5f = is_le_felt(0x5f, opcode);
```
ClementWalter, line 616: use `is_le_felt` only if there is a risk that the value is greater than the RANGE_CHECK (it's more expensice that `is_le`)

ClementWalter (changes requested):
see comments, you also need to fix the tests

src/utils/array.cairo
```diff
@@ -81,3 +81,33 @@ func slice{range_check_ptr}(dst: felt*, data_len: felt, data: felt*, offset: fel
...
+
+func contains{range_check_ptr}(arr_len: felt, arr: felt*, value: felt) -> felt {
+    alloc_locals;
+
+    if (arr_len == 0) {
+        return FALSE;
+    }
+
+    tempvar i = 0;
+    tempvar is_found = FALSE;
```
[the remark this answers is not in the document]
enitrat: ah indeed, it worked because the side effect was that [ap-1] was 0 anyway...

src/kakarot/account.cairo
```diff
@@ -366,6 +366,68 @@ namespace Account {
...
+    // @param bytecode The EVM bytecode to analyze.
+    // @param valid_jumpdests A dictionary to be filled with valid jump destinations.
+    // @return valid_jumpdests The updated dictionary with valid jump destinations.
+    //
+    // @dev This function iterates over the bytecode from the current index 'i'.
+    // If the opcode at the current index is between 0x5f and 0x7f (PUSHN opcodes) (inclusive),
+    // it skips the next 'n_args' opcodes, where 'n_args' is the opcode minus 0x5f.
+    // If the opcode is 0x5b (JUMPDEST), it marks the current index as a valid jump destination.
+    // The function then increments the index and recursively calls itself until it has processed the entire bytecode.
+    func init_valid_jumpdests{range_check_ptr}(bytecode_len: felt, bytecode: felt*) -> (
```
ClementWalter, line 382: could be `get_valid_jumpdest`, or even `get_jumpdests`, at least the `init` sounds like we're mutating something, which is not the case anymore

src/kakarot/account.cairo
```diff
@@ -366,6 +366,68 @@ namespace Account {
...
+            tempvar valid_jumpdests = valid_jumpdests;
+            tempvar i = i;
+        } else {
+            tempvar valid_jumpdests = valid_jumpdests;
+            tempvar i = i;
+        }
+
+        tempvar check_bound = is_le(bytecode_len, i);
+        tempvar valid_jumpdests = valid_jumpdests;
+        tempvar i = i;
```
ClementWalter, line 421: I think that we should iterate over bytecode_len and bytecode, without i, and `check_bound` would just be `jmp body if bytecode_len != 0`

src/kakarot/account.cairo
```diff
@@ -366,6 +366,67 @@ namespace Account {
...
+        } else {
+            tempvar valid_jumpdests = valid_jumpdests;
+            tempvar next_i = next_i;
+            tempvar range_check_ptr = range_check_ptr;
+        }
+
+        tempvar is_not_done = is_le(next_i + 1, bytecode_len);
+        tempvar range_check_ptr = range_check_ptr;
+        tempvar valid_jumpdests = valid_jumpdests;
+        tempvar i = next_i;
```
ClementWalter, line 420: is there a reason why we need to put `next_i` first and not `i` directly?

src/utils/utils.cairo
ClementWalter, line 1: remove unused imports

ClementWalter (changes requested):
we also need to squash the dict (`default_dict_finalize`) when we drop the message (end a given execution)

ClementWalter:
For the newly failing ef test, just add them in the skip file because it's out of resources

src/kakarot/interpreter.cairo
```diff
@@ -667,6 +667,9 @@ namespace Interpreter {
         Memory.finalize();
         Stack.finalize();
         State.finalize();
+        with evm {
```
ClementWalter, line 670: could have added the `evm` in the implicit, not a big deal but I feel like it'd make the whole thing lighter.

We can create another issue to put evm in impliciti everywhere if we feel so

src/kakarot/precompiles/precompiles.cairo
```diff
@@ -8,6 +8,7 @@ from starkware.cairo.common.memcpy import memcpy
 from starkware.cairo.common.cairo_builtins import HashBuiltin, BitwiseBuiltin
 from starkware.cairo.common.default_dict import default_dict_new
 from starkware.cairo.common.math_cmp import is_le, is_not_zero
+from starkware.cairo.common.dict_access import DictAccess
```
ClementWalter, line 11: unused

ClementWalter (approved):
still one unused import, let's merge anyway

PR #875, opened 2024-01-11
refactor: use evm_address instead of address

enitrat:
Time spent on this PR:

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Resolves #<Issue number>

## What is the new behavior?

-
-
-

### DIFF
--- a/src/backend/starknet.cairo
+++ b/src/backend/starknet.cairo
@@ -96,7 +96,7 @@ namespace Starknet {
 
     // @notice Populate a Environment with Starknet syscalls
     func get_env{syscall_ptr: felt*}(
-        origin: model.Address*, gas_price: felt
+        origin_evm_address: felt, gas_price: felt
     ) -> model.Environment* {
         alloc_locals;
         let (block_number) = get_block_number();
@@ -109,7 +109,7 @@ namespace Starknet {
             0xacdffe0cf08e20ed8ba10ea97a487004, 0x388ca486b82e20cc81965d056b4cdca
         );
         return new model.Environment(
-            origin=origin,
+            origin=origin_evm_address,
             gas_price=gas_price,
             chain_id=Constants.CHAIN_ID,
             prev_randao=Uint256(0, 0),
@@ -135,7 +135,7 @@ namespace Internals {
             return ();
         }
 
-        let (starknet_address) = Account.compute_starknet_address(accounts_start.key);
+        let starknet_address = Account.compute_starknet_address(accounts_start.key);
         let account = cast(accounts_start.new_value, model.Account*);
         _commit_account(account, starknet_address);
 
@@ -155,7 +155,7 @@ namespace Internals {
     ) {
         alloc_locals;
 
-        let starknet_account_exists = Account.is_registered(self.address);
+        let starknet_account_exists = Account.is_registered(self.address.evm);
 
         // Case new Account
         if (starknet_account_exists == 0) {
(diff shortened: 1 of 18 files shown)


tests/fixtures/EVM.cairo
```diff
@@ -95,7 +95,7 @@ func evm_call{
     let fp_and_pc = get_fp_and_pc();
     local __fp__: felt* = fp_and_pc.fp_val;
```
ClementWalter, line 96: you can probably remove this since you've removed the deref `&`

tests/fixtures/EVM.cairo
```diff
@@ -148,7 +148,7 @@ func evm_execute{
...
 
-    let env = Starknet.get_env(&origin, 0);
+    let env = Starknet.get_env(origin.evm, 0);
```
ClementWalter, line 151: same here

src/backend/starknet.cairo
```diff
@@ -96,7 +96,7 @@ namespace Starknet {
 
     // @notice Populate a Environment with Starknet syscalls
     func get_env{syscall_ptr: felt*}(
-        origin: model.Address*, gas_price: felt
+        origin_evm_address: felt, gas_price: felt
```
ClementWalter, line 99: I don't really like the `_evm_address` suffix everywhere. I think that we should consider that addresses are EVM address unless explicitely said to be something else

src/kakarot/instructions/system_operations.cairo
```diff
@@ -76,11 +76,9 @@ namespace SystemOperations {
         Memory.load_n(size.low, bytecode, offset.low);
 
         // Get target address
-        let evm_contract_address = CreateHelper.get_evm_address(
-            evm.message.address, popped_len, popped, size.low, bytecode
+        let target_evm_address = CreateHelper.get_evm_address(
```
ClementWalter, line 79: this is called `contract_address` in EELS so I think that we should either leave it as it was or just remove the `evm_` (see also my other comment regarding the fact that addresses should just be EVM by default, https://github.com/ethereum/execution-specs/blob/559ba54829a5d726ca95fe5701486f7bd303bf7f/src/ethereum/shanghai/vm/instructions/system.py#L163C5-L163C21)

src/kakarot/instructions/system_operations.cairo
```diff
@@ -548,25 +548,24 @@ namespace CallHelper {
             return child_evm;
         }
 
-        let (starknet_contract_address) = Account.compute_starknet_address(address);
-        tempvar call_address = new model.Address(starknet_contract_address, address);
-        let account = State.get_account(call_address);
+        let call_starknet_address = Account.compute_starknet_address(evm_address);
```
ClementWalter, line 551: unused

src/kakarot/instructions/system_operations.cairo
```diff
@@ -548,25 +548,24 @@ namespace CallHelper {
...
+        let called_account = State.get_account(evm_address);
```
ClementWalter, line 552: EELS uses `target` or `to`, we could use one or the other maybe?

tests/end_to_end/test_kakarot.py
```diff
@@ -7,6 +7,7 @@
 from starknet_py.net.client_models import TransactionStatus
 from starknet_py.net.full_node_client import FullNodeClient
 
+from scripts.utils.starknet import fund_address
```
ClementWalter, line 10: sorry I misled you but plz put the import inside the fixture so that we can keep the test file the same when/if we change the fixture

ClementWalter (approved):

PR #933, opened 2024-02-02
ci: added release workflow

d-roak:
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot/933)

### DIFF
--- a/.github/workflows/deployments.yml
+++ b/.github/workflows/deployments.yml
@@ -16,6 +16,7 @@ jobs:
       matrix:
         network: [testnet]
     env:
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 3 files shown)


.github/workflows/deployments.yml
```diff
@@ -16,6 +16,7 @@ jobs:
       matrix:
         network: [testnet]
     env:
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
d-roak, line 19: u need to confirm if this will work. I don't know if the token permissions are shared across repos

ClementWalter: this is fed by the github runner

d-roak: yep, but you are trying to use the token to get artifacts from `kakarot-ssj` right?

Eikix (approved):
lgtm

PR #949, opened 2024-02-05
Feat/fix deployer

### LINKED ISSUE #948: dev: reduce size of kakarot deployer image
- currently kakarot deployer is pulled from kakarot rpc for the docker compose of the kakarot zkevm entire system (EVM + RPC). it is 400mb!!! I think I see a big `COPY . .` that copies everything!!! Sure we can do that at build time, but ideally we should RM everything before end of build and only end up with the artifacts and that we want and the deploy scripts!

```[tasklist]
### Tasks
- [x] Fix the deployer dockerfile -> reduce the size of the image as much as possible!
- [ ] Bonus: delete unused docker images from the registry -> kakarot/katana and kakarot/devnet
```

Reward!

Eikix:
Time spent on this PR:

## Pull request type

- [ ] Bugfix
- [x] Feature
- [x] Build related changes

Resolves #948

- - -
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot/949)


ClementWalter:
I'm quite surprised that the final result is smaller because it looks like you copy everything from the builder. Isn't simply that you missed a `.dockerignore` ?

ClementWalter:
also, need to rebase and fix trunk

PR #959, opened 2024-02-07
dev: optimize class hash computation in deploy script

ClementWalter:
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot/959)


greged93 (dismissed):
lgtm

Makefile
```diff
@@ -79,7 +79,7 @@ build-sol:
 	forge build --names --force
 
 install-katana:
-	cargo install --git https://github.com/dojoengine/dojo --locked --rev e0054e7 katana
+	cargo install --git https://github.com/dojoengine/dojo --locked --rev dfe390 katana
 
 run-katana: install-katana
-	katana --validate-max-steps 16777216 --invoke-max-steps 16777216
+	katana --validate-max-steps 16777216 --invoke-max-steps 16777216 --gas-price 1 --disable-fee
```
Eikix, line 85: if we disable fee then one test end-to-end must be modified (one that uses actual fee from a tx to compute a balance of a wallet if iirc)

PR #979, opened 2024-02-14
fix: create finalization on REVERT opcode

enitrat:
Time spent on this PR:

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Resolves #<Issue number>

## What is the new behavior?

- Fixes the finalization of a parent context in a CREATE execution. We need to handle the case where the EVM has errored (exceptional or classic revert) before anything else.
-
-

### DIFF
--- a/src/kakarot/instructions/system_operations.cairo
+++ b/src/kakarot/instructions/system_operations.cairo
@@ -878,28 +878,17 @@ namespace CallHelper {
         let is_reverted = is_not_zero(evm.reverted);
         Stack.push_uint128(1 - is_reverted);
 
-        // Gas not used is returned when evm is not reverted
-        local gas_left;
-        local gas_refund;
-        if (evm.reverted == FALSE) {
-            assert gas_left = evm.message.parent.evm.gas_left + evm.gas_left;
-            assert gas_refund = evm.message.parent.evm.gas_refund + evm.gas_refund;
-        } else {
-            assert gas_left = evm.message.parent.evm.gas_left;
-            assert gas_refund = evm.message.parent.evm.gas_refund;
-        }
-
-        // If the call has halted exceptionnaly, the return_data is empty
-        // and nothing is copied to memory.
         if (evm.reverted == Errors.EXCEPTIONAL_HALT) {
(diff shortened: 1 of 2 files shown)


src/kakarot/instructions/system_operations.cairo
```diff
@@ -1074,6 +1063,30 @@ namespace CreateHelper {
     }(evm: model.EVM*) -> model.EVM* {
         alloc_locals;
 
+        // Reverted during execution - either REVERT or exceptional
+        if (evm.reverted != FALSE) {
+            let is_exceptional_revert = is_not_zero(Errors.REVERT - evm.reverted);
```
Eikix, line 1068: why would Errors.REVERT be equal to 1

src/kakarot/instructions/system_operations.cairo
```diff
@@ -1086,26 +1099,17 @@ namespace CreateHelper {
...
-            // The only case where a create opcode has returndata is if it reverted with the REVERT opcode.
-            tempvar is_exceptional_revert = is_not_zero(Errors.REVERT - evm.reverted) + (
-                1 - checks_success
-            );
-            let return_data_len = (1 - is_exceptional_revert) * evm.return_data_len;
-
             tempvar evm = new model.EVM(
                 message=evm.message.parent.evm.message,
-                return_data_len=return_data_len,
+                return_data_len=0,
```
Eikix, line 1112: what is happening in this case?
success is false, and revert is also false?

Eikix (commented):
not sure how to validate this PR, except that more tests pass and none others are broken

src/kakarot/instructions/system_operations.cairo
enitrat: REVERT is 1, EXCEPTIONAL_HALT is 2

src/kakarot/instructions/system_operations.cairo
enitrat: the execution of the create was successful, but the checks before writing the code in the destination accounts were not.

PR #982, opened 2024-02-15
skip collisions

enitrat:
Time spent on this PR:

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Resolves #<Issue number>

## What is the new behavior?

-
-
-


enitrat:
> how can an EOA have code?

Well, in our case, this should never happen. It's an invariant of the EVM: an EOA cannot have code. There are several tests in the EF-test suite asserting this.

PR #991, opened 2024-02-21
Fetch ssj artifacts for end to end in ci

ClementWalter:
`make setup` does fetch them but may be skipped if no deps changed

- - -
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot/991)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -141,9 +141,14 @@ jobs:
         uses: foundry-rs/foundry-toolchain@v1
         with:
           version: nightly
-      - name: Retrieve cached dependencies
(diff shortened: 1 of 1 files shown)


d-roak:
suggestion. add the fetch-artifacts as a dep in the makefile for the `make test-end-to-end` instead

ClementWalter:
> suggestion. add the fetch-artifacts as a dep in the makefile for the `make test-end-to-end` instead

pb is that locally I use a symlink and don't fetch, so it would override this + you don't want to dl again and again when you run them.

there is probably a better way to link artifacts though

PR #1034, opened 2024-03-13
Optimize get_jumpdests

feltroidprime:
- add `is_le_unchecked` function
- exit the loop without using `is_le` at each iteration. Only done once at exit. 

Before : 
image

After : 
image

- - -
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot/1034)

### DIFF
--- a/src/kakarot/account.cairo
+++ b/src/kakarot/account.cairo
@@ -423,8 +423,8 @@ namespace Account {
         let i = [ap - 1];
 
         tempvar opcode = [bytecode + i];
-        let is_opcode_ge_0x5f = is_le(0x5f, opcode);
-        let is_opcode_le_0x7f = is_le(opcode, 0x7f);
+        let is_opcode_ge_0x5f = Helpers.is_le_unchecked(0x5f, opcode);
+        let is_opcode_le_0x7f = Helpers.is_le_unchecked(opcode, 0x7f);
         let is_push_opcode = is_opcode_ge_0x5f * is_opcode_le_0x7f;
         let next_i = i + 1 + is_push_opcode * (opcode - 0x5f);  // 0x5f is the first PUSHN opcode, opcode - 0x5f is the number of arguments.
 
@@ -439,7 +439,17 @@ namespace Account {
             tempvar range_check_ptr = range_check_ptr;
         }
 
-        tempvar is_not_done = is_le(next_i + 1, bytecode_len);
+        // Checks if i+1 > bytecode_len. In this case, ap is filled with 0.
+        tempvar a = bytecode_len - (next_i + 1);
+        %{ memory[ap] = 0 if 0 <= (ids.a % PRIME) < range_check_builtin.bound else 1 %}
+        ap += 1;
+        // is_not_done == 1 <=> i+1 <= bytecode_len.
+        if ([ap - 1] == 0) {
+            tempvar is_not_done = 1;
+        } else {
+            tempvar is_not_done = 0;
+        }
+
         tempvar range_check_ptr = range_check_ptr;
         tempvar valid_jumpdests = valid_jumpdests;
         tempvar i = next_i;
@@ -449,7 +459,11 @@ namespace Account {
         jmp body if is_not_done != 0;
 
         end:
-        tempvar range_check_ptr = [ap - 3];
+        let range_check_ptr = [ap - 3];
+        let i = [ap - 1];
+        // Verify that i+1 > bytecode_len to ensure loop terminated correctly.
+        let check = Helpers.is_le_unchecked(i + 1, bytecode_len);
+        assert check = 0;
         return (valid_jumpdests_start, valid_jumpdests);
     }
 
--- a/src/kakarot/account.cairo
+++ b/src/kakarot/account.cairo
@@ -439,31 +439,25 @@ namespace Account {
             tempvar range_check_ptr = range_check_ptr;
         }
 
-        // Checks if i+1 > bytecode_len. In this case, ap is filled with 0.
-        tempvar a = bytecode_len - (next_i + 1);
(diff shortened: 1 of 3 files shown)


src/kakarot/account.cairo
```diff
@@ -439,7 +439,7 @@ namespace Account {
             tempvar range_check_ptr = range_check_ptr;
         }
 
-        tempvar is_not_done = is_le(next_i + 1, bytecode_len);
+        tempvar is_not_done = bytecode_len - next_i;  // if is_not_done == 0, end is reached.
```
ClementWalter, line 442: If the bytecode is malformed this can lead to an overflow (for example you deploy a byecode terminating with a PUSH1 with a CREATE2)

tests/fixtures/starknet.py
```diff
@@ -230,11 +230,8 @@ def _factory(entrypoint, **kwargs) -> list:
             data = profile_from_tracer_data(tracer_data)
 
             output_path = (
-                str(
-                    request.node.path.parent
-                    / f"{request.node.path.stem}.{entrypoint}({(json.dumps(kwargs) if kwargs else '')})"
-                )[:220]
-                + ".pb.gz"
+                str(request.node.path.parent / f"{request.node.path.stem}.{entrypoint}")
```
ClementWalter, line 233: In case of a parametrized test, you want to have the args in the name to be able to identity the test case. Just shorten the 220 to something lore reasonable and that fits in your case ?

src/kakarot/account.cairo
feltroidprime: Ok
The exit condition is currently `next_i >= bytecode_len`. 
So you confirm a case where `next_i > bytecode_len` is possible ?

tests/fixtures/starknet.py
feltroidprime: Changed, removed the '(' as linux bash doesn't work with it.

feltroidprime:
@ClementWalter  force pushed the required changes

src/kakarot/account.cairo
```diff
@@ -439,7 +439,17 @@ namespace Account {
...
+        tempvar a = bytecode_len - (next_i + 1);
+        %{ memory[ap] = 0 if 0 <= (ids.a % PRIME) < range_check_builtin.bound else 1 %}
+        ap += 1;
+        // is_not_done == 1 <=> i+1 <= bytecode_len.
+        if ([ap - 1] == 0) {
+            tempvar is_not_done = 1;
+        } else {
+            tempvar is_not_done = 0;
+        }
+
```
ClementWalter, line 452: isn't it just `tempvar is_not_done = 1 - [ap - 1]` ?

ClementWalter: and maybe ?

```
        tempvar a = (next_i + 1) - bytecode_len;
        %{ memory[ap] = 0 if 0 <= (ids.a % PRIME) < range_check_builtin.bound else 1 %}
        ap += 1;
        let is_not_done = [ap - 1];
```

ClementWalter (commented):
looks good, just a possible simplification

src/kakarot/account.cairo
feltroidprime: > and maybe ?
> 
> ```
>         tempvar a = (next_i + 1) - bytecode_len;
>         %{ memory[ap] = 0 if 0 <= (ids.a % PRIME) < range_check_builtin.bound else 1 %}
>         ap += 1;
>         let is_not_done = [ap - 1];
> ```

The issue is that jmp if [... ] is always of the form jmp if x!=0, we cannot do jump if x==0, and unfortunately the whitelisted hint is in reverse...

src/utils/utils.cairo
```diff
@@ -904,4 +904,29 @@ namespace Helpers {
...
+        // Ensure lhs <= rhs
+        assert [range_check_ptr] = a;
+        ap += 2;  // Two memory holes for known_ap_change in case of false case: Two instructions more: -1*a, and (-1*a) - 1.
+        tempvar range_check_ptr = range_check_ptr + 1;
+        tempvar res = 1;
+        ret;
+
+        false:
+        // Ensure rhs < lhs
+        assert [range_check_ptr] = (-a) - 1;
```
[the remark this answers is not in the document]
feltroidprime: No, it's just added by cairo-format when you type -a it changes to (-a)

src/kakarot/account.cairo
ClementWalter: I am not sure to follow. I don't really care about the value, the decision is actually binary (0 or not 0).
I have written `a` to be your `-a` to have first something like

```
tempvar a = (next_i + 1) - bytecode_len;
%{ memory[ap] = 0 if 0 <= (ids.a % PRIME) < range_check_builtin.bound else 1 %}
ap += 1;
// is_not_done == 1 <=> i+1 <= bytecode_len.
if ([ap - 1] == 0) {
    tempvar is_not_done = 0;
} else {
    tempvar is_not_done = 1;
}
```

but eventually `is_not_done` is just used as you said, to compare to `0`, so this could as well be

```
tempvar is_not_done = [ap - 1];
```

and using `let` because it's not necessary to add a step here

ClementWalter:
profile002
final profiling

src/kakarot/account.cairo
ClementWalter:
```suggestion
        // Check termination with whitelisted hint
        // ap = 0 <=> i >= bytecode_len.
        tempvar a = next_i - bytecode_len;
        %{ memory[ap] = 0 if 0 <= (ids.a % PRIME) < range_check_builtin.bound else 1 %}
        ap += 1;
        let is_not_done = [ap - 1];
```

ClementWalter (approved):

PR #1036, opened 2024-03-15
feat: blobhash (0x49)

enitrat:
Time spent on this PR:

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Resolves #<Issue number>

## What is the new behavior?

- Adds support for the blobhash opcode.

> the blobhash opcode reads index from the top of the stack as big-endian uint256, and replaces it on the stack with tx.blob_versioned_hashes[index] if index < len(tx.blob_versioned_hashes), and otherwise with a zeroed bytes32 value. The opcode has a gas cost of HASH_OPCODE_GAS.

As we don't support blob transactions, we would be in thecase where `index < len(tx.blob_versioned_hashes)` holds at all times. Thus, we simply push 0 to the stack.

- - -
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot/1036)

### DIFF
--- a/src/kakarot/instructions/block_information.cairo
+++ b/src/kakarot/instructions/block_information.cairo
@@ -41,7 +41,7 @@ namespace BlockInformation {
         jmp chainid;
         jmp selfbalance;
         jmp basefee;
-        jmp invalid_opcode_0x49;
+        jmp blobhash;
         jmp blobbasefee;
 
         blockhash:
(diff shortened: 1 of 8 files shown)


src/kakarot/instructions/block_information.cairo
```diff
@@ -122,6 +123,12 @@ namespace BlockInformation {
         Stack.push_uint128(evm.message.env.base_fee);
         jmp end;
 
+        blobhash:
+        let stack = cast([fp - 6], model.Stack*);
+        Stack.pop();
+        Stack.push_uint128(0);
```
ClementWalter, line 129: shouldn't we have a similar strategy as for the blockhashes?

enitrat: see PR description

ClementWalter:
> As we don't support blob transactions, we would be in thecase where index < len(tx.blob_versioned_hashes) holds at all times. Thus, we simply push 0 to the stack.

I think that we want instead to support blob transaction (0x3), with a fee equal to 1559 (0x2). I don't see the drawback of this approach

PR #1037, opened 2024-03-15
feat: blobbasefee (0x4a)

enitrat:
Time spent on this PR:

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Resolves #<Issue number>

## What is the new behavior?

- Implements the `blobbasefee` opcode. Since we don't support blob txs, always returns 0.
-
-

- - -
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot/1037)

### DIFF
--- a/src/kakarot/instructions/block_information.cairo
+++ b/src/kakarot/instructions/block_information.cairo
@@ -41,6 +41,8 @@ namespace BlockInformation {
         jmp chainid;
         jmp selfbalance;
         jmp basefee;
+        jmp invalid_opcode_0x49;
+        jmp blobbasefee;
 
         blockhash:
         let range_check_ptr = [fp - 8];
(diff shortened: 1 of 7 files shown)


src/kakarot/instructions/block_information.cairo
```diff
@@ -122,6 +124,14 @@ namespace BlockInformation {
         Stack.push_uint128(evm.message.env.base_fee);
         jmp end;
 
+        invalid_opcode_0x49:
+        jmp end;
```
ClementWalter, line 128: this will just create a cairovm error down the line.

```
let stack = cast([fp - 6], model.Stack*);
tempvar stack = stack;
```

would make this opcode transparent, but we want to revert the EVM instead right?

src/kakarot/instructions/block_information.cairo
```diff
@@ -122,6 +124,14 @@ namespace BlockInformation {
...
+
+        blobbasefee:
+        let stack = cast([fp - 6], model.Stack*);
+        Stack.push_uint128(0);
```
ClementWalter, line 132: shouldn't we add a storage var like for the block base fee?

src/kakarot/instructions/block_information.cairo
enitrat: Basically this would never be reached because in the global opcodes jmp table 0x49 just jumps to invalid opcode.

But either way: 0x49 is blobhash so this will disappear after a rebase with the other pr

PR #1042, opened 2024-03-24
refactor: bytes_i_to_uint256

### LINKED ISSUE #897: dev: refacto bytes_i_to_uint256
1. doesn't follow the projet naming convention
    - arr_len before arr
    - bytes instead of bytes_i (bytes in solidity is an array of variable lenght)
1. can most probably be optimized (use unit tests with --profile-cairo for this)

obatirou:
Time spent on this PR:

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Resolves #897

## What is the new behavior?

- rename `bytes_i_to_uint256` to `bytes_big_endian_to_uint256`
- simplify boolean condition for `is_bytes_len_16_bytes_or_less`
- follow conventions by having `bytes_len` before `bytes`
- rename `compute_half_uint256` to `compute_half_uint256_from_bytes`
- improved readability for `pow256_rev`
- use `pow256_rev` instead of `pow` from starkware lib: from 2419 to 1261 steps

With `pow` from starkware lib
before opti

With `pow256_rev`
after opti

- - -
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot/1042)


enitrat:
hi @obatirou thanks for the PR, it's normal that CI is failing given that the runner was recently updated for the Cancun fork, but not the `main` branch of this repo! will take a look tomorrow

src/utils/utils.cairo
```diff
@@ -92,36 +91,36 @@ namespace Helpers {
...
         alloc_locals;
 
-        if (i == 0) {
+        if (bytes_len == 0) {
             let res = Uint256(0, 0);
             return res;
         }
 
-        let is_sequence_32_bytes_or_less = is_le(i, 32);
+        let is_bytes_len_32_bytes_or_less = is_le(bytes_len, 32);
```
ClementWalter, line 106: we don't need this actually, we assume here that the len <= 32 and the checks are made before. Especialmy we don't want the cairo VM to raise

src/utils/utils.cairo
```diff
@@ -92,36 +91,36 @@ namespace Helpers {
...
         }
 
-        let is_sequence_16_bytes_or_less = is_le(i, 16);
+        let is_bytes_len_16_bytes_or_less = is_le(bytes_len, 16);
 
         // 1 - 16 bytes
-        if (is_sequence_16_bytes_or_less != FALSE) {
-            let (low) = compute_half_uint256(val=val, i=i, res=0);
+        if (is_bytes_len_16_bytes_or_less == TRUE) {
+            let (low) = compute_half_uint256_from_bytes(bytes_len, bytes, 0);
```
ClementWalter, line 115: let's just use `bytes_to_felt` function and remove the `compute_half_uint256`

src/utils/utils.cairo
```diff
@@ -92,36 +91,36 @@ namespace Helpers {
...
+            assert is_bytes_len_32_bytes_or_less = 1;
         }
 
-        let is_sequence_16_bytes_or_less = is_le(i, 16);
+        let is_bytes_len_16_bytes_or_less = is_le(bytes_len, 16);
 
         // 1 - 16 bytes
-        if (is_sequence_16_bytes_or_less != FALSE) {
-            let (low) = compute_half_uint256(val=val, i=i, res=0);
+        if (is_bytes_len_16_bytes_or_less == TRUE) {
```
ClementWalter, line 114: in cairo, you want to compare to FALSE because comparing to anything else is always reframed by the compiler as something like `jmp if cond != 0`

src/utils/utils.cairo
```diff
@@ -159,12 +158,16 @@ namespace Helpers {
...
-            return (res=res + [val]);
+    func compute_half_uint256_from_bytes{range_check_ptr}(
+        bytes_len: felt, bytes: felt*, res: felt
+    ) -> (res: felt) {
+        if (bytes_len == 1) {
+            return (res=res + [bytes]);
         }
-        let (temp_pow) = pow(256, i - 1);
-        let (res) = compute_half_uint256(val + 1, i - 1, res + [val] * temp_pow);
+        let temp_pow = pow256_rev(16 - (bytes_len - 1));
```
ClementWalter, line 167: fyi if you do smth like `pow_rev(16 - i)` you can as well define `pow` and do the more natural `pow(i)`

ClementWalter (changes requested):

ClementWalter (changes requested):
please remove the `_big_endian` in the function name as in solidity this is the default behavior (ie. that uint256(bytes32) is big endian actually)

enitrat:
> please remove the _big_endian in the function name as in solidity this is the default behavior (ie. that uint256(bytes32) is big endian actually)

@ClementWalter In SSJ there are occurences where we need to use conversions from little-endian bytes. I think it's for precompiles though, so it might not be a direct concern here. But I still think it's good to have it :)

ClementWalter:
> > please remove the _big_endian in the function name as in solidity this is the default behavior (ie. that uint256(bytes32) is big endian actually)
> 
> @ClementWalter In SSJ there are occurences where we need to use conversions from little-endian bytes. I think it's for precompiles though, so it might not be a direct concern here. But I still think it's good to have it :)

there are some places where we specifically put `_little_` but by default without any precision I think that it's obvious that it's end. It's the solidity/yul behavior, it's part of our EVM ubiquitous languages

src/utils/utils.cairo
obatirou: Should be better now

src/utils/utils.cairo
obatirou: Done

src/utils/utils.cairo
obatirou: Did not know that thank you 
Revert back to the initial condition

ClementWalter (approved):

ClementWalter (approved):
can you run once again the profiling to check the final steps count?

ClementWalter:
image
final results, ie steps down from 2425 to 601, ie x4 improvmenet

PR #1043, opened 2024-03-25
refactor: account model

### LINKED ISSUE #1023: epic: refactor account model in kakarot

context:  simplify and make future-proof the account abstraction (EOA and CA) in kakarot
currently, CAs don't have versioning, and will turn out to be hard to upgrade in the future.

```[tasklist]
### Tasks
- [ ] equalize APIs between CA and EOA -> both CA and EOA will have the same functions, i.e. __validate__, __execute__, and fetch_bytecode, set_bytecode, etc.
- [ ] CA has __validate__ and __execute__ same as EOA, add the logic of https://eips.ethereum.org/EIPS/eip-3607 -> check code is empty
- [ ] replace proxy pattern with replace_class_hash pattern
- [ ] https://github.com/kkrt-labs/kakarot/issues/1045
- [ ] https://github.com/kkrt-labs/kakarot/issues/1044
- [ ] remove the get_type() entrypoint
```

enitrat:
Time spent on this PR: 2.5d 

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Resolves #1023

## What is the new behavior?

- Remove EOA / CA distinction => Merged in `generic_account`
- Remove proxy pattern and use `replace_class`.
- Ensure deterministic address calculation using an `uninitialized_account` class whose sole purpose is to be upgraded to the latest account version

- - -
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot/1043)


docs/general/decode_a_cairo_trace.md
```diff
@@ -89,7 +89,7 @@ from starkware.cairo.lang.vm.reconstruct_traceback import reconstruct_traceback
 
 build_dir = Path("build")
 programs = {}
-for contract_name in ["externally_owned_account", "kakarot", "contract_account", "EVM"]:
+for contract_name in ["uninitialized_account", "kakarot", "generic_account", "EVM"]:
     artifact = build_dir / f"{contract_name}.json"
```
ClementWalter, line 93: remove `"contract_account"` as well right?

src/kakarot/accounts/uninitialized_account.cairo
```diff
@@ -0,0 +1,43 @@
...
+func constructor{
+    syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr, bitwise_ptr: BitwiseBuiltin*
+}(kakarot_address: felt, evm_address: felt) {
+    Account_kakarot_address.write(kakarot_address);
+    Account_evm_address.write(evm_address);
+    return ();
+}
+
+// @title EVM smart contract account representation.
+
```
ClementWalter, line 26: remove line break

src/kakarot/accounts/uninitialized_account.cairo
```diff
@@ -0,0 +1,43 @@
+// SPDX-License-Identifier: MIT
+
+%lang starknet
+
+// Starkware dependencies
+from openzeppelin.access.ownable.library import Ownable
```
ClementWalter, line 6: remove unused

docs/general/decode_a_cairo_trace.md
enitrat: indeed, spotted that already

src/kakarot/accounts/uninitialized_account.cairo
```diff
@@ -0,0 +1,43 @@
...
+
+// Local dependencies
+from kakarot.accounts.library import GenericAccount, Account_kakarot_address, Account_evm_address
+from starkware.starknet.common.syscalls import get_tx_info, get_caller_address, replace_class
+from starkware.cairo.common.math import assert_le, unsigned_div_rem
+from starkware.cairo.common.alloc import alloc
+
+@constructor
+func constructor{
+    syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr, bitwise_ptr: BitwiseBuiltin*
```
ClementWalter, line 18: I don't think that we need the bitwise here

src/kakarot/accounts/uninitialized_account.cairo
```diff
@@ -0,0 +1,43 @@
+// SPDX-License-Identifier: MIT
+
+%lang starknet
+
+// Starkware dependencies
+from openzeppelin.access.ownable.library import Ownable
+from starkware.cairo.common.cairo_builtins import HashBuiltin, BitwiseBuiltin, SignatureBuiltin
+from starkware.cairo.common.uint256 import Uint256
```
ClementWalter, line 8: unused

src/kakarot/accounts/uninitialized_account.cairo
```diff
@@ -0,0 +1,43 @@
+// SPDX-License-Identifier: MIT
+
+%lang starknet
+
+// Starkware dependencies
+from openzeppelin.access.ownable.library import Ownable
+from starkware.cairo.common.cairo_builtins import HashBuiltin, BitwiseBuiltin, SignatureBuiltin
```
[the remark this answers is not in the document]
enitrat: it's used in `__execute__` (forgot to remove unused imports)

src/kakarot/accounts/uninitialized_account.cairo
```diff
@@ -0,0 +1,43 @@
...
+from starkware.cairo.common.uint256 import Uint256
+
+// Local dependencies
+from kakarot.accounts.library import GenericAccount, Account_kakarot_address, Account_evm_address
+from starkware.starknet.common.syscalls import get_tx_info, get_caller_address, replace_class
```
ClementWalter, line 12: remove unused items

src/kakarot/accounts/uninitialized_account.cairo
```diff
@@ -0,0 +1,43 @@
+// SPDX-License-Identifier: MIT
+
+%lang starknet
+
+// Starkware dependencies
```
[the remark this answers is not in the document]
enitrat: We do this in the entire cairo codebase already

src/kakarot/accounts/uninitialized_account.cairo
```diff
@@ -0,0 +1,43 @@
...
+from openzeppelin.access.ownable.library import Ownable
+from starkware.cairo.common.cairo_builtins import HashBuiltin, BitwiseBuiltin, SignatureBuiltin
+from starkware.cairo.common.uint256 import Uint256
+
+// Local dependencies
+from kakarot.accounts.library import GenericAccount, Account_kakarot_address, Account_evm_address
+from starkware.starknet.common.syscalls import get_tx_info, get_caller_address, replace_class
+from starkware.cairo.common.math import assert_le, unsigned_div_rem
```
ClementWalter, line 13: remove unused

src/kakarot/accounts/uninitialized_account.cairo
```diff
@@ -0,0 +1,43 @@
...
+from starkware.cairo.common.alloc import alloc
```
ClementWalter, line 14: remve unused

src/kakarot/accounts/uninitialized_account.cairo
```diff
@@ -0,0 +1,43 @@
...
+    Account_kakarot_address.write(kakarot_address);
+    Account_evm_address.write(evm_address);
+    return ();
+}
+
+// @title EVM smart contract account representation.
+
+// @notice Initializes the account with the given Kakarot and EVM addresses.
+// @param kakarot_address The address of the main Kakarot contract.
+// @param evm_address The EVM address of the account.
```
ClementWalter, line 29: fix docstring

src/kakarot/accounts/library.cairo
```diff
@@ -4,9 +4,12 @@ from openzeppelin.access.ownable.library import Ownable
 from starkware.cairo.common.alloc import alloc
 from starkware.cairo.common.bool import FALSE
 from starkware.cairo.common.cairo_builtins import HashBuiltin, BitwiseBuiltin
-from starkware.cairo.common.math import unsigned_div_rem, split_int
+from starkware.cairo.common.math import unsigned_div_rem, split_int, split_felt
+from starkware.cairo.common.memcpy import memcpy
 from starkware.cairo.common.registers import get_label_location
```
ClementWalter, line 9: unused

src/kakarot/accounts/library.cairo
```diff
@@ -15,64 +18,311 @@ from starkware.starknet.common.syscalls import (
...
     storage_write,
     StorageReadRequest,
+    CallContract,
+    get_tx_info,
+    get_contract_address,
+    get_caller_address,
+    replace_class,
 )
 from starkware.cairo.common.memset import memset
+from utils.eth_transaction import EthTransaction
```
ClementWalter, line 28: line break between external and internal deps (like in python)

src/kakarot/accounts/library.cairo
```diff
@@ -15,64 +18,311 @@ from starkware.starknet.common.syscalls import (
...
 }
 
 @storage_var
-func is_initialized_() -> (res: felt) {
+func Account_is_initialized() -> (res: felt) {
 }
 
 @storage_var
-func evm_address() -> (evm_address: felt) {
+func Account_evm_address() -> (evm_address: felt) {
```
ClementWalter, line 47: wondering if we should give a meaningful name to the output, or just res. And if a name, then replace `res: felt` I guess

src/kakarot/accounts/library.cairo
```diff
@@ -15,64 +18,311 @@ from starkware.starknet.common.syscalls import (
...
     }(kakarot_address: felt, _evm_address) {
-        let (is_initialized) = is_initialized_.read();
+        let (is_initialized) = Account_is_initialized.read();
         assert is_initialized = 0;
-        is_initialized_.write(1);
+        Account_is_initialized.write(1);
         Ownable.initializer(kakarot_address);
-        evm_address.write(_evm_address);
+        Account_evm_address.write(_evm_address);
+        Account_kakarot_address.write(kakarot_address);
```
ClementWalter, line 84: isn't the `Account_kakarot_address` a bit redundant with the `Ownable.owner`?

src/kakarot/accounts/library.cairo
```diff
@@ -15,64 +18,311 @@ from starkware.starknet.common.syscalls import (
...
+
         // Give infinite ETH transfer allowance to Kakarot
         let (native_token_address) = IKakarot.get_native_token(kakarot_address);
         let (infinite) = uint256_not(Uint256(0, 0));
```
ClementWalter, line 88: we can just hardcode `Uint256(0xffff..fff, 0xff.ff)`, maybe as a constant if you prefer like
```
const UINT256_MAX = ...
```
there may be one sowhere already

src/kakarot/accounts/library.cairo
```diff
@@ -15,64 +18,311 @@ from starkware.starknet.common.syscalls import (
...
+    }() -> (is_initialized: felt) {
+        let is_initialized: felt = Account_is_initialized.read();
+        return (is_initialized=is_initialized);
+    }
+
+    // EOA functions
+
+    // Constants
+    // @dev see utils/interface_id.py
+    const INTERFACE_ID = 0x68bca1a;
```
ClementWalter, line 143: I'm not sure we should still use this (which was not used anymore btw).
This was to mimic https://eips.ethereum.org/EIPS/eip-165 but it doesn't really make sense now to me as we just want to version accounts (afaiu)

src/kakarot/accounts/library.cairo
```diff
@@ -15,64 +18,311 @@ from starkware.starknet.common.syscalls import (
...
+
+    struct Call {
+        to: felt,
+        selector: felt,
+        calldata_len: felt,
+        calldata: felt*,
+    }
+
+    // Tmp struct introduced while we wait for Cairo to support passing `[Call]` to __execute__
```
ClementWalter, line 152: lol I know that this is the real initial comment, but at this stage, let's not call it `tmp` anymore sa there won't be any change!

src/kakarot/accounts/library.cairo
```diff
@@ -15,64 +18,311 @@ from starkware.starknet.common.syscalls import (
...
+    // @param calldata_len The length of the calldata.
+    // @param calldata The calldata.
+    func validate{
+        syscall_ptr: felt*,
+        pedersen_ptr: HashBuiltin*,
+        bitwise_ptr: BitwiseBuiltin*,
+        range_check_ptr,
+    }(call_array_len: felt, call_array: CallArray*, calldata_len: felt, calldata: felt*) -> () {
+        alloc_locals;
+        if (call_array_len == 0) {
```
ClementWalter, line 173: maybe add a comment to say that we do the check only at the end of the recursion

src/kakarot/accounts/uninitialized_account.cairo
```diff
@@ -0,0 +1,43 @@
...
+@external
+func initialize{
+    syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr, bitwise_ptr: BitwiseBuiltin*
+}(new_class: felt) {
```
ClementWalter, line 33: I don't understand what this new_class would be here, since both EOA and CA now share the same implementation right? we can just do everything in the constructor

src/kakarot/accounts/generic_account.cairo
```diff
@@ -0,0 +1,217 @@
...
+        response=response,
+    );
+    return (response_len, response);
+}
+
+// @dev Return true if the interface_id is supported
+// @dev TODO: check what interfaces the contract should support and maybe create one for a kakarot account
+// @param interface_id The interface Id to verify if supported
+@view
+func supports_interface{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr}(
```
ClementWalter, line 146: see my comment in the library, this is not used

src/kakarot/accounts/generic_account.cairo
```diff
@@ -0,0 +1,217 @@
...
+@view
+func bytecode_len{
+    syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr, bitwise_ptr: BitwiseBuiltin*
+}() -> (len: felt) {
+    let (len) = GenericAccount.bytecode_len();
+    return (len=len);
+}
+
+// @notice Store a key-value pair.
+// @param key The storage address, with storage_var being storage_(key: Uint256)
```
ClementWalter, line 187: so actualluy the `Account_storage` is not used directly but was left for hinting that the storage, though beging writen directly with the address (as it was easier from the main Kakarot contract) actually is this one.
So you should update the comment, and the hard coded constant in Kakarot account.cairo

If you think that this is unclear, we can think about another way.

src/kakarot/accounts/generic_account.cairo
```diff
@@ -0,0 +1,217 @@
...
+// @param value The bytes32 stored value.
+@external
+func write_storage{
+    syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr, bitwise_ptr: BitwiseBuiltin*
+}(storage_addr: felt, value: Uint256) {
+    return GenericAccount.write_storage(storage_addr, value);
+}
+
+// @notice Read a given storage key
+// @param key The storage address, with storage_var being storage_(key: Uint256)
```
ClementWalter, line 197: see comment above regarding `storage_`

src/backend/starknet.cairo
```diff
@@ -42,46 +43,54 @@ namespace Starknet {
...
             constructor_calldata=constructor_calldata,
             deploy_from_zero=0,
         );
-        assert constructor_calldata[0] = kakarot_address;
-        assert constructor_calldata[1] = evm_address;
-        IAccount.initialize(starknet_address, class_hash, 2, constructor_calldata);
+
+        // Properly initialize the account once created
+        let (account_class_hash) = Kakarot_generic_account_class_hash.read();
+        IUninitializedAccount.initialize(starknet_address, account_class_hash);
```
ClementWalter, line 90: I don't understant why we still need this two fold deployment. This was done originally to make sure that the deployed address doesn't depend on the type (EOA or CA). But now that they share the same class hash, we can avoid this.

Reply to self: we want to have an address independent of the account class hash to avoid future issue if we upgrade the account

src/kakarot/accounts/uninitialized_account.cairo
```diff
@@ -0,0 +1,43 @@
...
+    return ();
+}
+
+// @title EVM smart contract account representation.
+
+// @notice Initializes the account with the given Kakarot and EVM addresses.
+// @param kakarot_address The address of the main Kakarot contract.
+// @param evm_address The EVM address of the account.
+@external
+func initialize{
```
ClementWalter, line 31: docstring
maybe `new_class` can be renamed to `implementation_class`?

src/kakarot/accounts/uninitialized_account.cairo
ClementWalter: I feel like we could completely empty the constructor because we deploy+initialize in the same function. Like that the mapping would be completely future proof, with the Starknet address solely depending on a transparent class

src/kakarot/accounts/uninitialized_account.cairo
```diff
@@ -0,0 +1,43 @@
...
+
+%lang starknet
+
+// Starkware dependencies
+from openzeppelin.access.ownable.library import Ownable
+from starkware.cairo.common.cairo_builtins import HashBuiltin, BitwiseBuiltin, SignatureBuiltin
+from starkware.cairo.common.uint256 import Uint256
+
+// Local dependencies
+from kakarot.accounts.library import GenericAccount, Account_kakarot_address, Account_evm_address
```
ClementWalter, line 11: don't import things from kakarot otherwise the class of this account depends on the kakarot library, which might change

src/kakarot/account.cairo
```diff
@@ -116,24 +116,17 @@ namespace Account {
...
-            // There is no way to access the nonce of an EOA currently
-            // But putting 1 shouldn't have any impact and is safer than 0
-            // since has_code_or_nonce is used in some places to trigger collision
-            let account = Account.init(
-                address=address, code_len=0, code=bytecode, nonce=1, balance=balance_ptr
-            );
-            return account;
-        }
+        // Special case: empty accounts, deployed but with no data, from create-SD transactions
+        // must
```
ClementWalter, line 123: missing end of sentence

src/kakarot/constants.cairo
```diff
@@ -13,14 +13,8 @@ namespace Constants {
     const TRANSACTION_GAS_LIMIT = 1000000;
```
ClementWalter, line 13: I think that we could remove this as well

src/kakarot/library.cairo
```diff
@@ -250,7 +246,7 @@ namespace Kakarot {
 
     // @notice returns the EVM address associated to a Starknet account deployed by kakarot.
     //         Prevents cases where some Starknet account has an entrypoint get_evm_address()
-    //         but isn't part of Kakarot system
+    //         but isn't part of Kakarot systemstorage_
```
ClementWalter, line 249: looks like a replace all mistake

tests/utils/reporting.py
```diff
@@ -23,7 +23,7 @@
 # A mapping to fix the mismatch between the debug_info and the identifiers.
 _label_scope = {
     "kakarot.constants.opcodes_label": "kakarot.constants",
-    "kakarot.accounts.contract.library.internal.pow_": "kakarot.accounts.contract.library.internal",
+    "kakarot.accounts.library.internal.pow_": "kakarot.accounts.library.internal",
```
ClementWalter, line 26: have you just check with a `--profile-cairo` that it was working?

tests/utils/syscall_handler.py
ClementWalter, line 1: you can remove the 

```
# Set account types
# We set all account types to be CA (contract account) as the only difference is that
# with EOA it doesn't try to fetch the nonce from the syscall, while here we actually
# want to have the EOA with the patched nonce.
def _account_type(contract_address, calldata):
    return [int.from_bytes(b"CA", "big")]

account_type_selector = get_selector_from_name("account_type")
cls.patches[account_type_selector] = _account_type
```

src/kakarot/accounts/generic_account.cairo
enitrat: ah yeah I missed that comment, but correctly updated _storage_addr

src/backend/starknet.cairo
enitrat: FYI: we use the same model in SSJ

tests/utils/reporting.py
enitrat: image

Looks like yes

src/kakarot/accounts/uninitialized_account.cairo
enitrat: Keeping the constructor arguments for now. It makes sure that accounts are bound to a specific deployment of Kakarot, and that external users cannot wrongly deploy + initialize accounts.

src/kakarot/accounts/library.cairo
```diff
@@ -15,64 +18,311 @@ from starkware.starknet.common.syscalls import (
...
-// Define the number of bytes per felt
 const BYTES_PER_FELT = 31;
 
-// @title ContractAccount main library file.
-// @notice This file contains the EVM smart contract account representation logic.
-namespace ContractAccount {
+// @title Account main library file.
+// @notice This file contains the EVM account representation logic.
+// @dev: Both EOAs and Contract Accounts are represented by this contract.
+namespace GenericAccount {
```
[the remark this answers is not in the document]
enitrat: `Account` is already a "reserved" keyword for internal account representation. What other name do you suggest?

src/kakarot/accounts/library.cairo
enitrat: can be simplified to only keep Ownable_owner

src/kakarot/accounts/library.cairo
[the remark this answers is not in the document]
ClementWalter: mh indeed there is the "storage" type Account and the "contract" type Account. Does this create a name clash ?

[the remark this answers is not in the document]
enitrat: I dont think it will (both should not be imported in the same files) but I prefer not to have two similar namespaces in the codebase. perhaps `AccountContract`?

src/kakarot/accounts/uninitialized_account.cairo
ClementWalter: up

src/kakarot/accounts/uninitialized_account.cairo
```diff
@@ -0,0 +1,55 @@
...
+
+const INITIALIZE_SELECTOR = 0x79dc0da7c54b95f10aa182ad0a46400db63156920adb65eca2654c0945a463;  // sn_keccak('initialize')
+
+@storage_var
+func Account_evm_address() -> (evm_address: felt) {
+}
+
+@storage_var
+func Account_kakarot_address() -> (kakarot_address: felt) {
+}
```
ClementWalter, line 17: maybe add a comment here to say that we use on purpose storage_slot collision (meaning that these storage var are also defined in the library, but we don't want to import it, so we redefine them. Add also a comment in the library next to them so that if at some point we change the name, we don't forget to change it here as well)

We could as well use a deterministic hard coded storage address if this is more clear

src/kakarot/accounts/uninitialized_account.cairo
enitrat: I think the line break between the title of the contract and the first @notice of the first function should stay. More aesthetic

src/kakarot/accounts/uninitialized_account.cairo
enitrat: created a `shared_storage.cairo` file to make it very explicit. Both `library.cairo` and `uninitialized_account.cairo` will import these variables.

src/kakarot/accounts/uninitialized_account.cairo
```diff
@@ -0,0 +1,45 @@
+// SPDX-License-Identifier: MIT
+
+%lang starknet
+
+from starkware.cairo.common.cairo_builtins import HashBuiltin, BitwiseBuiltin
+from starkware.cairo.common.alloc import alloc
+from starkware.starknet.common.syscalls import get_caller_address, replace_class, library_call
+
+from kakarot.accounts.shared_storage import Account_evm_address, Account_kakarot_address
```
ClementWalter, line 9: I'm hesitating because any change to this file would invalidate the deterministic evm to starknet address. I'm not even sure if just a renaming wouldn't make it as well

ClementWalter: we need to make sure by compiling and computing class hash in different cases,

tests/src/kakarot/instructions/test_environmental_information.py
```diff
@@ -114,20 +113,28 @@ def test_gasprice(self, cairo_run):
             cairo_run("test__exec_gasprice")
 
     class TestExtCodeHash:
-        @SyscallHandler.patch("IERC20.balanceOf", lambda addr, data: [0, 1])
         @SyscallHandler.patch(
-            "IAccount.account_type", lambda addr, data: [int.from_bytes(b"CA", "big")]
+            "IERC20.balanceOf",
+            lambda sn_addr, data: (
+                [0, 1] if sn_addr == EXISTING_ACCOUNT_SN_ADDR else [0, 0]
```
ClementWalter, line 119: isn't is`data[0]  == EXISTING_ACCOUNT` ?

src/kakarot/accounts/uninitialized_account.cairo
ClementWalter: see my comment, just double triple check class hash impact if anything changes in this file (name, content)

tests/src/kakarot/instructions/test_environmental_information.py
enitrat: The goal is to test with balance=0 and balance=1 for EXISTING_ACCOUNT, otherwise for an non-existing account the balance is _always_ 0

src/kakarot/accounts/uninitialized_account.cairo
enitrat: Duplicated storage vars in both files instead.

ClementWalter (approved):

PR #1049, opened 2024-03-27
fast uint256_add | uint256_sub

feltroidprime:
Using whitelisted hints:
uint256_add : 22 steps -> 10 steps
uint256_sub : ~60 steps -> 13 steps.

- - -
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot/1049)

### DIFF
--- a/src/kakarot/instructions/stop_and_math_operations.cairo
+++ b/src/kakarot/instructions/stop_and_math_operations.cairo
@@ -78,7 +78,7 @@ func uint256_add{range_check_ptr}(a: Uint256, b: Uint256) -> (res: Uint256, carr
     }
 }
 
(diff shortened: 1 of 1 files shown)


src/kakarot/instructions/stop_and_math_operations.cairo
enitrat, line 1: shouldn't be implemented here, this is only for EVM instructions. Perhaps `src.utils.math`

ClementWalter (changes requested):
plz move the functions in `src/utils/uint256.cairo` and replace all usages in the codebase, we're going to save lots more!

also, plz add tests

PR #1055, opened 2024-04-01
feat: ensure transaction inputs are in range [0, 2e128]

### LINKED ISSUE #961: feat: ensure transaction inputs are in range [0, 2e128]

The calldata passed can be huge and calculating the intrinsic cost for that will overflow PRIME and wrap back to 0.
To avoid these situations where gas is not accounted for correctly, user inputs should be verified to not be above the accepted limits. By checking they're in range [0,2**128], we could avoid these issues.

There are //TODO related to this issue in interpreter.cairo

AryanGodara:
Time spent on this PR:

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

There are no overflow checks for the input calldata at the moment.

Resolves #961 
## What is the new behavior?

- Update interpreter.cairo, address TODO regarding calldata overflow
- Check if the intrinsic gas cost overflows and wraps back to 0
- Assert that user input is within accepted limits

- - -
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot/1055)

### DIFF
--- a/src/kakarot/interpreter.cairo
+++ b/src/kakarot/interpreter.cairo
@@ -781,13 +781,6 @@ namespace Interpreter {
         // Compute intrinsic gas usage
         // See https://www.evm.codes/about#gascosts
 
-        // Splits the unsigned integer lift of a field element into the higher 128 bit and lower 128 bit.
-        // The unsigned integer lift is the unique integer in the range [0, PRIME) that represents the field element.
-        // For example, if value=17 * 2^128 + 8, then high=17 and low=8.
-        /// So, high = 0 if value < 2^128, and high >= 1 if value >= 2^128.
-        let (calldata_len_high, calldata_len_low) = split_felt(calldata_len);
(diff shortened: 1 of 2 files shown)


enitrat:
an assertion that calldata_len < 2e128 should be enough I guess

actually I think it would make more sense to have this assert inside `eth_send_transaction.cairo`, function `func validate`. This way we can reject transactions at the validation step

AryanGodara:
> an assertion that calldata_len < 2e128 should be enough I guess
> 
> actually I think it would make more sense to have this assert inside `eth_send_transaction.cairo`, function `func validate`. This way we can reject transactions at the validation step
> 
> Even then I'm not actually sure the sequencer can handle such a transaction

I see, I'll implement this changes asap!

> an assertion that calldata_len < 2e128 should be enough I guess

So, we don't have to check the intrinsic gas cost for overflow?

PR #1058, opened 2024-04-01
feat: account versioning

### LINKED ISSUE #1044: add versioning to CAs and a view endpoint to query it

### LINKED ISSUE #1045: make sure CA and EOA can be upgraded

enitrat:
Time spent on this PR: 0.5d

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Resolves #1045 Resolves #1044

## What is the new behavior?

- Added a versioning mechanism to accounts
- 
-

- - -
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot/1058)

### DIFF
--- a/src/kakarot/accounts/library.cairo
+++ b/src/kakarot/accounts/library.cairo
@@ -8,7 +8,7 @@ from starkware.cairo.common.math import unsigned_div_rem, split_int, split_felt
 from starkware.cairo.common.memcpy import memcpy
 from starkware.cairo.common.uint256 import Uint256, uint256_not, uint256_add, uint256_le
 from starkware.cairo.common.math_cmp import is_le
-from starkware.cairo.common.math import assert_not_zero
+from starkware.cairo.common.math import assert_not_zero, assert_le
 from starkware.starknet.common.syscalls import (
     StorageRead,
     StorageWrite,
@@ -25,7 +25,7 @@ from starkware.starknet.common.syscalls import (
 )
 from starkware.cairo.common.memset import memset
 
-from kakarot.interfaces.interfaces import IERC20, IKakarot
+from kakarot.interfaces.interfaces import IERC20, IKakarot, IAccount
 from kakarot.errors import Errors
 from kakarot.constants import Constants
 from utils.eth_transaction import EthTransaction
@@ -73,6 +73,9 @@ const BYTES_PER_FELT = 31;
 // @notice This file contains the EVM account representation logic.
 // @dev: Both EOAs and Contract Accounts are represented by this contract.
 namespace AccountContract {
+    // 000.001.000
+    const VERSION = 000001000;
+
     // @notice This function is used to initialize the smart contract account.
     // @dev The `evm_address` and `kakarot_address` were set during the uninitialized_account creation.
     // Reading them from state ensures that they always match the ones the account was created for.
@@ -99,7 +102,7 @@ namespace AccountContract {
         return ();
     }
 
-    // @notice This function is used to upgrade the smart contract account.
+    // @notice Upgrades the implementation of the account.
     // @param new_class The new class of the account.
     func upgrade{
         syscall_ptr: felt*,
@@ -107,17 +110,20 @@ namespace AccountContract {
         range_check_ptr,
         bitwise_ptr: BitwiseBuiltin*,
     }(new_class: felt) {
+        alloc_locals;
         // Access control check. Only the EOA owner should be able to upgrade its contract.
         // with `supports_interface`
         internal.assert_only_self();
-        // TODO: only valid classes should be allowed to be upgraded. Add a validation on the new class interface.
         assert_not_zero(new_class);
+        // TODO: only valid classes should be allowed to be upgraded. Add a validation on the new class interface.
+
+        let (new_version) = IAccount.library_call_version(new_class);
+        internal.assert_version_upgrade(new_version);
         replace_class(new_class);
         Account_implementation.write(new_class);
         return ();
     }
 
-    @view
     func get_implementation{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr}() -> (
         implementation: felt
     ) {
@@ -445,6 +451,22 @@ namespace internal {
         with_attr error_message("Only the account itself can call this function") {
             assert caller = this;
         }
+        return ();
+    }
+
+    // @notice asserts that the new version is greater than the current one
+    // @param new_version The new version of the class.
+    // @param old_version The previous version of the class.
+    func assert_version_upgrade{range_check_ptr}(new_version: felt) {
+        let (major_new, rem_new) = unsigned_div_rem(new_version, 1000000);
+        let (minor_new, patch_new) = unsigned_div_rem(rem_new, 1000);
+
+        let (major_old, rem_old) = unsigned_div_rem(AccountContract.VERSION, 1000000);
+        let (minor_old, patch_old) = unsigned_div_rem(rem_old, 1000);
+
+        assert_le(major_old, major_new);
+        assert_le(minor_old, minor_new);
+        assert_le(patch_old, patch_new);
         return ();
     }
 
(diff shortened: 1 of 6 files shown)


src/kakarot/accounts/library.cairo
```diff
@@ -99,25 +102,28 @@ namespace AccountContract {
         return ();
     }
 
-    // @notice This function is used to upgrade the smart contract account.
+    // @notice Upgrades the implementation of the account.
```
ClementWalter, line 105: I think that trunk doesn't catch this on cairo file but it should be "Upgrade" (try it on a python docstring)

src/kakarot/accounts/library.cairo
```diff
@@ -448,6 +454,22 @@ namespace internal {
         return ();
     }
 
+    // @notice asserts that the new version is greater than the current one
+    // @param new_version The new version of the class.
+    // @param old_version The previous version of the class.
+    func assert_version_upgrade{range_check_ptr}(new_version: felt) {
```
ClementWalter, line 460: why wouldn't we want to downgrade an account?

tests/utils/syscall_handler.py
```diff
@@ -262,6 +264,25 @@ def storage_write(self, segments, syscall_ptr):
...
+                class_hash: felt,
+            }
+        """
+        class_hash = segments.memory[syscall_ptr + 1]
+        address = self.contract_address
+        self.mock_replace_class(
+            address=address,
+            class_hash=class_hash,
+        )
+        self.class_hash = class_hash
```
ClementWalter, line 284: why would you want to do this? + you are not updating here the class attribute, but an instance one, that's why the test doesn't pass

tests/utils/syscall_handler.py
```diff
@@ -262,6 +264,25 @@ def storage_write(self, segments, syscall_ptr):
...
+
+            struct ReplaceClass {
+                selector: felt,
+                class_hash: felt,
+            }
+        """
+        class_hash = segments.memory[syscall_ptr + 1]
+        address = self.contract_address
+        self.mock_replace_class(
+            address=address,
```
ClementWalter, line 281: I am not sure that we want to add the address here, as the mock was meant to record the actual syscall, and address is not part of it. I don't see the value added by this change

tests/src/kakarot/accounts/test_contract_account.py
```diff
@@ -123,3 +123,40 @@ def test_should_read_bytecode(self, cairo_run, bytecode, storage):
...
+            )
+            SyscallHandler.mock_replace_class.assert_called_once_with(
+                address=0xABDE1, class_hash=0xBEEF
+            )
+            SyscallHandler.mock_storage.assert_called_once_with(
+                address=get_storage_var_address("Account_implementation"), value=0xBEEF
+            )
+
+        # TODO: Does not work
+        # assert SyscallHandler.class_hash == 0xbeef
```
ClementWalter, line 147: doesn't pass because you are update a particular instance attribute, and not a class attribute. to me, the `assert_called_once_with` is enough, it's SNOS responsibility to actually replace the class hash

tests/src/kakarot/accounts/test_contract_account.py
```diff
@@ -123,3 +123,40 @@ def test_should_read_bytecode(self, cairo_run, bytecode, storage):
...
+            with cairo_error():
+                cairo_run("test__upgrade", new_class=0xBEEF)
+            assert SyscallHandler.class_hash == 0xC1A55
+
+        @patch.object(SyscallHandler, "caller_address", 0x9876)
+        @patch.object(SyscallHandler, "contract_address", 0x1234)
+        def test_should_fail_caller_not_self(self, cairo_run):
+            with cairo_error():
+                cairo_run("test__upgrade", new_class=0xBEEF)
+            assert SyscallHandler.class_hash == 0xC1A55
```
ClementWalter, line 162: here use assert mock has no call instead

tests/src/kakarot/accounts/test_contract_account.py
```diff
@@ -123,3 +123,40 @@ def test_should_read_bytecode(self, cairo_run, bytecode, storage):
...
+            )
+            SyscallHandler.mock_storage.assert_called_once_with(
+                address=get_storage_var_address("Account_implementation"), value=0xBEEF
+            )
+
+        # TODO: Does not work
+        # assert SyscallHandler.class_hash == 0xbeef
+
+        @SyscallHandler.patch(
+            "ICairo1Helpers.library_call_version", lambda class_hash, data: [1]
```
ClementWalter, line 150: it doesn't change anything, but the full name is `IAccount.library_call_version` and not `ICairo1Helpers`.

tests/src/kakarot/accounts/test_contract_account.py
```diff
@@ -123,3 +123,40 @@ def test_should_read_bytecode(self, cairo_run, bytecode, storage):
...
+        )
+        def test_should_fail_downgrade_account(self, cairo_run):
+            with cairo_error():
+                cairo_run("test__upgrade", new_class=0xBEEF)
+            assert SyscallHandler.class_hash == 0xC1A55
```
ClementWalter, line 155: assert mock has no call instead

tests/utils/syscall_handler.py
enitrat: This way we can track the replace classes applied to specific accounts?

src/kakarot/accounts/library.cairo
enitrat: We don't want users to upgrade to a previous version that might be deprecated

tests/src/kakarot/accounts/test_contract_account.py
enitrat: makes sense, yes

tests/utils/syscall_handler.py
enitrat: but it doesn't make sense indeed

src/kakarot/accounts/library.cairo
ClementWalter: to me it's more a UX question than a core limitation we want to introduce, let's ask what the other thinks about this, but really doesn't seem obvious to me that we want to enforce this

src/kakarot/accounts/library.cairo
```diff
@@ -99,25 +102,25 @@ namespace AccountContract {
...
         bitwise_ptr: BitwiseBuiltin*,
     }(new_class: felt) {
+        alloc_locals;
         // Access control check. Only the EOA owner should be able to upgrade its contract.
-        // with `supports_interface`
         internal.assert_only_self();
-        // TODO: only valid classes should be allowed to be upgraded. Add a validation on the new class interface.
         assert_not_zero(new_class);
+        let (new_version) = IAccount.library_call_version(new_class);
+        internal.assert_version_upgrade(new_version);
```
ClementWalter, line 118: I missed this initially but other `internal` namespaces are `Internals`

tests/src/kakarot/accounts/test_contract_account.py
```diff
@@ -123,3 +124,41 @@ def test_should_read_bytecode(self, cairo_run, bytecode, storage):
             calls = [call(address=address) for address in addresses]
             mock_storage.assert_has_calls(calls)
             assert output[:output_len] == list(bytecode)
+
+    class TestUpgrade:
+        @SyscallHandler.patch(
+            "IAccount.library_call_version", lambda class_hash, data: [2000]
+        )
+        @patch.object(SyscallHandler, "mock_replace_class", mock.MagicMock())
```
ClementWalter, line 132: may not be the best way to do it but there might also be something with the `SycallHandler.patch` decorator wrecking the usual flow

tests/utils/syscall_handler.py
ClementWalter: but the `self.contract_address` is not updated in `call_contract`, so there is no notion of address

enitrat: ok, yes I misunderstood this

tests/src/kakarot/accounts/test_contract_account.py
enitrat: I'm not sure I get what you mean

ClementWalter: https://docs.python.org/3/library/unittest.mock.html#patch I'd have expected something like

```
@patch('tests.utils.syscall_handler.SyscallHandler.mock_replace_class')
def test_name(self, mock_replace_class):
```

but it doesn't seem to work well, so there may be something with the `@SyscallHandler` decorator breaking the `@patch` one.

In any case, I think that the `mock.MagicMock()` param is not required, see https://docs.python.org/3/library/unittest.mock.html#patch-object

ClementWalter (dismissed):

PR #1065, opened 2024-04-03
feat: deploy account from zero

### LINKED ISSUE #1046: dev: decide a deployment strategy for UninitializedAccount
## Option 1:
empty the constructor because we deploy+initialize in the same function. Like that the mapping would be completely future proof, with the Starknet address solely depending on a transparent class.

Main drawback: Let's imagine a case where there is no calldata constructor. In this case it would be possible for a third party to deploy an UninitializedAccount by sending a deploy_tx starknet. If the initialize is not protected to only accept Kakarot from the source, then there could be someone triggering deploy + initialize who would not be Kakarot. This is only a problem if we `deploy_from_zero = true`

## Option 2
Keep Kakarot address and EVM address in the constructor of UninitializedAccount. This effectively binds the calculation of a contract address to the associated kakarot contract.

### LINKED ISSUE #1064: feat: deploy accounts from zero address

Deploying accounts from zero enables anyone to deploy Kakarot accounts.
This means that we can have external actors responsible only for deploying & registering cairo contracts without going through the main Kakarot contract. 
The deployed address depends on the constructor parameters, namely `kakarot_address` and `evm_address`.  These two vars are written to storage once the account is deployed (but uninitialized), and anyone can call `initialize` to start the initialization process.

This initialization process can be called trustlessly by anyone, because we can read the lastest account contract class hash from Kakarot directly to setup the account with the latest implementation.

enitrat:
Time spent on this PR: 0.3d

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Resolves #1064 Resolves #1046

## What is the new behavior?

- Deploys accounts from zero to allow anyone to deploy an account for Kakarot.
-
-

- - -
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot/1065)

### DIFF
--- a/src/kakarot/accounts/uninitialized_account.cairo
+++ b/src/kakarot/accounts/uninitialized_account.cairo
@@ -43,7 +43,6 @@ func constructor{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr
 func initialize{
     syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr, bitwise_ptr: BitwiseBuiltin*
 }() {
-    let (caller) = get_caller_address();
     let (kakarot_address) = Account_kakarot_address.read();
     let (implementation_class) = IKakarot.get_account_contract_class_hash(kakarot_address);
     replace_class(implementation_class);
--- a/src/kakarot/accounts/uninitialized_account.cairo
+++ b/src/kakarot/accounts/uninitialized_account.cairo
@@ -12,48 +12,32 @@ namespace IKakarot {
     }
 }
 
-// We are intentionally causing a storage_slot collision here,
-// by defining these variables in both `uninitialized_account` and `account_contract`.
-// We are defining them here instead of in the account library, so as to not depend
-// on content of the account library in uninitialized_account and ensure a fixed class hash.
-@storage_var
-func Account_evm_address() -> (evm_address: felt) {
-}
-
-@storage_var
-func Account_kakarot_address() -> (kakarot_address: felt) {
-}
-
 const INITIALIZE_SELECTOR = 0x79dc0da7c54b95f10aa182ad0a46400db63156920adb65eca2654c0945a463;  // sn_keccak('initialize')
 
 // @title Uninitialized Contract Account. Used to get a deterministic address for an account, no
 // matter the actual implementation class used.
+
+// @notice Deploys and initializes the account with the Kakarot and EVM addresses it was deployed with.
+// @param kakarot_address The address of the main Kakarot contract.
+// @param evm_address The address of the EVM contract.
 @constructor
 func constructor{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr}(
     kakarot_address: felt, evm_address: felt
 ) {
-    Account_kakarot_address.write(kakarot_address);
-    Account_evm_address.write(evm_address);
-    return ();
-}
-
-// @notice Initializes the account with the Kakarot and EVM addresses it was deployed with.
-// @param implementation_class The address of the main Kakarot contract.
-@external
-func initialize{
-    syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr, bitwise_ptr: BitwiseBuiltin*
-}() {
-    let (kakarot_address) = Account_kakarot_address.read();
     let (implementation_class) = IKakarot.get_account_contract_class_hash(kakarot_address);
-    replace_class(implementation_class);
 
     let (calldata) = alloc();
-    assert [calldata] = implementation_class;
+    assert calldata[0] = kakarot_address;
+    assert calldata[1] = evm_address;
+    assert calldata[2] = implementation_class;
+
     library_call(
         class_hash=implementation_class,
         function_selector=INITIALIZE_SELECTOR,
-        calldata_size=1,
+        calldata_size=3,
         calldata=calldata,
     );
+
+    replace_class(implementation_class);
     return ();
 }
--- a/src/kakarot/accounts/uninitialized_account.cairo
+++ b/src/kakarot/accounts/uninitialized_account.cairo
@@ -6,6 +6,12 @@ from starkware.cairo.common.cairo_builtins import HashBuiltin, BitwiseBuiltin
 from starkware.cairo.common.alloc import alloc
 from starkware.starknet.common.syscalls import get_caller_address, replace_class, library_call
 
+@contract_interface
+namespace IKakarot {
+    func get_account_contract_class_hash() -> (account_contract_class_hash: felt) {
+    }
+}
+
 // We are intentionally causing a storage_slot collision here,
 // by defining these variables in both `uninitialized_account` and `account_contract`.
 // We are defining them here instead of in the account library, so as to not depend
@@ -36,13 +42,12 @@ func constructor{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr
 @external
 func initialize{
     syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr, bitwise_ptr: BitwiseBuiltin*
-}(implementation_class: felt) {
+}() {
     let (caller) = get_caller_address();
     let (kakarot_address) = Account_kakarot_address.read();
-    with_attr error_message("Only Kakarot") {
-        assert kakarot_address = caller;
-    }
+    let (implementation_class) = IKakarot.get_account_contract_class_hash(kakarot_address);
     replace_class(implementation_class);
+
     let (calldata) = alloc();
     assert [calldata] = implementation_class;
     library_call(
--- a/src/kakarot/accounts/uninitialized_account.cairo
+++ b/src/kakarot/accounts/uninitialized_account.cairo
@@ -2,9 +2,8 @@
 
 %lang starknet
 
-from starkware.cairo.common.cairo_builtins import HashBuiltin, BitwiseBuiltin
-from starkware.cairo.common.alloc import alloc
-from starkware.starknet.common.syscalls import get_caller_address, replace_class, library_call
+from starkware.cairo.common.cairo_builtins import HashBuiltin
+from starkware.starknet.common.syscalls import replace_class
 
 @contract_interface
 namespace IKakarot {
--- a/src/kakarot/accounts/uninitialized_account.cairo
+++ b/src/kakarot/accounts/uninitialized_account.cairo
@@ -12,12 +12,16 @@ namespace IKakarot {
     }
 }
 
-const INITIALIZE_SELECTOR = 0x79dc0da7c54b95f10aa182ad0a46400db63156920adb65eca2654c0945a463;  // sn_keccak('initialize')
+@contract_interface
+namespace IAccount {
+    func initialize(kakarot_address: felt, evm_address: felt, implementation_class: felt) {
+    }
+}
 
 // @title Uninitialized Contract Account. Used to get a deterministic address for an account, no
 // matter the actual implementation class used.
 
-// @notice Deploys and initializes the account with the Kakarot and EVM addresses it was deployed with.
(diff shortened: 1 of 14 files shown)


tests/end_to_end/test_kakarot.py
```diff
@@ -153,20 +153,80 @@ async def test_should_return_same_as_deployed_address(
...
+            fund_starknet_address,
+            register_account,
+            compute_starknet_address,
+            get_contract,
+        ):
+            evm_address = generate_random_evm_address(random.randint(0, 0x5EED))
+            starknet_address = await compute_starknet_address(evm_address)
+            await fund_starknet_address(starknet_address, PRE_FUND_AMOUNT / 1e18)
+
+            # TODO: change the sender's address to be the account itself
```
enitrat, line 183: @ClementWalter not sure how I can get this to pass easily. do you think the unit test is sufficient?

tests/end_to_end/test_kakarot.py
```diff
@@ -153,20 +153,80 @@ async def test_should_return_same_as_deployed_address(
...
         async def test_should_deploy_starknet_contract_at_corresponding_address(
             self,
+            starknet: FullNodeClient,
             fund_starknet_address,
             deploy_externally_owned_account,
             compute_starknet_address,
             get_contract,
         ):
-            evm_address = generate_random_evm_address()
+            evm_address = generate_random_evm_address(random.randint(0, 0x5EED))
```
enitrat, line 162: otherwise the tx would fail because of a deploy_syscall address collision but the test would pass as the contract was already deployed

enitrat:
@ClementWalter once I have your greenlight I adapt the EF-Test runner accordingly, and then we can merge everything

src/kakarot/accounts/uninitialized_account.cairo
```diff
@@ -6,50 +6,38 @@ from starkware.cairo.common.cairo_builtins import HashBuiltin, BitwiseBuiltin
...
+    func get_account_contract_class_hash() -> (account_contract_class_hash: felt) {
+    }
 }
 
 const INITIALIZE_SELECTOR = 0x79dc0da7c54b95f10aa182ad0a46400db63156920adb65eca2654c0945a463;  // sn_keccak('initialize')
 
 // @title Uninitialized Contract Account. Used to get a deterministic address for an account, no
 // matter the actual implementation class used.
+
+// @notice Deploys and initializes the account with the Kakarot and EVM addresses it was deployed with.
```
ClementWalter, line 20: Apply the same rules as for python files. Rules are defined by the ruff linter, see

https://docs.astral.sh/ruff/rules/#pydocstyle-d

and more precisely here

https://docs.astral.sh/ruff/rules/non-imperative-mood/

> First line of docstring should be in imperative mood: "{first_line}"

src/kakarot/accounts/uninitialized_account.cairo
```diff
@@ -6,50 +6,38 @@ from starkware.cairo.common.cairo_builtins import HashBuiltin, BitwiseBuiltin
...
-// We are defining them here instead of in the account library, so as to not depend
-// on content of the account library in uninitialized_account and ensure a fixed class hash.
-@storage_var
-func Account_evm_address() -> (evm_address: felt) {
-}
-
-@storage_var
-func Account_kakarot_address() -> (kakarot_address: felt) {
+@contract_interface
+namespace IKakarot {
```
ClementWalter, line 10: we could as well define a `namespace IAccount` with the initialize, the use it instead of the raw library call

src/kakarot/library.cairo
```diff
@@ -218,10 +220,46 @@ namespace Kakarot {
...
+    // @notice Return the class hash of the account implementation
+    // @return account_contract_class_hash The class hash of the account implementation
+    func get_account_contract_class_hash{
+        syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr
+    }() -> (account_contract_class_hash: felt) {
+        let (account_contract_class_hash) = Kakarot_account_contract_class_hash.read();
+        return (account_contract_class_hash,);
+    }
+
+    // @notice Registers an account by writing its EVM address to the Starknet address mapping.
```
ClementWalter, line 235: see comment on https://docs.astral.sh/ruff/rules/non-imperative-mood/

src/kakarot/library.cairo
```diff
@@ -218,10 +220,46 @@ namespace Kakarot {
...
+
+        let (existing_address) = Kakarot_evm_to_starknet_address.read(evm_address);
+        with_attr error_message("Kakarot: account already registered") {
+            assert existing_address = 0;
+        }
+
+        let (local caller_address: felt) = get_caller_address();
+        let starknet_address = Account.compute_starknet_address(evm_address);
+        local starknet_address = starknet_address;
+        let (local kakarot_address: felt) = get_contract_address();
```
ClementWalter, line 251: ?

src/kakarot/library.cairo
ClementWalter: I think that this function is more

```
@notice Register the calling Starknet address for the given EVM address
@dev    Only the corresponding computed Starknet address can make this call to ensure that registered accounts are actually deployed.
```

src/kakarot/library.cairo
```diff
@@ -218,10 +220,46 @@ namespace Kakarot {
...
+
+        with_attr error_message(
+                "Kakarot: register account address mismatch {starknet_address} != {caller_address} != {kakarot_address} ") {
```
ClementWalter, line 254: maybe an error like "caller should be corresponding starknet address" ?

tests/end_to_end/conftest.py
```diff
@@ -187,6 +187,26 @@ async def _factory(evm_address: Union[int, str]):
...
+    """
+
+    async def _factory(evm_address: Union[int, str]):
+        if isinstance(evm_address, str):
+            evm_address = int(evm_address, 16)
+        tx = await kakarot.functions["register_account"].invoke_v1(
+            evm_address, max_fee=max_fee
+        )
+        await wait_for_transaction(
+            tx.hash,
```
ClementWalter, line 203: remove `,` to make black format in one line

tests/end_to_end/test_kakarot.py
ClementWalter: can't you just do `generate_random_evm_address(0x5EED)`

ie : `generate_random_evm_address` only arg is actually a seed, so why do you need to do a random with this seed to feed the seed?

tests/end_to_end/test_kakarot.py
ClementWalter: not sure how to do this because the passing case is a starknet account calling the function to register its corresponding EVM account, but the only way to have this mapping correct is to start from the EVM address, and deploy the corresponding starknet account, and registration is done during deployment...

consequently I think that in terms of end-to-end test, this is covered elsewhere, and unit test is actually enough

tests/end_to_end/test_kakarot.py
```diff
@@ -153,20 +153,74 @@ async def test_should_return_same_as_deployed_address(
...
+
+            # TODO: change the sender's address to be the account itself
+            # kakarot = get_contract("kakarot", address=kakarot.address, provider=starknet_address)
+
+            # await deploy_externally_owned_account(evm_address)
+            # tx = await register_account(evm_address)
+            # receipt = await starknet.get_transaction_receipt(tx.hash)
+            # assert receipt.execution_status == "REVERTED"
+
+        async def test_should_fail_register_account_sender_not_account(
```
ClementWalter, line 191: `test_should_fail_when_sender_is_not_account` (no need to repeat the SUT name)

tests/end_to_end/test_kakarot.py
```diff
@@ -153,20 +153,74 @@ async def test_should_return_same_as_deployed_address(
...
+        ):
+            evm_address = generate_random_evm_address(random.randint(0, 0x5EED))
+            await compute_starknet_address(evm_address)
+
+            tx = await register_account(evm_address)
+            receipt = await starknet.get_transaction_receipt(tx.hash)
+            assert receipt.execution_status.name == "REVERTED"
+            assert "Kakarot: register account address mismatch" in receipt.revert_reason
+
+        async def test_should_fail_register_account_already_deployed(
```
ClementWalter, line 208: idem `test_should_fail_when_account_is_already_registered`

tests/src/kakarot/accounts/test_contract_account.py
```diff
@@ -20,52 +20,74 @@ def bytecode(self, request):
...
             )
             SyscallHandler.mock_storage.assert_any_call(
-                address=get_storage_var_address("Account_evm_address"), value=0xABDE1
+                address=get_storage_var_address("Account_is_initialized"), value=1
             )
 
+        @SyscallHandler.patch("IKakarot.register_account", lambda addr, data: [])
+        @SyscallHandler.patch("IKakarot.get_native_token", lambda addr, data: [0xDEAD])
+        @SyscallHandler.patch("IERC20.approve", lambda addr, data: [1])
+        def test_should_store_implementation(self, cairo_run):
```
ClementWalter, line 47: why not one single test `test_should_set_storage_variables`?

ClementWalter (commented):
looks good overall

src/kakarot/library.cairo
enitrat: sorry that's from a debugging session where I could only use the error message to have values...

src/kakarot/library.cairo
```diff
@@ -218,10 +220,46 @@ namespace Kakarot {
...
+        alloc_locals;
+
+        let (existing_address) = Kakarot_evm_to_starknet_address.read(evm_address);
+        with_attr error_message("Kakarot: account already registered") {
+            assert existing_address = 0;
+        }
+
+        let (local caller_address: felt) = get_caller_address();
+        let starknet_address = Account.compute_starknet_address(evm_address);
+        local starknet_address = starknet_address;
```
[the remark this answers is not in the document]
enitrat: error messages cannot be ap-based (!)

src/kakarot/library.cairo
enitrat: oops, forgot to remove this

src/kakarot/library.cairo
enitrat: `                "Kakarot: Caller should be {starknet_address}, got {caller_address}") {`

tests/src/kakarot/accounts/test_contract_account.py
enitrat: that's what the existing test design was like. I merged them.

src/kakarot/accounts/uninitialized_account.cairo
ClementWalter: but kakarot is de facto a python project and we enforce this linter for python files
it's opinionated but we need to be consistent with half of the codebase

src/kakarot/accounts/uninitialized_account.cairo
```diff
@@ -6,50 +6,37 @@ from starkware.cairo.common.cairo_builtins import HashBuiltin, BitwiseBuiltin
 from starkware.cairo.common.alloc import alloc
 from starkware.starknet.common.syscalls import get_caller_address, replace_class, library_call
```
ClementWalter, line 7: remove unused

src/kakarot/library.cairo
```diff
@@ -5,7 +5,7 @@
 from openzeppelin.access.ownable.library import Ownable
 from starkware.cairo.common.bool import FALSE, TRUE
 from starkware.cairo.common.cairo_builtins import HashBuiltin, BitwiseBuiltin
-from starkware.starknet.common.syscalls import get_caller_address, get_tx_info
+from starkware.starknet.common.syscalls import get_caller_address, get_tx_info, get_contract_address
```
ClementWalter, line 8: remove unused

ClementWalter (approved):
Lgtm

PR #1088, opened 2024-04-09
feat: seamless account upgrades

### LINKED ISSUE #1087: feat: enable seamless account upgrades
Accounts are Starknet contracts under the hood. With account abstraction, we can inject any logic we want in the account entrypoints. 

As such, we can guarantee seamless upgrades of accounts by doing the following:
- Query the latest account class from the Kakarot contract in __execute__
- If the account is not up to date, library_call the latest account class to execute the transaction
- replace the account's class

enitrat:
Time spent on this PR: 0.7d 

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Resolves #1087

## What is the new behavior?

- Enables seamless upgrades of accounts, hidden behind the __execute__ entrypoint.
-
-

- - -
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot/1088)

### DIFF
--- a/tests/end_to_end/UniswapV2/conftest.py
+++ b/tests/end_to_end/UniswapV2/conftest.py
@@ -3,7 +3,7 @@ import pytest_asyncio
 TOTAL_SUPPLY = 10000 * 10**18
 
 
-@pytest_asyncio.fixture(scope="module")
+@pytest_asyncio.fixture(scope="function")
 async def token_a(deploy_solidity_contract, owner):
     return await deploy_solidity_contract(
         "UniswapV2",
@@ -13,7 +13,7 @@ async def token_a(deploy_solidity_contract, owner):
     )
 
 
-@pytest_asyncio.fixture(scope="module")
+@pytest_asyncio.fixture(scope="function")
 async def token_b(deploy_solidity_contract, owner):
     return await deploy_solidity_contract(
         "UniswapV2",


src/kakarot/library.cairo
```diff
@@ -225,6 +225,16 @@ namespace Kakarot {
         return (starknet_contract_address=starknet_contract_address);
     }
 
+    // @notice Set the account implementation class hash
+    // @param account_contract_class_hash The new account implementation class hash
+    func set_account_contract_class_hash{
+        syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr
+    }(account_contract_class_hash: felt) {
+        Ownable.assert_only_owner();
```
ClementWalter, line 233: I know that we have this everywhere, but the `assert_only_owner` and other similar restriction should be put in the contract and not in the library.
I've created an issue to fix them all https://github.com/kkrt-labs/kakarot/issues/1094

tests/end_to_end/test_account.py
```diff
@@ -0,0 +1,69 @@
...
+async def class_hashes():
+    """
+    All declared class hashes.
+    """
+    from kakarot_scripts.utils.starknet import get_declarations
+
+    return get_declarations()
+
+
+@pytest_asyncio.fixture(scope="function")
```
ClementWalter, line 20: `scope='function'` so you want to deploy a new token each time?

tests/end_to_end/test_account.py
```diff
@@ -0,0 +1,69 @@
...
+@pytest.mark.asyncio(scope="session")
+@pytest.mark.AccountContract
+class TestAccount:
+    class TestAutoUpgradeOnTransaction:
+        async def test_should_upgrade_outdated_account_on_transfer(
+            self, invoke, token_a, owner, other, class_hashes
+        ):
+            # Given an initial version of 1000
+            prev_version = (await call(owner.starknet_contract.address, "version"))[0]
+            assert prev_version == 1000  # 000.001.000
```
ClementWalter, line 42: I think that you don't need to test against hard coded values

as there is no way to make an external library call, let's just use class hashes?

1. get current class hash
2. assert != class_hashes["account_contract_fixture"]
3. update kakarot
4. make token tx
5. get new class hash
6. assert new class hash == class_hashes["account_contract_fixture"]

?

tests/end_to_end/test_account.py
enitrat: no, once is enough

tests/end_to_end/test_account.py
```diff
@@ -0,0 +1,69 @@
...
+    """
+    from kakarot_scripts.utils.starknet import get_declarations
+
+    return get_declarations()
+
+
+@pytest_asyncio.fixture(scope="function")
+async def token_a(
+    deploy_solidity_contract,
+    owner,
```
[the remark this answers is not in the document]
enitrat: ok

src/kakarot/library.cairo
enitrat: done

tests/end_to_end/test_account.py
enitrat: ok, but i didn't understand what you mean by

> as there is no way to make an external library call, let's just use class hashes?

ClementWalter: I meant that it not possible to get the version of the new and prev class hash with a RPC call

docs/general/accounts.md
```diff
@@ -136,14 +136,7 @@ of accounts. The upgrade process works as follows:
...
-
-The account contract class exposes an `upgrade` entrypoint, which allows the
-owner of the account to upgrade the account contract class to a new version.
-However, this process is related to the _Starknet_ account, which is abstracted
-from the user by the system, while users can interact with Kakarot using an EVM
-EOA. To upgrade an account, the user will need to perform the upgrade action
-directly, as the caller of the upgrade function **must** be the account itself.
+## Upgrading accounts with AA
 
 One proposal for seamless upgrades would be to check if an account's class hash
```
ClementWalter, line 141: so it's not a proposal anymore, but the chosen design

solidity_contracts/src/UniswapV2/conftest.py
```diff
@@ -25,7 +25,7 @@ async def uniswap_pair_deployer(addresses):
     return addresses[14]
 
 
-@pytest_asyncio.fixture(scope="module")
+@pytest_asyncio.fixture(scope="function")
```
ClementWalter, line 28: why ? this makes deploying a new token each time you request the fixture

solidity_contracts/src/UniswapV2/conftest.py
```diff
@@ -38,7 +38,7 @@ async def token_a(
     )
 
 
-@pytest_asyncio.fixture(scope="module")
+@pytest_asyncio.fixture(scope="function")
```
ClementWalter, line 41: same

solidity_contracts/src/UniswapV2/conftest.py
ClementWalter, line 1: delete this file, it's a mistake, I didn't even know we had it

tests/end_to_end/test_account.py
```diff
@@ -0,0 +1,77 @@
...
+            erc20_token,
+            owner,
+            other,
+            class_hashes,
+        ):
+            # Given an initial class
+            prev_class = await starknet.get_class_hash_at(
+                owner.starknet_contract.address
+            )
+            assert prev_class == class_hashes["account_contract"]
```
ClementWalter, line 46: what if `class_hashes["account_contract"]` and `class_hashes["account_contract_fixture"]` are the same? reading the test there is no guarantee that the class is well change.

That's why you need to do in this order

- `assert prev_class != target_class`
- kakarot.upgrade(target_class)
- `assert new_class == target_class`

so that it's sure that the class has been changed to match the target one stored in kakarot

tests/fixtures/account_contract_fixture.cairo
ClementWalter, line 1: maybe you can simplify this file, and just import the real account contract, and add a "salt" in it to make sure the class hash is not the same. something like

```
from kakarot.accounts.account_contract import constructor

// make sure the class hash is different
const SALT = 'salt'
```

tests/end_to_end/conftest.py
```diff
@@ -108,6 +108,26 @@ def others(addresses):
     return addresses[2:]
 
 
+@pytest_asyncio.fixture(scope="function")
```
enitrat, line 111: made this function scope as I want to have a new account at disposition for each call

tests/end_to_end/test_account.py
enitrat: actually - yes, as I want the deployer to be the new_account whose scope is "function" now.

ClementWalter (approved):

PR #1097, opened 2024-04-11
add more rlp cases for long bytes for profiling

Eikix:
Time spent on this PR:0.01day

- added more cases to the rlp decode test, to help profile edge cases of large contracts (~24kb and even more) -> can we limit the cairo_run function to a certain numbers of steps? that way it'll fail the test if it takes too many steps? to replicate appchain behaviour?
- maybe good idea to also add a test in end-to-end tests to replicate more accurately appchain behaviour -> try to deploy an enormous chunk of bytecode

- - -
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot/1097)

### DIFF
--- a/tests/src/utils/test_rlp.py
+++ b/tests/src/utils/test_rlp.py
@@ -58,7 +58,9 @@ class TestRLP:
             assert output[2] == expected_len
 
     class TestDecode:
-        @pytest.mark.parametrize("payload_len", [55, 56])
+        @pytest.mark.parametrize(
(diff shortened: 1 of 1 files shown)


ClementWalter:
@Eikix is this a draft? ready for review? do you have profiling data regarding this new cases?

Eikix:
> @Eikix is this a draft? ready for review? do you have profiling data regarding this new cases?

I haven't yet run the profiling data! this was a PR trying to increase diversity of RLP decoding in prevision of https://github.com/kkrt-labs/kakarot/issues/1102

left it in draft mode so that we could first run profiling together before deciding what to do

Issue #1095, opened 2024-04-10
feat: optimize `verify_eth_signature`


enitrat:
### Describe the Feature Request

`verify_eth_signature` takes 89% of the steps of a `validate`. Investigate whether using the new builtin ec_op, through a Cairo1 helper class, would reduce this amount of steps.
![profile005](https://github.com/kkrt-labs/kakarot/assets/60658558/10b3a17e-8691-4905-93bc-67b1ad16d321)

enitrat:
Result:

Using Cairo1's `verify_eth_signature` for an empty `eth_send_transaction`: 
```bash
ExecutionResources(steps=252470, range_check_builtin_applications=16603, pedersen_builtin_applications=180, poseidon_builtin_applications=None, ec_op_builtin_applications=None, ecdsa_builtin_applications=None, bitwise_builtin_applications=3176, keccak_builtin_applications=2, memory_holes=None, segment_arena_builtin=None)
```

Using Cairo0's

```bash
ExecutionResources(steps=283130, range_check_builtin_applications=18366, pedersen_builtin_applications=180, poseidon_builtin_applications=None, ec_op_builtin_applications=None, ecdsa_builtin_applications=None, bitwise_builtin_applications=3180, keccak_builtin_applications=1, memory_holes=None, segment_arena_builtin=None)
```
overall, there are 30k steps less using Cairo1, likely due to a `secp256_ec_get_point_from_x_syscall`

enitrat:
Idea on how to proceed and make this compatible in the future:

- Use `Kakarot` as a source of truth regarding the latest available Cairo1Helpers contract
- When deploying an account, write to storage the value of `Kakarot_cairo1_helpers_class_hash` by reading it from Kakarot.
- When executing `__execute__` and if an account upgrade is available, before the upgrade process, start by reading the latest class_hash of `Cairo1Helpers` and writing it to the account's storage var. If we add new syscalls or helpers, they will be accessible by accounts directly
- Delegate all `verify_eth_signature` and `keccak` call to the library

This should save ~35k steps per transaction, which is a consequent optimization.

PR #1105, opened 2024-04-16
feat: account cairo1 helpers class upgrade

enitrat:
Time spent on this PR: 0.5d

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

First step for #1095 

## What is the new behavior?

-
- Adds a storage var to accounts to store the Cairo1Helpers class. Update upgradeability mechanism to pull the latest version when required.
-

- - -
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot/1105)

### DIFF
--- a/src/kakarot/accounts/account_contract.cairo
+++ b/src/kakarot/accounts/account_contract.cairo
@@ -8,7 +8,11 @@ from starkware.cairo.common.cairo_builtins import HashBuiltin, BitwiseBuiltin, S
 from starkware.cairo.common.uint256 import Uint256
 
 // Local dependencies
-from kakarot.accounts.library import AccountContract, Account_implementation
+from kakarot.accounts.library import (
+    AccountContract,
+    Account_implementation,
+    Account_cairo1_helpers_class_hash,
+)
 from kakarot.accounts.model import CallArray
 from kakarot.interfaces.interfaces import IKakarot, IAccount
 from starkware.starknet.common.syscalls import get_tx_info, get_caller_address, replace_class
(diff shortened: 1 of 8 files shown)


src/kakarot/accounts/account_contract.cairo
```diff
@@ -136,8 +140,27 @@ func __execute__{
     let latest_class = AccountContract.get_latest_class();
```
ClementWalter, line 140: update the `get_latest_class` to `get_latest_classes` and return latest account class and latest helper class.

src/kakarot/accounts/account_contract.cairo
```diff
@@ -136,8 +140,27 @@ func __execute__{
...
+            tempvar range_check_ptr = range_check_ptr;
+            tempvar pedersen_ptr = pedersen_ptr;
+        } else {
+            tempvar syscall_ptr = syscall_ptr;
+            tempvar range_check_ptr = range_check_ptr;
+            tempvar pedersen_ptr = pedersen_ptr;
+        }
+        let pedersen_ptr = cast([ap - 1], HashBuiltin*);
+        let range_check_ptr = [ap - 2];
+        let syscall_ptr = cast([ap - 3], felt*);
```
ClementWalter, line 161: write in the order of the branches, ie `ap-3`, `ap-2`, `ap-1`

tests/end_to_end/test_account.py
```diff
@@ -38,50 +72,57 @@ async def test_should_upgrade_outdated_account_on_transfer(
             new_account,
             other,
             class_hashes,
+            cleanup,
```
ClementWalter, line 75: use `@pytest.mark.usefixture("cleanup")` instead of requesting it as an (unused) argument.
you can also define the fixture with `@pytest.fixture(autouse=True)` if you want it to be used for all the tests

tests/end_to_end/test_account.py
```diff
@@ -26,6 +27,39 @@ async def erc20_token(deploy_solidity_contract, new_account):
...
+        "kakarot",
+        "set_account_contract_class_hash",
+        class_hashes["account_contract"],
+    )
+    await invoke(
+        "kakarot", "set_cairo1_helpers_class_hash", class_hashes["Cairo1Helpers"]
+    )
+
+
+async def assert_transfer_success(erc20_token, new_account, other):
```
ClementWalter, line 43: maybe use the `Counter.inc()` instead as this is not really minimal to illustrate the fact that the original tx is well processed even though we actually upgrade the classe

enitrat:
#1107

src/kakarot/accounts/account_contract.cairo
```diff
@@ -133,19 +137,37 @@ func __execute__{
...
-    if (latest_class != this_class) {
-        // Must be done before library_call, otherwise entering an infinite recursive loop.
-        Account_implementation.write(latest_class);
+    if (latest_account_class != this_class) {
+        // Update storage
+        Account_implementation.write(latest_account_class);
+        let (kakarot_address) = Ownable_owner.read();
+        let (this_cairo1helpers_class) = Account_cairo1_helpers_class_hash.read();
+
+        if (latest_cairo1helpers_class != this_cairo1helpers_class) {
```
ClementWalter, line 148: what if we change the helper but not the account class? I think that they need to be in separate branches, and that overall the function can be simplified like

```
let (account_class, helpers_class) = AccountContract.get_latest_classes();
let (this_helpers_class) = Account_helpers_class_hash.read();

if (helpers_class != this_cairo1helpers_class) {
    // Note this could be done in any cases but would cost a useless write
    Account_cairo1_helpers_class_hash.write(latest_cairo1helpers_class);
}

// Do the call with the latest account class, possibly same class, works in all the cases
let (response_len, response) = IAccount.library_call___execute__(
    class_hash=latest_account_class,
    call_array_len=call_array_len,
    call_array=call_array,
    calldata_len=calldata_len,
    calldata=calldata,
);
```

we don't need to actually store the implementation if we always call kakarot to actually check what is the implementation.

PR #1109, opened 2024-04-17
add vyper contracts

### LINKED ISSUE #1108: feat: add vyper contracts
Add Vyper contracts and support for Vyper

swetshaw:
Time spent on this PR: 1 hour

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Other (please describe): Added Vyper contracts

Resolves #1108

- - -
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot/1109)


ClementWalter (changes requested):
I think that you miss a `poetry add vyper` somewhere, and also don't all contracts we don't use, only the one we need (`3074InvokePoc` I guess) and fix case in naming

possibly also add a readme or command in the makefile to actually compile these files

vyper_contracts/src/3074invokePoc.vy
```diff
@@ -0,0 +1,23 @@
...
+
+from ethereum.ercs import IERC20
+
+# counter to keep track of the number of invokes
+counter: public(uint256)
+
+# authorise this contract to spend the tokens on behalf of the user
+@external
+def autherizeInvoke(sig: Bytes[97]):
+    authorize(self, sig)
```
ClementWalter, line 11: are you sure the `authorize` can be in another function than the `authcall`?

vyper_contracts/src/3074invokePoc.vy
```diff
@@ -0,0 +1,23 @@
...
+@external
+def invoke(token: IERC20, receiver: address, amount: uint256):
+    self.counter += 1
+    authcall token.approve(receiver, amount)
+    authcall token.transfer(receiver, amount)
+
+# add a function that returns a value
+@external
+def getCounter() -> uint256 :
+    return self.counter
```
ClementWalter, line 23: add new line at the end of files (if using VSCode, use the setting `"files.insertFinalNewline": true,`)

Makefile
```diff
@@ -64,8 +64,15 @@ build-sol:
...
+install-vyper:
+	pip3 install git+https://github.com/charles-cooper/vyper@feat/eip-3074
+	vyper --version
+
 install-katana:
 	cargo install --git https://github.com/dojoengine/dojo --locked --tag v0.6.1-alpha.1 katana
 
 run-katana:
-	katana --chain-id test --validate-max-steps 16777216 --invoke-max-steps 16777216 --eth-gas-price 0 --strk-gas-price 0 --disable-fee
+	katana --chain-id test --validate-max-steps 16777216 --invoke-max-steps 16777216 --gas-price 0 --disable-fee
```
ClementWalter, line 78: why do you change this?

Makefile
```diff
@@ -64,8 +64,15 @@ build-sol:
 	git submodule update --init --recursive
 	forge build --names --force
 
+build-vyper:
+	poetry run python vyper_contracts/scripts/compile.py
+
+install-vyper:
```
ClementWalter, line 70: you don't need this as poetry install everything

Makefile
```diff
@@ -64,8 +64,15 @@ build-sol:
 	git submodule update --init --recursive
 	forge build --names --force
 
+build-vyper:
+	poetry run python vyper_contracts/scripts/compile.py
```
ClementWalter, line 68: why not just `poetry run vyper -f abi,bytecode,bytecode_runtime yourFileName.vy` (maybe check the flags)

Makefile
swetshaw: It was not recognising --eth-gas-price and --strk-gas-price, suggested to change to --gas-price. May be I was using older version of katana. Let me upgrade my katana and try again

Makefile
swetshaw: It does work but it prints the result in the cli instead of generating build files. That is why added the script in  `compile.py` to put the result in build folder so that can be used later while deploying.

Not sure if there are other ways to do it

Makefile
swetshaw: It worked after installing katana from the makefile. Will revert the changes

ClementWalter: usually if it's not related at all with your PR, then you should not update it

ClementWalter (dismissed):

vyper_contracts/src/EIP3074/AuthorizedTransfer.vy
```diff
@@ -0,0 +1,14 @@
...
+from ethereum.ercs import IERC20
+
+# counter to keep track of the number of invokes
+counter: public(uint256)
+
+# authorise this contract to spend the tokens on behalf of the user
+@external
+def authorizeInvoke(sig: Bytes[97], token: IERC20, receiver: address, amount: uint256):
+    self.counter += 1
+    authorize(self, sig)
```
enitrat, line 12: I don't think authorize should take self as an argument? The address should be the one of the authority

enitrat: perhaps add a `sender` address function parameter that we will auth with

ClementWalter (approved):

PR #1130, opened 2024-04-29
feat: store jumpdests

enitrat:
Time spent on this PR:

Benchmark: using `test_should_add_liquidity` (a test that switches back-and-fourth between multiple contracts multiple times and thus should benefit from the caching mechanism)

**Before**
```
TRACE executor: Transaction resource usage. usage="Steps: 8093994 | Bitwise: 1420 | Keccak: 115 | L1 Gas: 364303 | Pedersen: 621 | Range Checks: 391939"
```

**After**
```
Transaction resource usage. usage="Steps: 6344085 | Bitwise: 1420 | Keccak: 115 | L1 Gas: 653027 | Pedersen: 1503 | Range Checks: 308089"
```

=> 1.7M steps saved
## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Resolves #<Issue number>

## What is the new behavior?

-
-
-

- - -
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot/1130)

### DIFF
--- a/src/backend/starknet.cairo
+++ b/src/backend/starknet.cairo
@@ -314,35 +314,31 @@ namespace Internals {
 
         let (local keys_start: felt*) = alloc();
 
-        tempvar keys_len = 0;
         tempvar keys = keys_start;
         tempvar dict = dict_start;
         tempvar remaining = dict_len;
 
         loop:
-        let keys_len = [ap - 4];
         let keys = cast([ap - 3], felt*);
         let dict = cast([ap - 2], DictAccess*);
         let is_valid = dict.new_value;
 
         if (is_valid != 0) {
             assert [keys] = dict.key;
-            tempvar keys_len = keys_len + 1;
             tempvar keys = keys + 1;
             tempvar dict = dict + DictAccess.SIZE;
         } else {
-            tempvar keys_len = keys_len;
             tempvar keys = keys;
             tempvar dict = dict + DictAccess.SIZE;
         }
         tempvar remaining = dict_end - dict;
 
-        static_assert keys_len == [ap - 4];
         static_assert keys == [ap - 3];
         static_assert dict == [ap - 2];
 
         jmp loop if remaining != 0;
 
+        let keys_len = keys - keys_start;
         IAccount.write_jumpdests(starknet_address, jumpdests_len=keys_len, jumpdests=keys_start);
         return ();
     }
--- a/src/backend/starknet.cairo
+++ b/src/backend/starknet.cairo
@@ -222,10 +222,13 @@ namespace Internals {
         // Save storages
         Internals._save_storage(starknet_address, self.storage_start, self.storage);
 
-        // Update bytecode if required (SELFDESTRUCTed contract, redeployed)
-        let (bytecode_len) = IAccount.bytecode_len(starknet_address);
-        if (bytecode_len != self.code_len) {
+        // Update bytecode and jumpdests if required (SELFDESTRUCTed contract, redeployed)
+        if (self.created != FALSE) {
             IAccount.write_bytecode(starknet_address, self.code_len, self.code);
+            let (valid_indexes) = alloc();
+            Internals._save_valid_jumpdests(
+                starknet_address, self.valid_jumpdests_start, self.valid_jumpdests, 0, valid_indexes
+            );
             return ();
         }
 
@@ -294,5 +297,44 @@ namespace Internals {
         );
 
         return _save_storage(starknet_address, storage_start + DictAccess.SIZE, storage_end);
+    }
+
+    func _save_valid_jumpdests{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr}(
+        starknet_address: felt,
+        jumpdests: DictAccess*,
+        jumpdests_end: DictAccess*,
+        valid_indexes_len: felt,
+        valid_indexes: felt*,
+    ) {
+        if (jumpdests == jumpdests_end) {
+            IAccount.write_jumpdests(
+                contract_address=starknet_address,
+                jumpdests_len=valid_indexes_len,
+                jumpdests=valid_indexes,
+            );
+            return ();
+        }
+
+        let is_valid = jumpdests.new_value;
+
+        // If the value is 0, it means the jumpdest is invalid and should not be saved.
+        if (is_valid == FALSE) {
+            return _save_valid_jumpdests(
+                starknet_address,
+                jumpdests + DictAccess.SIZE,
+                jumpdests_end,
+                valid_indexes_len,
+                valid_indexes,
+            );
+        }
+
+        assert valid_indexes[valid_indexes_len] = jumpdests.key;
+        return _save_valid_jumpdests(
+            starknet_address,
+            jumpdests + DictAccess.SIZE,
+            jumpdests_end,
+            valid_indexes_len + 1,
+            valid_indexes,
+        );
     }
 }
--- a/src/backend/starknet.cairo
+++ b/src/backend/starknet.cairo
@@ -307,46 +307,41 @@ namespace Internals {
         starknet_address: felt, dict_start: DictAccess*, dict_end: DictAccess*
     ) {
         alloc_locals;
-        let (local keys_start: felt*) = alloc();
         let dict_len = dict_end - dict_start;
-        let (local len, _) = unsigned_div_rem(dict_len, DictAccess.SIZE);
-        local range_check_ptr = range_check_ptr;
-
         if (dict_len == 0) {
             return ();
         }
 
+        let (local keys_start: felt*) = alloc();
+
         tempvar keys_len = 0;
         tempvar keys = keys_start;
-        tempvar len = len;
         tempvar dict = dict_start;
+        tempvar remaining = dict_len;
 
         loop:
         let keys_len = [ap - 4];
         let keys = cast([ap - 3], felt*);
-        let len = [ap - 2];
-        let dict = cast([ap - 1], DictAccess*);
+        let dict = cast([ap - 2], DictAccess*);
         let is_valid = dict.new_value;
 
         if (is_valid != 0) {
             assert [keys] = dict.key;
             tempvar keys_len = keys_len + 1;
             tempvar keys = keys + 1;
-            tempvar len = len - 1;
             tempvar dict = dict + DictAccess.SIZE;
         } else {
             tempvar keys_len = keys_len;
             tempvar keys = keys;
-            tempvar len = len - 1;
             tempvar dict = dict + DictAccess.SIZE;
         }
+        tempvar remaining = dict_end - dict;
 
         static_assert keys_len == [ap - 4];
         static_assert keys == [ap - 3];
-        static_assert len == [ap - 2];
-        static_assert dict == [ap - 1];
+        static_assert dict == [ap - 2];
 
-        jmp loop if len != 0;
+        jmp loop if remaining != 0;
 
         IAccount.write_jumpdests(starknet_address, jumpdests_len=keys_len, jumpdests=keys_start);
         return ();
--- a/src/backend/starknet.cairo
+++ b/src/backend/starknet.cairo
@@ -210,9 +210,8 @@ namespace Internals {
             _save_storage(starknet_address, self.storage_start, self.storage);
 
             // Save valid jumpdests
-            let (valid_indexes) = alloc();
             Internals._save_valid_jumpdests(
-                starknet_address, self.valid_jumpdests_start, self.valid_jumpdests, 0, valid_indexes
+                starknet_address, self.valid_jumpdests_start, self.valid_jumpdests
             );
             return ();
         }
@@ -231,9 +230,8 @@ namespace Internals {
         // Update bytecode and jumpdests if required (SELFDESTRUCTed contract, redeployed)
         if (self.created != FALSE) {
             IAccount.write_bytecode(starknet_address, self.code_len, self.code);
-            let (valid_indexes) = alloc();
             Internals._save_valid_jumpdests(
-                starknet_address, self.valid_jumpdests_start, self.valid_jumpdests, 0, valid_indexes
+                starknet_address, self.valid_jumpdests_start, self.valid_jumpdests
             );
             return ();
         }
@@ -306,40 +304,51 @@ namespace Internals {
     }
 
     func _save_valid_jumpdests{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr}(
-        starknet_address: felt,
-        jumpdests: DictAccess*,
-        jumpdests_end: DictAccess*,
-        valid_indexes_len: felt,
-        valid_indexes: felt*,
+        starknet_address: felt, dict_start: DictAccess*, dict_end: DictAccess*
     ) {
         alloc_locals;
-        if (jumpdests == jumpdests_end) {
-            IAccount.write_jumpdests(
-                starknet_address, jumpdests_len=valid_indexes_len, jumpdests=valid_indexes
-            );
+        let (local keys_start: felt*) = alloc();
+        let dict_len = dict_end - dict_start;
+        let (local len, _) = unsigned_div_rem(dict_len, DictAccess.SIZE);
+        local range_check_ptr = range_check_ptr;
+
+        if (dict_len == 0) {
             return ();
         }
 
-        let is_valid = jumpdests.new_value;
+        tempvar keys_len = 0;
+        tempvar keys = keys_start;
+        tempvar len = len;
+        tempvar dict = dict_start;
 
-        // If the value is 0, it means the jumpdest is invalid and should not be saved.
-        if (is_valid == FALSE) {
-            return _save_valid_jumpdests(
-                starknet_address,
-                jumpdests + DictAccess.SIZE,
-                jumpdests_end,
-                valid_indexes_len,
-                valid_indexes,
-            );
+        loop:
+        let keys_len = [ap - 4];
+        let keys = cast([ap - 3], felt*);
+        let len = [ap - 2];
+        let dict = cast([ap - 1], DictAccess*);
+        let is_valid = dict.new_value;
+
+        if (is_valid != 0) {
+            assert [keys] = dict.key;
+            tempvar keys_len = keys_len + 1;
+            tempvar keys = keys + 1;
+            tempvar len = len - 1;
+            tempvar dict = dict + DictAccess.SIZE;
+        } else {
+            tempvar keys_len = keys_len;
+            tempvar keys = keys;
+            tempvar len = len - 1;
+            tempvar dict = dict + DictAccess.SIZE;
         }
 
-        assert valid_indexes[valid_indexes_len] = jumpdests.key;
-        return _save_valid_jumpdests(
-            starknet_address,
-            jumpdests + DictAccess.SIZE,
-            jumpdests_end,
-            valid_indexes_len + 1,
-            valid_indexes,
-        );
+        static_assert keys_len == [ap - 4];
+        static_assert keys == [ap - 3];
+        static_assert len == [ap - 2];
+        static_assert dict == [ap - 1];
+
+        jmp loop if len != 0;
+
+        IAccount.write_jumpdests(starknet_address, jumpdests_len=keys_len, jumpdests=keys_start);
+        return ();
     }
 }
--- a/src/backend/starknet.cairo
+++ b/src/backend/starknet.cairo
@@ -208,6 +208,12 @@ namespace Internals {
             IAccount.set_nonce(starknet_address, self.nonce);
             // Save storages
             _save_storage(starknet_address, self.storage_start, self.storage);
+
+            // Save valid jumpdests
+            let (valid_indexes) = alloc();
+            Internals._save_valid_jumpdests(
+                starknet_address, self.valid_jumpdests_start, self.valid_jumpdests, 0, valid_indexes
+            );
             return ();
         }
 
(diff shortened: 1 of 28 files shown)


src/kakarot/account.cairo
```diff
@@ -502,73 +530,35 @@ namespace Account {
...
-    // @dev This function iterates over the bytecode from the current index 'i'.
-    // If the opcode at the current index is between 0x5f and 0x7f (PUSHN opcodes) (inclusive),
-    // it skips the next 'n_args' opcodes, where 'n_args' is the opcode minus 0x5f.
-    // If the opcode is 0x5b (JUMPDEST), it marks the current index as a valid jump destination.
-    // It continues by jumping back to the body flag until it has processed the entire bytecode.
-    func get_jumpdests{range_check_ptr}(bytecode_len: felt, bytecode: felt*) -> (
+    // @notice Return the valid jumpdests related to this account's code, cached during the execution of the transaction.
+    // @param self The pointer to the Account
+    // @return The pointer to the dictionary of the valid jumpdests
+    func cached_jumpdests(self: model.Account*) -> (
```
ClementWalter, line 536: remove this

src/kakarot/account.cairo
```diff
@@ -502,73 +530,35 @@ namespace Account {
...
-            tempvar range_check_ptr = range_check_ptr;
-        } else {
-            tempvar valid_jumpdests = valid_jumpdests;
-            tempvar next_i = next_i;
-            tempvar range_check_ptr = range_check_ptr;
-        }
+        return (self.valid_jumpdests_start, self.valid_jumpdests);
+    }
+
+    func cache_valid_jumpdests{range_check_ptr}(
```
ClementWalter, line 542: I'd just call this `set_valid_jumpdests` as it's just setting an attribute. The "cache" meaning is not related to this function

src/kakarot/interfaces/interfaces.cairo
```diff
@@ -76,7 +76,10 @@ namespace IAccount {
     func set_nonce(nonce: felt) {
     }
 
-    func is_empty() -> (is_empty: felt) {
+    func is_jumpdest_valid(index: felt) -> (is_valid: felt) {
```
ClementWalter, line 79: can we always use the two words `valid_jumpdest`:

```
is_valid_jumpdest(...)
set_valid_jumpdests(...)
```

src/kakarot/evm.cairo
```diff
@@ -278,3 +287,32 @@ namespace EVM {
...
+
+        let code_starknet_address = Account.compute_starknet_address(code_address);
+        let (is_valid) = IAccount.is_jumpdest_valid(code_starknet_address, index);
+
+        if (is_valid == 0) {
+            return FALSE;
+        }
+
+        dict_write{dict_ptr=valid_jumpdests}(index, TRUE);
+        return TRUE;
```
ClementWalter, line 316: so this is

```
dict_write{dict_ptr=valid_jumpdests}(index, is_valid);

return is_valid;
```

and you remove the if/else at the cost of writing 0 while it's not necessary, meaning that you save some steps for valid cases and use some more for invalid. I think it's then better

src/kakarot/interpreter.cairo
```diff
@@ -1029,8 +1040,18 @@ namespace Internals {
             return evm;
         }
 
+        // Write bytecode and cached the final code valid jumpdests to Account
```
ClementWalter, line 1043: *cache the final

tests/end_to_end/test_kakarot.py
```diff
@@ -60,15 +60,17 @@ async def class_hashes():
 @pytest_asyncio.fixture(scope="session")
 async def origin(evm: Contract, addresses):
```
ClementWalter, line 61: unused addresses arg

tests/end_to_end/test_kakarot.py
```diff
@@ -116,7 +118,7 @@ async def test_execute(
                     int(x)
                     for x in params["stack"]
                     .format(
-                        account_address=int(addresses[0].address, 16),
+                        account_address=origin,
```
ClementWalter, line 121: remove address arg then

tests/src/backend/test_starknet.py
```diff
@@ -1,6 +1,32 @@
...
+            valid_jumpdests = [(0x01, 0, 1), (0x10, 0, 1), (0x101, 0, 1)]
+            contract_address = 0x97283590
+            valid_indexes = cairo_run(
+                "test__save_valid_jumpdests",
+                jumpdests=valid_jumpdests,
+                contract_address=contract_address,
+            )
+
+            assert valid_indexes == [
+                valid_jumpdests[i][0] for i in range(len(valid_jumpdests))
```
ClementWalter, line 25:
```
valid_jumpdest[0] for valid_jumpdest in valid_jumpdests
```

tests/end_to_end/test_kakarot.py
enitrat: from where?

src/backend/starknet.cairo
```diff
@@ -295,4 +304,42 @@ namespace Internals {
 
         return _save_storage(starknet_address, storage_start + DictAccess.SIZE, storage_end);
     }
+
+    func _save_valid_jumpdests{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr}(
```
ClementWalter, line 308: so this is like `dict_keys` filtered for values > 0, I think that you can refactor this similar to the dict_keys util. Then, call the `write_jumpdests`.

src/kakarot/accounts/library.cairo
```diff
@@ -440,7 +470,7 @@ namespace Internals {
     }
 
     // @notice Store the bytecode of the contract.
-    // @param index The current free index in the bytecode_ storage.
+    // @param index The current free index in the Account_bytecode storage.
```
ClementWalter, line 473: well actually there is no `index` param so this line can be removed

src/kakarot/accounts/library.cairo
```diff
@@ -493,6 +523,44 @@ namespace Internals {
...
+    // @param jumpdests The jumpdests of the contract.
+    func write_jumpdests{syscall_ptr: felt*}(jumpdests_len: felt, jumpdests: felt*) {
+        alloc_locals;
+
+        if (jumpdests_len == 0) {
+            return ();
+        }
+
+        // selector!("Account_valid_jumpdests")
+        local base_address = 0x01eb4a0d1c10536edc8183a507de5f884c0a18b4877da7c986a6100d6bfda343;
```
ClementWalter, line 538: so this is `Account_valid_jumpdests.addr();` ?

src/kakarot/accounts/library.cairo
```diff
@@ -493,6 +523,44 @@ namespace Internals {
...
+        tempvar jumpdests_len = jumpdests_len;
+
+        body:
+        let syscall_ptr = cast([ap - 2], felt*);
+        let jumpdests_len = [ap - 1];
+        let initial_jumpdests_len = [fp - 4];
+        let jumpdests = cast([fp - 3], felt*);
+        let base_address = [fp];
+
+        tempvar index_to_store = jumpdests[initial_jumpdests_len - jumpdests_len];
```
ClementWalter, line 549: could be a `let` here I guess

src/kakarot/account.cairo
```diff
@@ -502,73 +530,26 @@ namespace Account {
...
-            tempvar range_check_ptr = range_check_ptr;
-        } else {
-            tempvar valid_jumpdests = valid_jumpdests;
-            tempvar next_i = next_i;
-            tempvar range_check_ptr = range_check_ptr;
-        }
+    func set_valid_jumpdests{range_check_ptr}(
+        self: model.Account*, valid_jumpdests_start: DictAccess*, valid_jumpdests: DictAccess*
+    ) -> model.Account* {
+        let (copy_start, copy) = default_dict_copy(valid_jumpdests_start, valid_jumpdests);
```
ClementWalter, line 536: why do we need to copy here?

src/kakarot/accounts/library.cairo
enitrat: yes but if I use `.addr()` I need to manually handle `pedersen_ptr` and `range_check_ptr` and re-assign them at each iteration for nothing

src/kakarot/account.cairo
enitrat: that's what I was discussing the other day, I could simply __move__ the dict from the message to the account, and replace the one in the message with a new, default, empty one.

However, I cannot have the same pointers in both structs, as eventually the message's one is going to be squashed and that's going to have repercussions on the account's one.

src/kakarot/evm.cairo
```diff
@@ -278,3 +287,28 @@ namespace EVM {
...
+
+namespace Internals {
+    func is_valid_jumpdest{
+        syscall_ptr: felt*,
+        pedersen_ptr: HashBuiltin*,
+        range_check_ptr,
+        valid_jumpdests: DictAccess*,
+    }(code_address: felt, is_create: felt, index: felt) -> felt {
+        alloc_locals;
+        let (is_cached) = dict_read{dict_ptr=valid_jumpdests}(index);
```
ClementWalter, line 299: so this is more `is_valid` like below

src/kakarot/interpreter.cairo
```diff
@@ -1029,8 +1040,16 @@ namespace Internals {
...
+        // Write bytecode and cache the final code valid jumpdests to Account
         let account = State.get_account(evm.message.address.evm);
         let account = Account.set_code(account, evm.return_data_len, evm.return_data);
+        let account = Account.set_created(account, TRUE);
+
+        let (valid_jumpdests_start, valid_jumpdests) = Helpers.initialize_jumpdests(
+            account.code_len, account.code
+        );
+        let account = Account.set_valid_jumpdests(account, valid_jumpdests_start, valid_jumpdests);
+
```
ClementWalter, line 1052: we may want to just create a `new Account` with all the parameters set to their final value, instead of using the methods and creating 5x a new account

src/kakarot/evm.cairo
enitrat: not really, because it could be valid, but not cached. here the semantic of the variable name really carries the intent.

tests/end_to_end/test_kakarot.py
ClementWalter: from the

```
        async def test_execute(
            self,
            starknet: FullNodeClient,
            eth: Contract,
            params: dict,
            request,
            evm: Contract,
            addresses,
            max_fee,
            origin,
        ):
```

tests/src/backend/test_starknet.py
```diff
@@ -1,6 +1,32 @@
...
     class TestGetEnv:
         def test_should_return_env_with_given_origin_and_gas_price(self, cairo_run):
             env = cairo_run("test__get_env", origin=1, gas_price=2)
             assert env["origin"] == 1
             assert env["gas_price"] == 2
+
+    class TestSaveTestStarknetJumpdests:
+        @SyscallHandler.patch("IAccount.write_jumpdests", lambda addr, data: [])
+        def test_should_save_jumpdests_to_storage(self, cairo_run):
+            valid_jumpdests = [(0x01, 0, 1), (0x10, 0, 1), (0x101, 0, 1)]
```
ClementWalter, line 16: no false case here
I think that the test would be more readable with an input defined like

```
valid_jumpdests = {
    0x1: True
    0x10: False
    0x101: True
}
```

and use something like the `dict_new` function for the core lib

```
func dict_new() -> (res: DictAccess*) {
    %{
        if '__dict_manager' not in globals():
            from starkware.cairo.common.dict import DictManager
            __dict_manager = DictManager()

        memory[ap] = __dict_manager.new_dict(segments, initial_dict)
        del initial_dict
    %}
    ap += 1;
    return (res=cast([ap - 1], DictAccess*));
}
```

tests/src/kakarot/accounts/test_contract_account.py
```diff
@@ -141,6 +141,55 @@ def test_should_read_bytecode(self, cairo_run, bytecode, storage):
             mock_storage.assert_has_calls(calls)
             assert output[:output_len] == list(bytecode)
 
+    class TestJumpdests:
+        class TestWriteJumpdests:
+            def test__should_store_valid_jumpdests(self, cairo_run):
+                jumpdests = [0x02, 0x10, 0xFF]
+                base_address = get_storage_var_address("Account_valid_jumpdests")
```
ClementWalter, line 148: write this after the `cairo_run` as it's unused before the assert step

tests/src/kakarot/accounts/test_contract_account.py
```diff
@@ -141,6 +141,55 @@ def test_should_read_bytecode(self, cairo_run, bytecode, storage):
...
+                cairo_run("test__write_jumpdests", jumpdests=jumpdests)
+
+                SyscallHandler.mock_storage.assert_any_call(
```
ClementWalter, line 151: can't we build the `calls` list object and use `mock_storage.assert_calls`?

tests/src/kakarot/accounts/test_contract_account.py
```diff
@@ -141,6 +141,55 @@ def test_should_read_bytecode(self, cairo_run, bytecode, storage):
...
+            def storage(self, jumpdests):
+                base_address = get_storage_var_address("Account_valid_jumpdests")
+                valid_addresses = [base_address + jumpdest for jumpdest in jumpdests]
+
+                def _storage(address):
+                    return 1 if address in valid_addresses else 0
+
+                return _storage
+
+            @pytest.mark.parametrize("jumpdests", [[0x02, 0x10, 0xFF]])
```
ClementWalter, line 172: create a fixture for `jumpdets`, here this is a hack

tests/src/kakarot/accounts/test_contract_account.py
```diff
@@ -141,6 +141,55 @@ def test_should_read_bytecode(self, cairo_run, bytecode, storage):
...
+                self, cairo_run, jumpdests, storage
+            ):
+                with patch.object(SyscallHandler, "mock_storage", side_effect=storage):
+                    assert cairo_run("test__is_valid_jumpdest", index=0x02) == 1
+                    assert cairo_run("test__is_valid_jumpdest", index=0x10) == 1
+                    assert cairo_run("test__is_valid_jumpdest", index=0xFF) == 1
+                    assert cairo_run("test__is_valid_jumpdest", index=0xABC) == 0
+
+                    base_address = get_storage_var_address("Account_valid_jumpdests")
+                    SyscallHandler.mock_storage.assert_any_call(
```
ClementWalter, line 183: see above for asserting the whole list of calls in one line

tests/utils/helpers.cairo
```diff
@@ -115,6 +116,43 @@ namespace TestHelpers {
         return memory;
     }
 
+    func init_jumpdests_with_values(jumpdests_len: felt, jumpdests: felt*) -> (
```
ClementWalter, line 119: see `dict_new`

src/kakarot/accounts/library.cairo
ClementWalter: indeed, or put them in locals to be able to rebind at the end.
we can stay like that, but then plz update the comment from cairo1 style to

```
// hardcoded value of Account_valid_jumpdests.addr() to avoid rebinding ...
```

src/backend/starknet.cairo
```diff
@@ -295,4 +302,53 @@ namespace Internals {
...
+    ) {
+        alloc_locals;
+        let (local keys_start: felt*) = alloc();
+        let dict_len = dict_end - dict_start;
+        let (local len, _) = unsigned_div_rem(dict_len, DictAccess.SIZE);
+        local range_check_ptr = range_check_ptr;
+
+        if (dict_len == 0) {
+            return ();
+        }
```
ClementWalter, line 317: do the early return immediately, not after some other computations

```

let dict_len = dict_end - dict_start;
if (dict_len == 0) {
    return ();
}

```

src/backend/starknet.cairo
```diff
@@ -295,4 +302,53 @@ namespace Internals {
...
         return _save_storage(starknet_address, storage_start + DictAccess.SIZE, storage_end);
     }
+
+    func _save_valid_jumpdests{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr}(
+        starknet_address: felt, dict_start: DictAccess*, dict_end: DictAccess*
+    ) {
+        alloc_locals;
+        let (local keys_start: felt*) = alloc();
+        let dict_len = dict_end - dict_start;
+        let (local len, _) = unsigned_div_rem(dict_len, DictAccess.SIZE);
```
ClementWalter, line 312: no need to use a range_check, there is no way to have `dict_len` not a multiple of `DictAccess.SIZE`.

But in any case you don't really need this, you can retrieve in the body loop `dict_end` with `[fp - 3]`, update `dict_start = dict_start + DictAccess.SIZE` and break when start == end

ClementWalter: *to be compared in step with having this extra `len` variable in the loop, looks like it the same for the loop part, but you'd still save the initial div

src/backend/starknet.cairo
```diff
@@ -295,4 +302,53 @@ namespace Internals {
...
+            tempvar keys_len = keys_len + 1;
+            tempvar keys = keys + 1;
+            tempvar len = len - 1;
+            tempvar dict = dict + DictAccess.SIZE;
+        } else {
+            tempvar keys_len = keys_len;
+            tempvar keys = keys;
+            tempvar len = len - 1;
+            tempvar dict = dict + DictAccess.SIZE;
+        }
```
ClementWalter, line 342: can you move the common part out of the `if/else`? `jmp if != 0` doesn't update `[ap]` so it should still work as is. If not, maybe use raw `jmp` instead

src/kakarot/accounts/library.cairo
```diff
@@ -493,6 +522,47 @@ namespace Internals {
         return ();
     }
 
+    // @notice Store the jumpdests of the contract.
+    // @param index The current free index in the jumpdests_ storage.
```
ClementWalter, line 526: remove

src/kakarot/accounts/library.cairo
```diff
@@ -493,6 +522,47 @@ namespace Internals {
...
+
+        if (jumpdests_len == 0) {
+            return ();
+        }
+
+        let (local base_address) = Account_valid_jumpdests.addr();
+        tempvar syscall_ptr = syscall_ptr;
+        tempvar jumpdests_len = jumpdests_len;
+        local pedersen_ptr: HashBuiltin* = pedersen_ptr;
+        local range_check_ptr = range_check_ptr;
```
ClementWalter, line 542: assign `local` before the `tempvar` as the locals are here to save the `.addr()` returned values (sort of cleanup of the previous line) while the `tempvar` are here to prepare the loop (next statement)

src/kakarot/account.cairo
```diff
@@ -56,13 +57,44 @@ namespace Account {
...
         );
     }
 
+    // @notice Finalizes the creation of an account by setting the account's bytecode, valid jumpdests
+    //        and marking it as created during this transaction.
+    // @param self The pointer to the Account
+    // @param code_len The length of the code
+    // @param code The pointer to the code
+    // @param valid_jumpdests_start The pointer to the start of the valid jumpdests dict
+    // @param valid_jumpdests The pointer to the next free valid jumpdests dict slot
```
ClementWalter, line 75: fix docstring

src/kakarot/account.cairo
```diff
@@ -502,73 +559,26 @@ namespace Account {
...
-            dict_write{dict_ptr=valid_jumpdests}(i, TRUE);
-            tempvar valid_jumpdests = valid_jumpdests;
-            tempvar next_i = next_i;
-            tempvar range_check_ptr = range_check_ptr;
-        } else {
-            tempvar valid_jumpdests = valid_jumpdests;
-            tempvar next_i = next_i;
-            tempvar range_check_ptr = range_check_ptr;
-        }
+    func set_valid_jumpdests{range_check_ptr}(
```
ClementWalter, line 562: add docstring, and especially `@dev` for the why it's important to `default_dict_copy`

src/backend/starknet.cairo
enitrat: I don't understand how I can move the common part (dict = dict + DictAccess.SIZE) outside of the if/else. It just won't compile.

src/backend/starknet.cairo
```diff
@@ -295,4 +302,48 @@ namespace Internals {
...
+        let dict = cast([ap - 2], DictAccess*);
+        let is_valid = dict.new_value;
+
+        if (is_valid != 0) {
+            assert [keys] = dict.key;
+            tempvar keys_len = keys_len + 1;
+            tempvar keys = keys + 1;
+            tempvar dict = dict + DictAccess.SIZE;
+        } else {
+            tempvar keys_len = keys_len;
```
ClementWalter, line 334: `keys_len` is not necessary, just recompute at the end with `keys - keys_start` (or remove `keys` and `assert [keys_start + keys_len] == dict.key`)

src/kakarot/account.cairo
```diff
@@ -56,13 +57,42 @@ namespace Account {
...
         );
     }
 
+    // @notice Finalizes the creation of an account by setting the account's bytecode, valid jumpdests
+    //        and marking it as created during this transaction.
+    // @param self The pointer to the Account
+    // @param code_len The length of the code
+    // @param code The pointer to the code
+    // @return The updated Account with the code and valid jumpdests set
+    func finalize_creation{range_check_ptr}(
```
ClementWalter, line 75: I'd have kept the original `set_code` name

ClementWalter (approved):

ClementWalter (approved):

PR #1146, opened 2024-05-15
docs: Cairo Precompiles design

enitrat:
Time spent on this PR:

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Resolves #<Issue number>

## What is the new behavior?

-
-
-

- - -
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot/1146)

### DIFF
--- a/docs/general/cairo_precompiles.md
+++ b/docs/general/cairo_precompiles.md
@@ -1,0 +1,156 @@
+# Cairo Precompiles
+
+Kakarot zkEVM being a Starknet appchain, it is technically possible to run Cairo
+Contracts on Kakarot. The purpose of this document is to explain the design
+behind the Cairo precompiles, which are the Cairo contracts that are deployed on
+Kakarot to provide additional functionality to the users.
+
+## Requirements
+
+From a developer / user perspective, the precompiles should be easy to interact
+with from EVM contracts. As such, we need to provide a way to call the
+precompiles from EVM contracts, and to return the results to the EVM contracts.
+This involves:
+
+- Converting the EVM calldata to Cairo inputs
+- Converting the Cairo outputs to EVM return data
+
+From a security perspective, the precompiles should _never_ produce Cairo errors
+that would cause the transaction to revert. As we cannot catch contract call
+errors in Cairo, we need to ensure that executing a `call_contract` syscall to a
+precompile will never cause the transaction to revert, meaning that:
+
+- The target Cairo addresses should always correspond to a deployed contract
+- The selector of the Cairo function being called should always be present in
+  the contract called
+- The Cairo contract called should never _panic_
+
+From these principles, we can derive the following design.
+
+## Design
+
+Interacting with the Cairo precompiles will only be done by calling specific EVM
+contracts that have been whitelisted for this purpose. This ensures that the
+Cairo precompiles are only called by contracts that have been reviewed and
+approved regarding the security concerns mentioned above.
+
+As such, the execution flow of an EVM message will be as follows:
+
+```mermaid
+flowchart TD
+    A[Solidity contract] -->|ICairo.call_contract| B
+    B{address type}
+    B -->|EVM Contract| D[Execute contract]
+    B -->|EVM Precompiles| E[Execute EVM precompile]
+    B -->|Cairo Precompile| F{Whitelisted caller code ?}
+    F --> |yes| G[Call Cairo contract]
+    F --> |no| H[Call fails]
+```
+
+## Implementation
+
+The solidity interface for the Cairo precompiles is as follows:
+
+```solidity
+// SPDX-License-Identifier: MIT
+
+pragma solidity >=0.7.0 <0.9.0;
+
+/// @notice Solidity interface that allows developers to interact with Cairo precompile methods. The precompile has one method `call_contract`.
+
+/// @dev The Cairo precompile contract's address.
+address constant CAIRO_PRECOMPILE_ADDRESS = 0x00000000000000000000000000000000000abde1;
+
+/// @dev The Cairo precompile contract's instance.
+Cairo constant CAIRO_PRECOMPILE_CONTRACT = ICairo(CAIRO_PRECOMPILE_ADDRESS);
+
+interface ICairo {
+    /// @dev Call Cairo contract deployed on the Starknet appchain
+    function call_contract(uint256 contractAddress, uint256 functionSelector, bytes calldata data) external returns (bytes memory);
+}
+```
+
+It contains a single method `call_contract` that allows the user to call a Cairo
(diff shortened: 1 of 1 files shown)


docs/general/cairo_precompiles.md
```diff
@@ -0,0 +1,170 @@
...
+Interacting with the Cairo precompiles will only be done by calling specific EVM
+contracts that have been whitelisted for this purpose. This ensures that the
+Cairo precompiles are only called by contracts that have been reviewed and
+approved regarding the security concerns mentioned above.
+
+As such, the execution flow of an EVM message will be as follows:
+
+```mermaid
+flowchart TD
+    A[Solidity contract] -->|ICairo.call_contract| B
```
ClementWalter, line 41: this is more `CALL` than `ICairo.call_contract`

docs/general/cairo_precompiles.md
```diff
@@ -0,0 +1,170 @@
...
+    B{address type}
+    B -->|EVM Contract| D[Execute contract]
+    B -->|EVM Precompiles| E[Execute EVM precompile]
+    B -->|Cairo Precompile| F{Whitelisted caller code ?}
```
ClementWalter, line 45: whitelisted starknet address and entrypoint

docs/general/cairo_precompiles.md
```diff
@@ -0,0 +1,170 @@
...
+## Implementation
+
+The solidity interface for the Cairo precompiles is as follows:
+
+```solidity
+// SPDX-License-Identifier: MIT
+
+pragma solidity >=0.7.0 <0.9.0;
+
+/// @notice Solidity interface that allows developers to interact with Cairo precompile methods. The precompile has one method `call_contract`.
```
ClementWalter, line 59: technically speaking, this method has no impact as we will just skip the bytes4 selector of the EVM `CALLDATA` to retrieve only the arguments (contractAddress, functionSelector and data)

docs/general/cairo_precompiles.md
```diff
@@ -0,0 +1,170 @@
...
+    /// @dev The cairo function selector to call - `set_counter()`
+    uint256 constant FUNCTION_SET_COUNTER = 0x0107cf8c3d109449e1beb4ac1ba726d3673b6f088ae454a9e0f18cb225be4712;
+
+    /// @notice Calls the Cairo contract to increment its internal counter
+    /// @return The return data of the call, serialized as bytes
+    function incrementCairoCounter() public returns (bytes memory) {
+        // `inc` takes no arguments, so data is empty
+        bytes memory data = "";
+
+        return CAIRO_PRECOMPILE_CONTRACT.call_contract(CAIRO_CONTRACT_ADDRESS, FUNCTION_SELECTOR_INC, data);
```
ClementWalter, line 134: missing somewhere

```
CAIRO_PRECOMPILE_CONTRACT = ICairoPrecompile(0xabde1)
```

docs/general/cairo_precompiles.md
enitrat: Whitelist happens at the solidity level, not starknet 🤔 (or both?)

docs/general/cairo_precompiles.md
enitrat: Yes, it's just an interface for easy interactions 🤔

docs/general/cairo_precompiles.md
```diff
@@ -0,0 +1,170 @@
...
+        uint128 newCounterHigh = uint128(newCounter >> 128);
+        bytes memory data = abi.encode(newCounterLow, newCounterHigh);
+
+        return CAIRO_PRECOMPILE_CONTRACT.call_contract(CAIRO_CONTRACT_ADDRESS, FUNCTION_SET_COUNTER, data);
+    }
+}
+```
+
+Once deployed, the contract can be called to increment the counter in a Cairo
+contract deployed at address `0xabc`. The deployment address will need to be
```
ClementWalter, line 152: at *starknet* address

docs/general/cairo_precompiles.md
```diff
@@ -0,0 +1,170 @@
...
+
+Internally, a new logic flow is implemented when processing message calls. If
+the target address is the Cairo precompile address, we check if the code_address
+of the message is whitelisted. If it is, we execute the Cairo contract. If it is
+not, we revert the transaction.
+
+To execute the Cairo contract, we need to convert the calldata in bytes to the
+expected Cairo inputs, serialized to an `Array<felt252>`. This serialization
+process must be done upfront, in the Solidity contract, before calling the
+precompile. Each 256-bit word in the calldata is truncated to 251 bit and packed
```
ClementWalter, line 163: the CALLDATA (`bytes`, ie `bytes1[]`) that we get in the EVM CALL (in Cairo), which is `abi.encode(signature)` is actually an `Array<felt252>` where each `felt < 256`

So at the Cairo level, we need to
- pick the corresponding arguments by slicing this big `felt*`
- pack each 32 bytes word into a (truncated) felt

it makes sense, we sort of need to introduce a x32 coef in the calldata size because each byte is already a felt, so we need 32 felt for 1 felt

ClementWalter: you can rephrase this `Each 256-bit word in the calldata is truncated to 251 bit and packed` with more context, and it's not 251 but just say native field element (approx 252 bits currently)

ClementWalter: maybe add an example with the counter `inc(felt)` and call with `data = [0, 0, ..., 0, 1]` to call `inc(1)`

docs/general/cairo_precompiles.md
ClementWalter: I think that we whitelist starknet contract + its entrypoints, not EVM caller. Once a precompile (starknet contract) is deemed legit, why not letting anyone else calling it ? because also a wrong calldata can trigger a VM error ?

docs/general/cairo_precompiles.md
ClementWalter: and we could also allow for `library_call` or other starknet syscalls

docs/general/cairo_precompiles.md
ClementWalter: in cairo1, `call_contract` syscall can fail safe right? so maybe we could add in the `Cairo1Helper` a function to actually do the `call_contract` with the new syscall, instead of the whitelist mechanism?

enitrat: @ClementWalter no, call_contract cannot fail safe. Ariel told me about the possibility to handle syscall errors for release 0.15 (EOY at best)

docs/general/cairo_precompiles.md
enitrat: done

PR #1148, opened 2024-05-17
feat: uninitialized jumpdests fallback

enitrat:
Time spent on this PR: 0.5d

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Resolves #<Issue number>

## What is the new behavior?

- Adds a fallback so that if a jumpdest index read is 0 and the jumpdests have not been initialized, initialize them and re-query the index
-
-

- - -
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot/1148)

### DIFF
--- a/src/kakarot/accounts/library.cairo
+++ b/src/kakarot/accounts/library.cairo
@@ -466,29 +466,32 @@ namespace AccountContract {
         tempvar syscall_ptr = syscall_ptr + StorageRead.SIZE;
         tempvar value = response.value;
 
-        if (value == 0) {
-            // Jumpdest is invalid - we verify that the jumpdests have been stored, and if not,
-            // we store them. call the appropriate function check & store
-            let (initialized) = Account_jumpdests_initialized.read();
-            if (initialized == 0) {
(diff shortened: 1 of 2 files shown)


src/kakarot/accounts/library.cairo
```diff
@@ -454,6 +466,28 @@ namespace AccountContract {
...
+                );
+                Internals.write_jumpdests(
+                    jumpdests_len=jumpdests_len,
+                    jumpdests=cast(valid_jumpdests_start, felt*),
+                    iteration_size=DictAccess.SIZE,
+                );
+                return is_valid_jumpdest(index=index);
+            }
+            return value;
+        }
```
ClementWalter, line 490: rewrite with inverted conditions so as to early return as soon as possible

```
if (value != FALSE) {
    return TRUE
}

let (initialized) = Account_jumpdests_initialized.read();

if (initialized != FALSE) {
    return FALSE
}

let (bytecode_len) = Account_bytecode_len.read();
// ...
```

PR #1178, opened 2024-06-04
feat: unpack packed calldata

### LINKED ISSUE #1175: feat: unpack transaction calldata

Calldata is currently sent with 1 byte per felt. Considering a maximum of 30k felts in calldata, and 58k steps to unpack 30k bytes of packed calldata:

Option 1: no-packing: 30k * 0.128gas/felt = 3840 gas for max calldata
Option 2: 31-bytes calldata packing: (967 * 0.128gas/felt) + (0.0025gas/steps * 471k steps) = 1 301 gas for max calldata

Option 2 (with packing) is 3x cheaper, and solves the problem of calldata limit on the starknet chain.

Thus, implement calldata unpacking when receiving a transaction. The RPC will need to pack the calldata.

enitrat:
Time spent on this PR: 1.3d

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Resolves #1175

## What is the new behavior?

-
-
-

- - -
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot/1178)

### DIFF
--- a/src/kakarot/accounts/library.cairo
+++ b/src/kakarot/accounts/library.cairo
@@ -32,7 +32,7 @@ from kakarot.errors import Errors
 from kakarot.constants import Constants
 from utils.eth_transaction import EthTransaction
 from utils.uint256 import uint256_add
-from utils.bytes import bytes_to_bytes8_little_endian, felt_to_bytes_i
+from utils.bytes import bytes_to_bytes8_little_endian
 from utils.signature import Signature
 from utils.utils import Helpers
 
--- a/src/kakarot/accounts/library.cairo
+++ b/src/kakarot/accounts/library.cairo
@@ -193,17 +193,10 @@ namespace AccountContract {
         let packed_tx_data_len = [call_array].data_len;
         let packed_tx_data = calldata + [call_array].data_offset;
 
-        let total_len_bytes = [packed_tx_data];
-        let full_words_count = packed_tx_data_len - 2;
-        let full_words_bytes = full_words_count * 31;
-        let full_words_ptr = packed_tx_data + 1;
-        let last_word_bytes = total_len_bytes - full_words_bytes;
-        let last_word = packed_tx_data[packed_tx_data_len - 1];
-
-        let (tx_data) = alloc();
-        let tx_data_len = total_len_bytes;
-        Helpers.felt_array_to_bytes31_array(full_words_count, full_words_ptr, tx_data);
-        felt_to_bytes_i(tx_data + full_words_bytes, last_word, last_word_bytes);
+        let tx_data_len = [packed_tx_data];
+        let (tx_data) = Helpers.load_packed_bytes(
+            packed_tx_data_len - 1, packed_tx_data + 1, tx_data_len
+        );
 
         let tx = EthTransaction.decode(tx_data_len, tx_data);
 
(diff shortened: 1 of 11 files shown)


src/kakarot/library.cairo
```diff
@@ -101,6 +105,15 @@ namespace Kakarot {
 
         let env = Starknet.get_env(origin, gas_price);
 
+        let full_words_len = packed_data_len - 2;
+        let last_word_bytes = packed_data[packed_data_len - 2];
+        let last_word = packed_data[packed_data_len - 1];
```
ClementWalter, line 110: what is there is no need for a last word? you put 0 in the previous felt?

I feel this custom encoding a bit unsual (inserting a len in the middle of the data).

I think that it would be closed to a regular `bytes` encoding if we just put at `packed_data[0]` the `len` of the calldata (in bytes), then the packed data.
The final word is can just be shifted eventually with

```
(full_words, last_word_shift) = unsigned_div_rem(len, 31);
let last_word_shift = 32 - last_word_shift
```

tests/utils/helpers.py
```diff
@@ -241,3 +241,54 @@ def merge_access_list(access_list):
             }
         )
     return merged_list
+
+
+def pack_calldata(data: Union[List[int], str]) -> List[int]:
```
ClementWalter, line 246: given we don't this encoding but add the len as the first item, I'd just do

```
from textwrap import wrap
# data is bytes
return [
    len(data),
    *[int(chunk) for chunk in wrap(data.hex(), 2 * 31)]
]
```

src/kakarot/accounts/library.cairo
```diff
@@ -189,17 +189,19 @@ namespace AccountContract {
...
+        // Unpack the tx data
+        let packed_tx_data_len = [call_array].data_len;
+        let packed_tx_data = calldata + [call_array].data_offset;
+
+        let tx_data_len = [packed_tx_data];
+        let (tx_data) = Helpers.load_packed_bytes(
+            packed_tx_data_len - 1, packed_tx_data + 1, tx_data_len
         );
 
+        let tx = EthTransaction.decode(tx_data_len, tx_data);
```
ClementWalter, line 201: unused

tests/src/kakarot/test_kakarot.py
```diff
@@ -126,10 +126,13 @@ def test_register_account_should_fail_caller_not_resolved_address(
             with cairo_error():
                 cairo_run("test__register_account", evm_address=EVM_ADDRESS)
 
-    @pytest.mark.NoCI
     class TestEthCall:
         @pytest.mark.SolmateERC20
-        async def test_erc20_transfer(self, get_contract):
+        @SyscallHandler.patch(
```
enitrat, line 131: I dont see a reason to skip this in CI, but to run we need to mock the valid jumdpests

1. either we conside the bytecode "valid" (optimisticallly assuming Solmate is rightfully compiled) and we can say that each query returns True
2. or we run a real analysis on the bytecode (probably not worth it)

src/utils/bytes.cairo
```diff
@@ -84,6 +84,18 @@ func felt_to_bytes(dst: felt*, value: felt) -> felt {
     return bytes_len;
 }
 
+// @notice Split a felt into an array of bytes of size i
+// @dev The array is padded with zeros if the felt is smaller than i bytes
+func felt_to_bytes_i{range_check_ptr}(dst: felt*, value: felt, bytes_len: felt) -> felt {
```
ClementWalter, line 89: unused

tests/src/kakarot/test_kakarot.py
```diff
@@ -216,7 +224,7 @@ async def test_case(
         async def test_failing_contract(self, cairo_run):
             initial_state = {
                 CONTRACT_ADDRESS: {
-                    "code": bytes.fromhex("0xADD_CODE"),
+                    "code": bytes.fromhex("ADDC0DE1"),
```
ClementWalter, line 227: can fix the `0xADD_DATA` then as well

tbh this skipped test as been added as a placeholder for debugging a future contract, I think that we could even delete it

tests/src/kakarot/test_kakarot.py
enitrat: If the purpose is to run it with profiling I can keep it NoCI - otherwise I think returning True is reasonable.

tests/src/kakarot/test_kakarot.py
enitrat: We can keep the placeholder, but I changed this to show the actual format expected in this bytes.fromhex()

ClementWalter (approved):

PR #1251, opened 2024-07-03
feat: add delegatecall in CairoLib

enitrat:
Time spent on this PR: 0.4d

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Resolves #<Issue number>

## What is the new behavior?

- Adds a feature to delegatecall to a CairoLib to preserve the msg.sender context
-
-

### DIFF
--- a/solidity_contracts/src/CairoPrecompiles/CairoCounterCaller.sol
+++ b/solidity_contracts/src/CairoPrecompiles/CairoCounterCaller.sol
@@ -32,11 +32,14 @@ contract CairoCounterCaller {
     }
 
     /// @notice Calls the Cairo contract to increment its internal counter
+    /// @dev The delegatecall preserves the caller's context, so the caller's address will
+    /// be the caller of this function.
     function incrementCairoCounter() external {
-        cairoCounter.callContract("inc");
+        cairoCounter.delegatecallContract("inc");
     }
 
     /// @notice Calls the Cairo contract to set its internal counter to an arbitrary value
+    /// @dev Called with a regular call, the caller's address will be this contract's address
     /// @dev The counter value is split into two 128-bit values to match the Cairo contract's expected inputs (u256 is composed of two u128s)
     /// @param newCounter The new counter value to set
     function setCairoCounter(uint256 newCounter) external {
--- a/src/kakarot/interpreter.cairo
+++ b/src/kakarot/interpreter.cairo
@@ -65,8 +65,6 @@ namespace Interpreter {
         let is_pc_ge_code_len = is_le(evm.message.bytecode_len, pc);
         if (is_pc_ge_code_len != FALSE) {
             let is_precompile = Precompiles.is_precompile(evm.message.code_address.evm);
-            let caller_address = evm.message.caller;
-            // Caller of the contract that is calling the precompile
             if (is_precompile != FALSE) {
                 // If the precompile is called straight from an EOA, the sender_context is the EOA
                 // Otherwise, the sender_context is the caller of the contract that is calling the precompile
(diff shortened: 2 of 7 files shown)


solidity_contracts/src/CairoPrecompiles/CairoCounterCaller.sol
```diff
@@ -32,11 +32,14 @@ contract CairoCounterCaller {
     }
 
     /// @notice Calls the Cairo contract to increment its internal counter
+    /// @dev The delegatecall preserves the caller's context, so the caller's address will
+    /// be the caller of this function.
     function incrementCairoCounter() external {
-        cairoCounter.callContract("inc");
+        cairoCounter.delegatecallContract("inc");
```
obatirou, line 38: With this delegatecall, now the storage of this function can be modified. 
The cairoCounter should be transformed as an `immutable` as a best practice.

src/kakarot/precompiles/kakarot_precompiles.cairo
```diff
@@ -32,14 +32,13 @@ namespace KakarotPrecompiles {
     // @notice Executes a cairo contract/class.
     // @param input_len The length of the input in bytes.
     // @param input The input data.
-    // @param caller_address The address of the contract that calls the precompile
-    // @param sender_context The address of the sender in the context of the caller contract.
+    // @param caller_address The address of the caller of the precompile. Delegatecall rules applies.
```
ClementWalter, line 35:
```suggestion
    // @param caller_address The address of the caller of the precompile. Delegatecall rules apply.
```

src/kakarot/interpreter.cairo
```diff
@@ -65,27 +65,28 @@ namespace Interpreter {
...
             if (is_precompile != FALSE) {
                 // If the precompile is called straight from an EOA, the sender_context is the EOA
                 // Otherwise, the sender_context is the caller of the contract that is calling the precompile
                 // This is only relevant for the Kakarot Cairo Module precompile.
                 let parent_context = evm.message.parent;
                 let is_parent_zero = Helpers.is_zero(cast(parent_context, felt));
                 if (is_parent_zero != FALSE) {
-                    tempvar sender_context = evm.message.caller;
+                    // Case A: The precompile is called straight from an EOA
+                    tempvar caller_code_address = evm.message.caller;
```
obatirou, line 76: I think it is a little bit confusing as If this is an EOA there is no code
And `caller_code_address` is defined
> The address of the code of the contract that calls the precompile.

Unless I miss something ?

src/kakarot/precompiles/precompiles.cairo
```diff
@@ -57,11 +57,11 @@ namespace Precompiles {
...
     // @notice Executes associated function of precompiled evm_address.
     // @dev This function uses an internal jump table to execute the corresponding precompile impmentation.
-    // @param evm_address The precompile evm_address.
+    // @param precompile_address The precompile evm_address.
     // @param input_len The length of the input array.
     // @param input The input array.
-    // @param caller_address The address of the contract that calls the precompile.
-    // @param sender_context The address of the sender in the context of the caller contract.
+    // @param caller_code_address The address of the code of the contract that calls the precompile.
+    // @param caller_address The address of the caller of the precompile. Delegatecall rules applies.
```
ClementWalter, line 64:
```suggestion
    // @param caller_address The address of the caller of the precompile. Delegatecall rules apply.
```

ClementWalter (commented):
ok for me, just two typos in the docstring

src/kakarot/interpreter.cairo
obatirou: also the above comment is not updated and mention sender_context that is no more relevant

obatirou: But in fact after discussion for an EOA the code composant exist but is only empty so not a problem to have `caller_code_address == EOA address`.

solidity_contracts/src/CairoPrecompiles/CairoCounterCaller.sol
enitrat: Is this actually a risk, knowing that we're actually calling `delegateCall` on a "precompile" address that will never touch storage? 

I'm open to making this immutable of course

enitrat: also `{call | staticcall | delegatecall}contract` is probably a bad naming because it doesn't explicitly indicate that we're calling a cairo contract.

I propose renaming to `{call | staticcall | delegatecall}Cairo`

obatirou: This is not a risk in this context as, as you said, this "precompile" will never touch storage.
And futur precompiles should not also.
I agree the naming `{call | staticcall | delegatecall}Cairo` is better

Note: my remark  above was more of a nitpick and best practice around delegatecall.
It is also a gas opti as access to immutable will be less expensive than a storage read.

obatirou (approved):

ClementWalter (approved):
I like the renaming of the CairoLib functions

PR #1253, opened 2024-07-04
feat: deploy starknet account if CairoLib caller is not deployed

enitrat:
Time spent on this PR: 0.3d

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Resolves #<Issue number>

## What is the new behavior?

- Deploys a SN account for a caller of a Cairo contract if its account is not deployed on starknet already.
-
-

- - -
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot/1253)

### DIFF
--- a/src/backend/starknet.cairo
+++ b/src/backend/starknet.cairo
@@ -19,7 +19,7 @@ from starkware.starknet.common.syscalls import (
 )
 
 from kakarot.account import Account
-from kakarot.precompiles.precompiles import Precompiles
+from kakarot.precompiles.precompiles_helpers import PrecompilesHelpers
 from kakarot.constants import Constants
 from kakarot.interfaces.interfaces import IERC20, IAccount
 
@@ -172,7 +172,7 @@ namespace Internals {
     }(self: model.Account*, native_token_address) {
         alloc_locals;
 
-        let is_precompile = Precompiles.is_precompile(self.address.evm);
+        let is_precompile = PrecompilesHelpers.is_precompile(self.address.evm);
         if (is_precompile != FALSE) {
             return ();
         }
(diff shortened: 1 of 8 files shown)


src/backend/starknet.cairo
```diff
@@ -183,42 +183,31 @@ namespace Internals {
...
-                starknet_address, self.valid_jumpdests_start, self.valid_jumpdests
-            );
-            return ();
+            tempvar syscall_ptr = syscall_ptr;
+            tempvar pedersen_ptr = pedersen_ptr;
+            tempvar range_check_ptr = range_check_ptr;
+        } else {
+            tempvar syscall_ptr = syscall_ptr;
+            tempvar pedersen_ptr = pedersen_ptr;
+            tempvar range_check_ptr = range_check_ptr;
```
ClementWalter, line 192: if we update the signature of `Starknet.deploy` to not return the (useless) starknet address, we can save 3 steps here

```suggestion
        } else {
            tempvar syscall_ptr = syscall_ptr;
            tempvar pedersen_ptr = pedersen_ptr;
            tempvar range_check_ptr = range_check_ptr;
         }
         let syscall_ptr = cast([ap - 3], felt*);
         let pedersen_ptr = ...
```

src/kakarot/precompiles/kakarot_precompiles.cairo
```diff
@@ -82,11 +83,22 @@ namespace KakarotPrecompiles {
...
+                let (deployed_contract) = Starknet.deploy(caller_address);
+                tempvar caller_starknet_address = deployed_contract;
+                tempvar syscall_ptr = syscall_ptr;
+                tempvar pedersen_ptr = pedersen_ptr;
+                tempvar range_check_ptr = range_check_ptr;
+            } else {
+                tempvar caller_starknet_address = caller_starknet_address;
+                tempvar syscall_ptr = syscall_ptr;
+                tempvar pedersen_ptr = pedersen_ptr;
+                tempvar range_check_ptr = range_check_ptr;
```
ClementWalter, line 96: ok, forget my comment above but here you can do
```suggestion
                // Deploy account
                Starknet.deploy(caller_address);
            } else {
                tempvar syscall_ptr = syscall_ptr;
                tempvar pedersen_ptr = pedersen_ptr;
                tempvar range_check_ptr = range_check_ptr;
                tempvar caller_starknet_address = caller_starknet_address;                
```

enitrat: I don't get it - how do you get the caller starknet address?

tests/utils/syscall_handler.py
```diff
@@ -510,6 +511,61 @@ def send_message_to_l1(self, segments, syscall_ptr):
...
+    def patch_deploy(cls, value: callable):
+        """
+        Patch the deploy syscall with the value.
+
+        :param value: The value to patch with, a callable that will be called with the class hash,
+            the contract address salt, the constructor calldata and the deploy from zero flag.
+        """
+        cls.patches["deploy"] = value
+        yield
+        del cls.patches["deploy"]
```
ClementWalter, line 567: if we use the same `patches` object, why not keeping one single `patch` decorator and doing `@SyscallHandler.patch("deploy", ...)`

maybe better to do `@SyscallHandler.patch("deploy_syscall", ...)`

src/kakarot/precompiles/kakarot_precompiles.cairo
ClementWalter: the return data of the `Starknet.deploy` is in this order

```
syscall_ptr: felt*
pedersen_ptr: HashBuiltin*
range_check_ptr
account_address
```

so you don't need to tempvar, just use the same ordering

tests/utils/syscall_handler.py
enitrat: tbh I think having "patch_storage" "patch_call" "patch_deploy" would make more sense than a one-fits-all

enitrat: I personnally prefer this way and would even create the `patch_storage` / `patch_call` distinction but it's not very useful refacto

ClementWalter: it's debatable, the official lib ha one unique `patch` decorator

ClementWalter (approved):

PR #1290, opened 2024-07-19
fix: assert value is 0 at the end of load_bytecode

### LINKED ISSUE #1279: bug: assert value is 0 at the end of load_bytecode
## Bug Report

At the end of [`load_bytecode`](../../blob/main/src/kakarot/accounts/library.cairo), we need to assert that `value` is 0

```
jmp cond if remaining_bytes != 0;

assert value = 0;
let bytecode = cast([fp], felt*);
return (bytecode=bytecode);
```

### LINKED ISSUE #1293: feat: add a patch_hint test util

For some test regarding soundness, we need to write malicious hints instead of the right one.

### What

Add a patch_hint test utils to be used to patch hints at runtime

obatirou:
Time spent on this PR:

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Do not assert the value is 0 at the end of load_bytecode leading to prover being able to 

Resolves #1279, resolves #1293

## What is the new behavior?
assert value is 0 at the end of load_bytecode

- - -
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot/1290)

### DIFF
--- a/src/kakarot/accounts/library.cairo
+++ b/src/kakarot/accounts/library.cairo
@@ -652,13 +652,17 @@ namespace Internals {
 
         jmp cond if remaining_bytes != 0;
 
-        assert value = 0;
+        with_attr error_message("Value is not empty") {
+            assert value = 0;
+        }
         let bytecode = cast([fp], felt*);
         return (bytecode=bytecode);
 
         cond:
         jmp body if count != 0;
-        assert value = 0;
+        with_attr error_message("Value is not empty") {
+            assert value = 0;
+        }
         jmp read;
     }
 }
--- a/src/kakarot/accounts/library.cairo
+++ b/src/kakarot/accounts/library.cairo
@@ -652,6 +652,7 @@ namespace Internals {
 
         jmp cond if remaining_bytes != 0;
 
+        assert value = 0;
         let bytecode = cast([fp], felt*);
         return (bytecode=bytecode);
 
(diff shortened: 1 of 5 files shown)


src/kakarot/accounts/library.cairo
```diff
@@ -684,6 +684,7 @@ namespace Internals {
 
         jmp cond if remaining_bytes != 0;
 
+        assert value = 0;
```
ClementWalter, line 687: my bad in the PR description, as per zellic comment we need to do this at the end of each chunk, not at the end of the whole read. So missing here as well

```
        cond:
        jmp body if count != 0;
        assert value = 0;
        jmp read;
```

ClementWalter (changes requested):
add missing condition

tests/utils/hints.py
```diff
@@ -45,3 +48,28 @@ def new_default_dict(
...
+                )
+            )
+            for hint_ in v
+        ]
+        for k, v in program.hints.items()
+    }
+    with patch.object(program, "hints", new=patched_hints):
+        yield
+
+    program.hints = hints_before
```
ClementWalter, line 75:
```suggestion
    patched_hints = {
        k: [
            (
                hint_
                if hint_.code != hint
                else CairoHint(
                    accessible_scopes=hint_.accessible_scopes,
                    flow_tracking_data=hint_.flow_tracking_data,
                    code=new_hint,
                )
            )
            for hint_ in v
        ]
        for k, v in program.hints.items()
    }
    with patch.object(program, "hints", new=patched_hints):
        yield
```

it shouldn't be necessary to keep a copy of the original hints, it's the purpose of `patch`

ClementWalter (approved):

PR #1298, opened 2024-07-23
Fix memory expansion cost

### LINKED ISSUE #1297: bug: max_memory_expansion_cost overflow and only one size is 0
### Kakarot version

For the Gas.max_memory_expansion_cost function, this function seems not to handle the case where one memory access has zero size and the other does not, e.g. One access has the offset and size (32,  32) and the other has (0xffff...ffff, 0).

Also, it seems the Gas.max_memory_expansion_cost function adds (offset_1.high + size_1.high + offset_2.high + size_2.high) * Gas.MEMORY_COST_U128 to the total expansion cost, and we believe this can overflow since offset and size are fully controllable by a user and Gas.MEMORY_COST_U128 is the 238-bit value, which overflows when multiplied with 128-bit value.

ClementWalter:
Time spent on this PR: 0.3d

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

- overflow not checked
- wrong result when one of the sizes is 0

Resolves #1297

## What is the new behavior?

Fixed

- - -
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot/1298)


src/kakarot/gas.cairo
```diff
@@ -137,25 +137,37 @@ namespace Gas {
...
+        no_expansion:
+        let range_check_ptr = [fp - 8];
+        let expansion = model.MemoryExpansion(cost=0, new_words_len=words_len);
+        return expansion;
 
+        expansion_cost_saturated:
+        let range_check_ptr = [fp - 8];
         let expansion = model.MemoryExpansion(
-            cost=expansion_cost, new_words_len=low_expansion.new_words_len
+            cost=Gas.MEMORY_COST_U128, new_words_len=0x8000000000000000000000000000000
```
obatirou, line 170: Maybe add the same explanation as in `memory_expansion_cost_saturated` to have context on why their is a hardcoded value and what will happen in this case ?
https://github.com/kkrt-labs/kakarot/blob/7bc70ad9965ecb69d5242957594339975a0b3a8c/src/kakarot/gas.cairo#L114

obatirou (approved):

PR #1307, opened 2024-07-25
Fix RLP.decode

### LINKED ISSUE #1306: bug: RLP assertions and test

To recap, summary of transaction decoding/validation is needed:
- calls to RLP.decode do not check the return value indicating how many items were decoded
- all functions handling specific transaction types do not check the subitems length
- most if not all usages of RLP subitems do not assert the subitem type to be (or not be) a list

Also need for fixing the RLP serialization for testing.

ClementWalter:
## Pull request type

- [x] Bugfix
- [x] Feature
- [ ] Build related changes

`RLP.decode` doesn't check that the result is consistent with the given len prefix.
In other words, given two encoded items `b1` and `b2`, `RLP.decode(b1 + b2)` will actually decode
`b1` and `b2` while it should raise because `b1 + b2` is not a valid RLP (it's a list with missing prefix).

Resolves #1306

## What is the new behavior?

- - -
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot/1307)

### DIFF
--- a/src/utils/eth_transaction.cairo
+++ b/src/utils/eth_transaction.cairo
@@ -28,12 +28,12 @@ namespace EthTransaction {
     ) -> model.EthTransaction* {
         // see https://github.com/ethereum/EIPs/blob/master/EIPS/eip-155.md
         alloc_locals;
-        let (items: RLP.Item*) = alloc();
-        RLP.decode(items, tx_data_len, tx_data);
+        let (tx_items: RLP.Item*) = alloc();
+        RLP.decode(tx_items, tx_data_len, tx_data);
 
-        assert [items].is_list = TRUE;
-        let items_len = [items].data_len;
-        let items = cast([items].data, RLP.Item*);
+        assert [tx_items].is_list = TRUE;
+        let items_len = [tx_items].data_len;
+        let items = cast([tx_items].data, RLP.Item*);
 
         // Pre eip-155 txs have 6 fields, post eip-155 txs have 9 fields
         // We check for both cases here, and do the remaining ones in the next if block
@@ -91,12 +91,12 @@ namespace EthTransaction {
     ) -> model.EthTransaction* {
         alloc_locals;
 
-        let (items: RLP.Item*) = alloc();
-        RLP.decode(items, tx_data_len - 1, tx_data + 1);
+        let (tx_items: RLP.Item*) = alloc();
+        RLP.decode(tx_items, tx_data_len - 1, tx_data + 1);
 
-        assert [items].is_list = TRUE;
-        let items_len = [items].data_len;
-        let items = cast([items].data, RLP.Item*);
+        assert [tx_items].is_list = TRUE;
+        let items_len = [tx_items].data_len;
+        let items = cast([tx_items].data, RLP.Item*);
 
         assert items_len = 8;
         assert items[0].is_list = FALSE;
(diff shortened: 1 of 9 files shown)


src/utils/eth_transaction.cairo
```diff
@@ -31,27 +31,36 @@ namespace EthTransaction {
...
 
-        let nonce = Helpers.bytes_to_felt(sub_items[0].data_len, sub_items[0].data);
-        let gas_price = Helpers.bytes_to_felt(sub_items[1].data_len, sub_items[1].data);
-        let gas_limit = Helpers.bytes_to_felt(sub_items[2].data_len, sub_items[2].data);
+        assert items[0].is_list = FALSE;
+        assert items[1].is_list = FALSE;
+        assert items[2].is_list = FALSE;
+        assert items[3].is_list = FALSE;
+        assert items[4].is_list = FALSE;
+        assert items[5].is_list = FALSE;
```
enitrat, line 43: I would add a comment here so that we don't forget in the future that first we have the generic LegacyTx case (pre-eip155 and post) - here we just verify that the payload is composed of non-list items (and have it clear for external readers)

src/utils/eth_transaction.cairo
```diff
@@ -31,27 +31,36 @@ namespace EthTransaction {
         let (items: RLP.Item*) = alloc();
         RLP.decode(items, tx_data_len, tx_data);
 
-        // the tx is a list of fields, hence first level RLP decoding
-        // is a single item, which is indeed the sought list
         assert [items].is_list = TRUE;
-        let sub_items_len = [items].data_len;
-        let sub_items = cast([items].data, RLP.Item*);
+        let items_len = [items].data_len;
```
enitrat, line 35: I don't like shadowing the previous definition of `items`

tests/src/utils/test_rlp.py
```diff
@@ -1,25 +1,17 @@
...
-            [0, 55, 256],
-        )
-        def test_should_match_decoded_rlp_type_string(self, cairo_run, payload_len):
-            # generate random string of bytes. if payload_len is 0, then generate a single byte inferior to 0x80
-            data = (
-                random.randbytes(payload_len)
-                if payload_len != 0
-                else random.randint(0, 0x80)
-            )
+        @given(data=lists(binary()) | binary())
```
enitrat, line 13: is this generating a random sequence of bytes ?

enitrat (approved):

src/utils/eth_transaction.cairo
ClementWalter: it depends on the context but here the first `items` it not realy the list items, it's the struct containing the items.
I think that it's clearer to read items[0], items[1], etc. when you think about the tx, because it's like in regular languages, where the tx would be `[item, item, item]`. You don't call "item" "sub_item" here.

tests/src/utils/test_rlp.py
ClementWalter: either random bytes or a list of random bytes, ie either `b'1234'` or `[b'1234']`

src/utils/eth_transaction.cairo
enitrat: rename first `items` to `payload` maybe? thus we have payload which is a RLP-list, we decode to a cairo-list of items

enitrat (approved):

PR #1322, opened 2024-08-02
bug: checks memory expansion costs utils

### LINKED ISSUE #1266: bug: no overflow checks in memory expansion costs utils
## Bug Report

Problematic functions in [gas.cairo](../../blob/main/src/kakarot/gas.cairo) are:

### `memory_expansion_cost_saturated`
- [ ] `offset.high + size.high != 0` can be broken when selecting `(P - 1)/2` and `(P + 1)/2`

### `calculate_gas_extend_memory`
- [ ] `max_offset` is a `felt` possibly `> 2**128` so one needs to use `is_le_felt`

### `max_memory_expansion_cost`
```
let expansion_cost = low_expansion.cost + (
    offset_1.high + size_1.high + offset_2.high + size_2.high
) * Gas.MEMORY_COST_U128;
```
- [ ] offsets and sizes can be selected such that the sum is zero while there are not
- [ ] use of `is_le` instead of `is_le_felt`

obatirou:
Time spent on this PR:

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Resolves #1266

- `memory_expansion_cost_saturated`: offset.high + size.high != 0 can be broken when selecting `(P - 1)/2` and `(P + 1)/2`
- `calculate_gas_extend_memory`: max_offset is a felt possibly > 2**128 so one needs to use `is_le_felt`

## What is the new behavior?

- Condition is updated for `memory_expansion_cost_saturated`
- Usage of `is_le_felt` in `calculate_gas_extend_memory`

- - -
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot/1322)

### DIFF
--- a/src/kakarot/gas.cairo
+++ b/src/kakarot/gas.cairo
@@ -73,7 +73,8 @@ namespace Gas {
         return memory_cost;
     }
 
-    // @notice Compute the expansion cost of max_offset for the memory
+    // @notice Compute the expansion cost of max_offset for the memory.
+    // @dev Assumption max_offset < 2**133 necessary for unsigned_div_rem usage.
     // @param words_len The current length of the memory.
     // @param max_offset The target max_offset to be applied to the given memory.
     // @return cost The expansion gas cost: 0 if no expansion is triggered, and the new size of the memory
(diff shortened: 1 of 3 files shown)


src/kakarot/gas.cairo
```diff
@@ -73,15 +73,18 @@ namespace Gas {
         return memory_cost;
     }
 
-    // @notice Compute the expansion cost of max_offset for the memory
+    // @notice Compute the expansion cost of max_offset for the memory.
+    // @dev Assumption max_offset < 2**133 necessary for unsigned_div_rem usage.
```
enitrat, line 77: should this assumption be verified? How are we sure that this case never happens during execution?

enitrat (approved):

PR #1329, opened 2024-08-05
bug: missing overflow checks effective gas price calculation

### LINKED ISSUE #1316: bug: missing overflow checks effective gas price calculation
### Current behavior

Calculation of effective gas price can overflow 

>L275~L303 in src/kakarot/accounts/library.cairo ) can be overflowed, which allows an attacker to submit a transaction with the gas price lower than the base fee.
Let's assume the base fee is 100 (for simplicity). If tx.max_fee_per_gas is 101 and tx.max_priority_fee_per_gas is -100 (so its value is P - 100 ), all checks in the code would be passed (knowing that is_le checks that 0 <= b - a < 2^128 would be helpful to understand the behaviors) and the effective_gas_price would be set to zero, which means the gas price would be zero.

### Expected behavior
Overflow checks are implemented

### Related code
https://github.com/kkrt-labs/kakarot/blob/v0.8.4/src/kakarot/accounts/library.cairo#L275-L305

obatirou:
Time spent on this PR:

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Resolves #1316

## What is the new behavior?

Overflow checks

- - -
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot/1329)

### DIFF
--- a/src/kakarot/accounts/library.cairo
+++ b/src/kakarot/accounts/library.cairo
@@ -9,6 +9,7 @@ from starkware.cairo.common.math import unsigned_div_rem, split_int, split_felt
 from starkware.cairo.common.memcpy import memcpy
 from starkware.cairo.common.uint256 import Uint256, uint256_not, uint256_le
 from starkware.cairo.common.math_cmp import is_nn, is_le_felt
+from starkware.cairo.common.math import assert_le_felt
 from starkware.starknet.common.syscalls import (
     StorageRead,
     StorageWrite,
@@ -235,14 +236,12 @@ namespace AccountContract {
         let (contract_address) = get_contract_address();
         let (balance) = IERC20.balanceOf(native_token_address, contract_address);
 
-        let is_gas_limit_too_high = is_le_felt(2 ** 64, tx.gas_limit);
         with_attr error_message("Gas limit too high") {
-            assert is_gas_limit_too_high = FALSE;
+            assert_le_felt(tx.gas_limit, 2 ** 64 - 1);
         }
 
-        let is_max_fee_per_gas_too_high = is_le_felt(2 ** 128, tx.max_fee_per_gas);
         with_attr error_message("Max fee per gas too high") {
-            assert is_max_fee_per_gas_too_high = FALSE;
+            assert_le_felt(tx.max_fee_per_gas, 2 ** 128 - 1);
         }
 
         let max_gas_fee = tx.gas_limit * tx.max_fee_per_gas;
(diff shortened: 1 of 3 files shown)


src/kakarot/accounts/library.cairo
```diff
@@ -235,6 +235,16 @@ namespace AccountContract {
         let (contract_address) = get_contract_address();
         let (balance) = IERC20.balanceOf(native_token_address, contract_address);
 
+        let is_gas_limit_too_high = is_le_felt(2 ** 64, tx.gas_limit);
+        with_attr error_message("Gas limit too high") {
+            assert is_gas_limit_too_high = FALSE;
+        }
+
+        let is_max_fee_per_gas_too_high = is_le_felt(2 ** 128, tx.max_fee_per_gas);
```
obatirou, line 243: Couldn't this be replace by `[range_check_ptr] = tx.max_fee_per_gas` ?

enitrat: yes, if you want to verify that tx.max_fee_per_gas is in bound [0, 2**128[

src/kakarot/accounts/library.cairo
```diff
@@ -235,6 +235,16 @@ namespace AccountContract {
         let (contract_address) = get_contract_address();
         let (balance) = IERC20.balanceOf(native_token_address, contract_address);
 
+        let is_gas_limit_too_high = is_le_felt(2 ** 64, tx.gas_limit);
+        with_attr error_message("Gas limit too high") {
+            assert is_gas_limit_too_high = FALSE;
```
enitrat, line 240: can be `assert_le_felt` directly

src/kakarot/accounts/library.cairo
```diff
@@ -256,7 +266,9 @@ namespace AccountContract {
             assert enough_fee = TRUE;
         }
 
-        let max_fee_greater_priority_fee = is_nn(tx.max_fee_per_gas - tx.max_priority_fee_per_gas);
+        let max_fee_greater_priority_fee = is_le_felt(
```
enitrat, line 269: can be assert_le_felt directly

src/kakarot/accounts/library.cairo
obatirou: I think it would worth changing the condition as it could be cheaper instead of using `is_le_felt`. I will test it.

obatirou: Was cheaper with direct range_check_ptr usage. Change done with [0cad9e5](https://github.com/kkrt-labs/kakarot/pull/1329/commits/0cad9e55d56d223597643b4405b557bdae9a027c)

Before 4048 after 4030 for the test `test_pass_authorized_pre_eip155_transaction`

Before 
profile001
After
profile002

enitrat (approved):

Issue #1319, opened 2024-07-31
opti: bytes_used_128


obatirou:
## Opti:

[bytes_used_128](https://github.com/kkrt-labs/kakarot/blob/a41fe3b3e2f4d77a81028777e37813775b824e55/src/utils/utils.cairo#L693) could be optimized [see ssj implementation](https://github.com/kkrt-labs/kakarot-ssj/blob/3e5b6c8d3e35b13609b5f7fd216b3e12bcdabb01/crates/utils/src/helpers.cairo#L1258-L1280)

adrianvrj:
Hi, is this open for contributors?
If so, can I get assigned?

obatirou:
> Hi, is this open for contributors? If so, can I get assigned?

Hi, thank you for your interest, issues open to contribution have a specific tag `open for contribution`

obatirou:
I do not think this is necessary to tackle this issue as it will not be a significative improvement

ClementWalter:
Closing this as we don't plan to make core cairo zero utils updates unless proven to be significant bottlenecks

PR #1339, opened 2024-08-09
fix: validate destination address cairo_message

### LINKED ISSUE #1338: bug: parse destination EVM address in cairo_message
>  I am trying to run the e2e tests because I'm suspicious of the usage of Helpers.bytes32_to_felt to parse the EVM destination address in cairo_message.
I think anyone can call that precompile, and the address is taken straight from calldata, so it is not safe to make the assumption it will be a well-formed EVM address. bytes32_to_felt silently overflows if the input does not fit in a felt.

Anyone can call cairo_message from L2 to send a message to L1 (see #1314)
As such, we mustt ensure that the EVM destination address in cairo_message is always well-formed.

obatirou:
Time spent on this PR:

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Resolves #1338

## What is the new behavior?

Destination address for cairo_message is checked to be well formatted

- - -
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot/1339)

### DIFF
--- a/src/kakarot/precompiles/kakarot_precompiles.cairo
+++ b/src/kakarot/precompiles/kakarot_precompiles.cairo
@@ -148,7 +148,12 @@ namespace KakarotPrecompiles {
         // [to_address: address][data_offset: uint256][data_len: uint256][data: bytes[]]
 
         // Load target EVM address
-        let target_address = Helpers.bytes32_to_felt(input);
+        let invalid_address = Helpers.bytes_to_felt(12, input);
+        if (invalid_address != 0) {
+            let (revert_reason_len, revert_reason) = Errors.precompileInputError();
+            return (revert_reason_len, revert_reason, CAIRO_MESSAGE_GAS, TRUE);
+        }
+        let target_address = Helpers.bytes20_to_felt(input + 12);
 
         let data_bytes_len = Helpers.bytes32_to_felt(input + 2 * 32);
         let data_fits_in_input = is_nn(input_len - 3 * 32 - data_bytes_len);
--- a/tests/src/kakarot/precompiles/test_precompiles.py
+++ b/tests/src/kakarot/precompiles/test_precompiles.py
@@ -261,12 +261,20 @@ class TestPrecompiles:
                     [],
                     True,
                 ),
+                (
+                    0x75002,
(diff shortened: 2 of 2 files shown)


src/kakarot/precompiles/kakarot_precompiles.cairo
```diff
@@ -148,7 +148,12 @@ namespace KakarotPrecompiles {
         // [to_address: address][data_offset: uint256][data_len: uint256][data: bytes[]]
 
         // Load target EVM address
-        let target_address = Helpers.bytes32_to_felt(input);
+        let invalid_address = Helpers.bytes_to_felt(12, input);
```
enitrat, line 151: do you think we can use bytes_used and check whether its ==20? if not, it's not a valid address and we revert. Perhaps it would be more efficient, although it's an isolated part of the codebase so the impact would be low

obatirou: So you mean in pseudo code:
```
let evm_address = Helpers.bytes_to_felt(32, input);
bytes_used = Helpers.bytes_used_256(evm_address)
if (bytes_used != 20) {
    <revert> 
}
```
We would need to create a new function `bytes_used_256` as with `bytes_used_128`, it would be necessary to split the felt and call  `bytes_used_128` on low and high which might not be more efficient.

A new issue could be created for `bytes_used_256` as I think it is encountered several times in the code base and could be done on the work that would be done in the optimisation issue of `bytes_used_128` https://github.com/kkrt-labs/kakarot/issues/1319

enitrat: hm, actually we just need to check if evm_address > 2**160 then we're sure it overflows the size of an EVM address, no need to know the actual bytes count.

obatirou: In fact I think it is not possible either: `let evm_address = Helpers.bytes_to_felt(32, input);` will also fail silently in case bytes do not fit in a felt  like the initial implementation. So the suggested alternative solution seems not viable.

enitrat (approved):
LGTM

PR #1356, opened 2024-08-21
Fix underconstrained segments

### LINKED ISSUE #1313: bug: under constrained segments
### Kakarot version

We need to revise in the whole code base places where we allocate segments and then "fill" them with utils like `RLP.decode` or `load_packed_bytes` to make sure that the segments are actually constrained.

We also need to make sure that every `arr` is used with the proper `arr_len` to use the right segment arena

ClementWalter:
Time spent on this PR: 0.5

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Segments are not constrained enough, making it possible for a malicious prover to tweak tx data.

Resolves #1313

## What is the new behavior?

Added more `assert` to make sure that we don't read data beyond `data_len`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot/1356)

### DIFF
--- a/src/utils/rlp.cairo
+++ b/src/utils/rlp.cairo
@@ -1,5 +1,6 @@
 from starkware.cairo.common.bool import FALSE
 from starkware.cairo.common.math_cmp import is_nn
+from starkware.cairo.common.math import assert_not_zero
 from starkware.cairo.common.alloc import alloc
 
 from utils.utils import Helpers
@@ -24,6 +25,10 @@ namespace RLP {
     //  original array and data_size is the size of the data.
     func decode_type{range_check_ptr}(data_len: felt, data: felt*) -> (felt, felt, felt) {
         alloc_locals;
+
+        with_attr error_message("RLP data is empty") {
+            assert_not_zero(data_len);
+        }
 
         let prefix = [data];
 
--- a/src/utils/rlp.cairo
+++ b/src/utils/rlp.cairo
@@ -22,18 +22,15 @@ namespace RLP {
         is_list: felt,
     }
 
-    // Decodes the type of an RLP item.
-    // @dev type=0 means the item is a byte array, type=1 means the item is a list.
-    // @returns (type, data_offset, data_size) where data_offset is the offset of the data in the
-    //  original array and data_size is the size of the data.
-    func decode_type{range_check_ptr}(data_len: felt, data: felt*) -> (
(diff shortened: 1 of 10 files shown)


src/utils/rlp.cairo
```diff
@@ -22,38 +26,50 @@ namespace RLP {
...
         alloc_locals;
 
+        with_attr error_message("RLP data is empty") {
+            assert_not_zero(data_len);
+        }
+
         let prefix = [data];
 
         // Char
         let is_le_127 = is_nn(0x7f - prefix);
```
obatirou, line 41: could this overflow ?
prefix is user input right ? 
It could be > 2 ** 128

tests/src/kakarot/accounts/test_account_contract.py
```diff
@@ -779,3 +779,169 @@ def test_should_pass_all_data_len(self, cairo_run, bytecode):
...
+                        "execute_before": SyscallHandler.block_timestamp + 1,
+                    },
+                    call_array=[
+                        {"to": 0, "selector": 0, "data_offset": 0, "data_len": 0},
+                    ],
+                    calldata=[],
+                    signature=[],
+                )
+
+        @pytest.mark.parametrize("data_len", [0, 2**128])
```
obatirou, line 898: maybe prefer hypothesis to benefit from the ci config

tests/src/utils/test_rlp.py
```diff
@@ -26,11 +26,25 @@ def test_should_match_prefix_reference_implementation(self, cairo_run, data):
 
             assert output == [expected_type, expected_offset, expected_len]
 
+        def test_should_raise_when_data_len_is_zero(self, cairo_run):
+            with cairo_error("RLP data is empty"):
+                cairo_run("test__decode_type", data=[])
+
+        @pytest.mark.parametrize("prefix", [0xB8, 0xF8])
```
obatirou, line 33: Prefer hypothesis 
Should there be a test also for when the prefix is > ff ?

obatirou (approved):

tests/src/utils/test_rlp.py
enitrat: It falls under the prefix >0xf7 so it should be fine

enitrat (approved):

PR #1359, opened 2024-08-22
feat: chain_id mod 53

### LINKED ISSUE #1354: feat: take chain_id module 2**53
## Feature Request

Common tooling breaks when the chain_id is > 2 ** 53. As the chain_id of Kakarot depends on the underlying Starknet chain_id, we don't necessarily control that this chain_id is inferior to 2 ** 53.

Currently the modulus is 2 ** 32. Update it to 2 ** 53.

obatirou:
Time spent on this PR:

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Resolves #1354

## What is the new behavior?

chain_id mod 53

- - -
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot/1359)

### DIFF
--- a/tests/src/kakarot/instructions/test_block_information.py
+++ b/tests/src/kakarot/instructions/test_block_information.py
@@ -1,10 +1,9 @@
-from collections import OrderedDict
 from unittest.mock import patch
 
 import pytest
 
 from kakarot_scripts.constants import COINBASE
-from tests.utils.constants import BIG_CHAIN_ID, BLOCK_GAS_LIMIT, CHAIN_ID, Opcodes
+from tests.utils.constants import BLOCK_GAS_LIMIT, CHAIN_ID, Opcodes
 from tests.utils.syscall_handler import SyscallHandler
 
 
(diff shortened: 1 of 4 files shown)


tests/src/kakarot/instructions/test_block_information.py
```diff
@@ -46,6 +46,6 @@ def test__exec_block_information(self, cairo_run, opcode, expected_result):
             }
         ),
     )
-    def test__exec_chain_id__should_return_mod_64(self, cairo_run):
+    def test__exec_chain_id__should_return_mod_53(self, cairo_run):
         output = cairo_run("test__exec_block_information", opcode=Opcodes.CHAINID)
-        assert output == hex(BIG_CHAIN_ID % 2**32)
+        assert output == hex(BIG_CHAIN_ID % 2**53)
```
ClementWalter, line 51: I think that this test belongs to kakarot unit test, for a `eth_chain_id` test.

enitrat: what do you mean? it should also be testing that the block info is correct imo

ClementWalter: block info tests that the opcode returns the chain id, but the opcode is not responsible for any kind of "special" computation around this value. So the test with `BIG_CHAIN_ID` doesn't belong to the opcode's unit tests

ClementWalter (approved):

PR #1383, opened 2024-09-03
Fix possible overflow in decode_raw

### LINKED ISSUE #1376: bug: missing range_check assertion for RLP.decode
## Bug Report

The `decode_type` method is unsafe. The `decode_raw` method does

```
let (rlp_type, offset, len) = decode_type(data);
local remaining_data_len = data_len - len - offset;
with_attr error_message("RLP data too short for declared length") {
    assert_nn(remaining_data_len);
}
```

This is however not enough as `len` and `offset` can be arbitrarily big.

What do: add the following range_check checks 

```
assert [range_check_ptr] = len;
assert [range_check_ptr + 1] = offset;
assert [range_check_ptr + 2] = len + offset;
```

ClementWalter:
Time spent on this PR: 0.2

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

In `rlp.decode_raw`, the `assert_nn(remaining_data_len);` doeesn't check that `remaining_data_len` has not overflown.

Resolves #1376

## What is the new behavior?

Each of the decoded values are asserted in RANGE_CHECK, as well as their sum.

A test has been added, failing first and passing after fix.

### DIFF
--- a/src/utils/rlp.cairo
+++ b/src/utils/rlp.cairo
@@ -85,6 +85,10 @@ namespace RLP {
         let (rlp_type, offset, len) = decode_type(data);
         local remaining_data_len = data_len - len - offset;
         with_attr error_message("RLP data too short for declared length") {
+            assert [range_check_ptr] = offset;
+            assert [range_check_ptr + 1] = len;
+            assert [range_check_ptr + 2] = offset + len;
+            let range_check_ptr = range_check_ptr + 3;
             assert_nn(remaining_data_len);
         }
 
--- a/tests/src/utils/test_rlp.py
+++ b/tests/src/utils/test_rlp.py
@@ -2,6 +2,7 @@ import pytest
 from hypothesis import given
 from hypothesis.strategies import binary, lists, recursive
 from rlp import codec, decode, encode
+from starkware.cairo.lang.cairo_constants import DEFAULT_PRIME
 
 from tests.utils.constants import TRANSACTIONS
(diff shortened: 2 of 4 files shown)


src/utils/rlp.cairo
```diff
@@ -60,7 +62,11 @@ namespace RLP {
 
         local len_bytes_count = prefix - 0xf7;
         let list_len = Helpers.bytes_to_felt(len_bytes_count, data + 1);
-        return (TYPE_LIST, 1 + len_bytes_count, list_len);
+        tempvar offset = 1 + len_bytes_count;
+        assert [range_check_ptr] = offset;
+        assert [range_check_ptr + 1] = list_len;
+        let range_check_ptr = range_check_ptr + 2;
+        return (TYPE_LIST, offset, list_len);
```
obatirou, line 69: To be sure to understand: `len_bytes_count` can be arbitrary hence `Helpers.bytes_to_felt(len_bytes_count, data + 1);` can overflow silently, hence `list_len` can pass the `range_check`.
But the check in `decode_raw` on the `data_len` allows to be sure the `list_len` is correct. 
If this is the case, `decode_type` should be marked as unsafe (`decode_type_unsafe`) and add a comment that the length of the data needs to be checked.

obatirou: I wonder if bytes_to_felt should be marked as unsafe also

obatirou (approved):

PR #1390, opened 2024-09-04
fix: ensure values < RC_BOUND for charge gas

### LINKED ISSUE #1388: bug: EVM.charge_gas assume values < RC_BOUND
### Kakarot version

After #1323 `EVM.charge_gas` changed from accepting `felt` to assuming that there were `< RC_BOUND`.
However, in some places, this assumption doesn't hold.

### What

The places where we can charge more are places were we compute a dynamic gas based on stack inputs (like CODECOPY, MCOPY, etc.).

We need to early return in them before actually calling the function.

### How

Given the `BLOCK_GAS_LIMIT`, get the maximum memory expansion cost fitting into this limit
- use `from ethereum.shanghai.vm.gas import calculate_memory_gas_cost` in python

Using this MAX_MEMORY_SIZE constant, update the `Gas.memory_expansion_cost_saturated` to return corresponding saturated value if any of `size` or `offset` are above
- keep using first `XXX.high != 0`
- then use `let (q, r) = unsigned_div_rem(XXX.low, MAX_MEMORY_SIZE)` and `if q == 0` early return saturated

When saturating, don't use `EVM.charge_gas` but directly create a reverted EVM

Otherwise, process with computation, knowing that all values < MAX_MEMORY_SIZE, so largely in range_check

obatirou:
Time spent on this PR:

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Resolves #1388

## What is the new behavior?
Ensures values < RC_BOUND for charge gas

- - -
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot/1390)


kakarot_scripts/constants.py
```diff
@@ -19,8 +19,9 @@
 logger.setLevel(logging.INFO)
 load_dotenv()
 
-# Hardcode block gas limit to 7M
+# Hardcode block gas limit to 7M if this is changed, also update the MAX_MEMORY_SIZE in the Kakarot contract
```
ClementWalter, line 22: give more context: max memory size to have memory expansion cost fitting in block gas limit

ClementWalter: Also add that this memory size actually gives `7_000_130` gas, so it's just a little bit more and consequently is safe to use a flag for out of gas

src/kakarot/evm.cairo
```diff
@@ -102,6 +102,19 @@ namespace EVM {
         );
     }
 
+    func out_of_gas(self: model.EVM*, revert_reason_len: felt, revert_reason: felt*) -> model.EVM* {
+        return new model.EVM(
```
ClementWalter, line 106: if it's out of gas, load the revert reason inside and don't take it as an argument

src/kakarot/gas.cairo
```diff
@@ -55,6 +55,7 @@ namespace Gas {
     const TX_ACCESS_LIST_ADDRESS_COST = 2400;
     const TX_ACCESS_LIST_STORAGE_KEY_COST = 1900;
     const BLOBHASH = 3;
+    const MAX_MEMORY_SIZE = 0x1cdbe1;  // for a block gas limit of 7M
```
ClementWalter, line 58: update the `MEMORY_COST_U128` to be the associated memory cost (ie 7000130) and rename

src/kakarot/gas.cairo
```diff
@@ -150,13 +152,19 @@ namespace Gas {
...
 
         let max_offset_1 = (1 - is_zero_1) * (offset_1.low + size_1.low);
         let max_offset_2 = (1 - is_zero_2) * (offset_2.low + size_2.low);
         let max_expansion_is_2 = is_le_felt(max_offset_1, max_offset_2);
         let max_offset = max_offset_1 * (1 - max_expansion_is_2) + max_offset_2 *
             max_expansion_is_2;
+        let (q, _) = unsigned_div_rem(max_offset, MAX_MEMORY_SIZE);
+        tempvar is_saturated = is_not_zero(q);
+        tempvar range_check_ptr = range_check_ptr;
+        jmp expansion_cost_saturated if is_saturated != 0;
```
ClementWalter, line 166: don't need to allocate a `is_saturated`, use `q != 0` in the `jmp` directly

src/kakarot/gas.cairo
```diff
@@ -150,13 +152,19 @@ namespace Gas {
...
         tempvar is_saturated = 1 - is_not_saturated;
+        tempvar range_check_ptr = range_check_ptr;
         jmp expansion_cost_saturated if is_saturated != 0;
 
         let max_offset_1 = (1 - is_zero_1) * (offset_1.low + size_1.low);
         let max_offset_2 = (1 - is_zero_2) * (offset_2.low + size_2.low);
         let max_expansion_is_2 = is_le_felt(max_offset_1, max_offset_2);
         let max_offset = max_offset_1 * (1 - max_expansion_is_2) + max_offset_2 *
             max_expansion_is_2;
+        let (q, _) = unsigned_div_rem(max_offset, MAX_MEMORY_SIZE);
```
ClementWalter, line 163: we need to benchmark `is_le_felt + unsigned_div_rem` usage versus `is_nn + 2x unsigned_div_rem` maybe better to first to the check then use `is_nn`

src/kakarot/gas.cairo
```diff
@@ -50,11 +50,13 @@ namespace Gas {
     const COLD_ACCOUNT_ACCESS = 2600;
     const WARM_ACCESS = 100;
     const INIT_CODE_WORD_COST = 2;
-    const MEMORY_COST_U128 = 0x200000000000000000000000000018000000000000000000000000000000;
     const TX_BASE_COST = 21000;
     const TX_ACCESS_LIST_ADDRESS_COST = 2400;
     const TX_ACCESS_LIST_STORAGE_KEY_COST = 1900;
     const BLOBHASH = 3;
+    const MAX_MEMORY_SIZE_U32 = 0x100000000;
```
ClementWalter, line 57: this is `2**32`, I'd rather use `2**32` directly, enough with `MEMORY_COST_U32`

src/kakarot/gas.cairo
```diff
@@ -50,11 +50,13 @@ namespace Gas {
...
+    const MEMORY_COST_U32 = 0x200018000000;
+    const MAX_WORDS_LEN_U32 = 0x8000000;
```
ClementWalter, line 59: can also use 2**27 I guess

kakarot_scripts/constants.py
```diff
@@ -19,10 +19,11 @@
 logger.setLevel(logging.INFO)
 load_dotenv()
 
-# Hardcode block gas limit to 7M
 BLOCK_GAS_LIMIT = 7_000_000
 DEFAULT_GAS_PRICE = int(1e9)
 BEACON_ROOT_ADDRESS = "0x000F3df6D732807Ef1319fB7B8bB8522d0Beac02"
+MAX_MEMORY_SIZE_U32 = 0x100000000
```
ClementWalter, line 25: see other comment, I don't think it's useful

src/kakarot/instructions/sha3.cairo
```diff
@@ -18,6 +18,7 @@ from kakarot.stack import Stack
 from kakarot.storages import Kakarot_cairo1_helpers_class_hash
 from utils.bytes import bytes_to_bytes8_little_endian
 from utils.maths import unsigned_div_rem
+from kakarot.errors import Errors
```
ClementWalter, line 21:
```suggestion
```

kakarot_scripts/constants.py
```diff
@@ -19,10 +19,11 @@
...
+MEMORY_COST_U32 = 0x200018000000
```
ClementWalter, line 26: use the python reference implementation

```
from ethereum.shanghai.vm.gas import calculate_memory_gas_cost
```

ClementWalter (approved):

PR #1412, opened 2024-09-13
fix(makefile): avoid relinking on makefile

antiyro:
### Key Modifications:
1. **Avoid Always Running `clean` in `build`:** Removed `$(MAKE) clean` from `build`, so it won't delete and relink files every time.
2. **Timestamp Tracking for Artifacts:**
   - The `SSJ_DIR` target only rebuilds if `$(SSJ_ZIP)` changes.
   - `curl` only runs if the zip file doesn't exist.
3. **Directory creation conditional:** The build directory `build/ssj` is only recreated when necessary.

- - -
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot/1412)


ClementWalter (approved):

Makefile
```diff
@@ -4,37 +4,37 @@ ifneq ("$(wildcard .env)","")
...
 
-.PHONY: build test coverage
+.PHONY: build test coverage clean
 
 # 154615699 corresponds to release v0.1.7 of Kakarot SSJ.
 KKRT_SSJ_RELEASE_ID = 154615699
 # Kakarot SSJ artifacts for precompiles.
 KKRT_SSJ_BUILD_ARTIFACT_URL = $(shell curl -L https://api.github.com/repos/kkrt-labs/kakarot-ssj/releases/${KKRT_SSJ_RELEASE_ID} | jq -r '.assets[0].browser_download_url')
-KATANA_VERSION = v1.0.0-alpha.11
+KATANA_VERSION = v1.0.0-alpha.0
```
enitrat, line 13:
```suggestion
KATANA_VERSION = v1.0.0-alpha.11
```

revert

ClementWalter: I think it's on purpose to improve madara relative benchmark

enitrat: either way, we're going to bump this on Monday with the next katana nightly, we need to test the ec precompiles

enitrat (approved):
Thank you for your contribution @antiyro

PR #1494, opened 2024-10-10
feat: Ownable2Step coinbase

### LINKED ISSUE #1488: feat: Upgrade OZ and use Ownable2Step for Coinbase

`Ownable2Step` is safe as it allows to make sure that the new owner will be able to withdraw funds (because it allows to make sure that the new owner is able to trigger a tx to the contract).

### What

- [ ] Update OZ submodule to the latest audited version (V5.0.2 at the time of writing this issue)
- [ ] Make required changes to existing solidity files so that it builds and works (beware of the updated remappings in the OZ lib itself)
- [ ] Use `Ownable2Step` for `Coinbase.sol` and update end to end tests

obatirou:
Time spent on this PR:

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Resolves #1488

## What is the new behavior?

Coinbase use Ownable2Step from OZ with version 5.0.2

- - -
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot/1494)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -13,3 +13,6 @@
 [submodule "solidity_contracts/lib/base64-sol"]
 	path = solidity_contracts/lib/base64-sol
 	url = https://github.com/Brechtpd/base64
+[submodule "solidity_contracts/lib/openzeppelin-contracts"]
+	path = solidity_contracts/lib/openzeppelin-contracts
+	url = https://github.com/OpenZeppelin/openzeppelin-contracts
--- a/.gitmodules
+++ b/.gitmodules
@@ -8,7 +8,7 @@
 	path = solidity_contracts/lib/v3-core
 	url = https://github.com/Uniswap/v3-core
 [submodule "solidity_contracts/lib/openzeppelin"]
-	path = solidity_contracts/lib/openzeppelin
+	path = solidity_contracts/lib/openzeppelin-uniswapV3
 	url = https://github.com/OpenZeppelin/openzeppelin-contracts
 [submodule "solidity_contracts/lib/base64-sol"]
 	path = solidity_contracts/lib/base64-sol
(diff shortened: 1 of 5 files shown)


.gitmodules
```diff
@@ -13,3 +13,6 @@
 [submodule "solidity_contracts/lib/base64-sol"]
 	path = solidity_contracts/lib/base64-sol
 	url = https://github.com/Brechtpd/base64
+[submodule "solidity_contracts/lib/openzeppelin-contracts"]
+	path = solidity_contracts/lib/openzeppelin-contracts
+	url = https://github.com/OpenZeppelin/openzeppelin-contracts
```
ClementWalter, line 18: already un submodules

ClementWalter (changes requested):
just the openzeppeline contracts are already in the submodules

obatirou:
> just the openzeppeline contracts are already in the submodules

Yes there are used in the tests for uniswapv3. The version of the library used for the uniswapv3 is 3.4 which will bring breaking changes leading to code modifications to uniswapv3
My suggestion here is to keep both. 
It is also possible to delete the uniswapv3 test and code as it is skipped

.gitmodules
obatirou: Yes there are used in the tests for uniswapv3. The version of the library used for the uniswapv3 is 3.4 which if upgraded to 5.0 will bring breaking changes leading to code modifications to uniswapv3
My suggestion here is to keep both. 
Other solution would be to delete the uniswapv3 test and code as it is skipped

ClementWalter: or make the submodule name explicit

obatirou: Done 
Should be better

ClementWalter (approved):

PR #1499, opened 2024-10-11
dev(scripts): deploy default DualVM tokens

enitrat:
- Deploys DualVM tokens for all Starkgate-bridged token addresses, mirroring the underlying starknet token.
- Skips deployment if the l2_token_address is the same as Kakarot's native token.

- - -
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot/1499)


kakarot_scripts/deployment/dualvm_token_deployments.py
```diff
@@ -0,0 +1,188 @@
...
+    register_lazy_account,
+    remove_lazy_account,
+)
+
+logging.basicConfig()
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+# %%
+
+
```
ClementWalter, line 35:
```suggestion


# %%
```

kakarot_scripts/deployment/dualvm_token_deployments.py
```diff
@@ -0,0 +1,188 @@
...
+async def deploy_new_token(
+    token_name: str,
+    token: Dict[str, Any],
+    kakarot: str,
+    evm_deployments: Dict[str, Any],
+) -> None:
+    """
+    Deploys a new DualVMToken for a corresponding Starknet ERC20 token.
```
ClementWalter, line 107:
```suggestion
    Deploy a new DualVMToken for a corresponding Starknet ERC20 token.
```

kakarot_scripts/deployment/dualvm_token_deployments.py
```diff
@@ -0,0 +1,188 @@
...
+        return await load_tokens("sepolia")
+    if network["name"] not in ("mainnet", "sepolia"):
+        raise ValueError(f"No known token addresses for network: {network['name']}")
+    return await load_tokens(network["name"])
+
+
+async def load_tokens(network_name: str) -> List[Dict[str, Any]]:
+    file_name = network_name.replace("starknet-", "")
+    file_path = TOKEN_ADDRESSES_DIR / f"{file_name}.json"
+    return json.loads(file_path.read_text())
```
ClementWalter, line 97:
```suggestion
def load_tokens(network_name: str) -> List[Dict[str, Any]]:
    file_name = network_name.replace("starknet-", "")
    file_path = TOKEN_ADDRESSES_DIR / f"{file_name}.json"
    return json.loads(file_path.read_text())
```

kakarot_scripts/deployment/dualvm_token_deployments.py
```diff
@@ -0,0 +1,188 @@
...
+    """
+    if network["type"] == NetworkType.DEV:
+        return await load_tokens("sepolia")
+    if network["name"] not in ("mainnet", "sepolia"):
+        raise ValueError(f"No known token addresses for network: {network['name']}")
+    return await load_tokens(network["name"])
+
+
+async def load_tokens(network_name: str) -> List[Dict[str, Any]]:
+    file_name = network_name.replace("starknet-", "")
```
ClementWalter, line 95: useless given the fact that you renamed the networks

kakarot_scripts/deployment/dualvm_token_deployments.py
```diff
@@ -0,0 +1,188 @@
...
+async def get_tokens(network) -> List[Dict[str, Any]]:
+    """
+    Get the list of tokens for a given network.
+    If in dev mode, will return the sepolia token list.
+    """
+    if network["type"] == NetworkType.DEV:
+        return await load_tokens("sepolia")
+    if network["name"] not in ("mainnet", "sepolia"):
+        raise ValueError(f"No known token addresses for network: {network['name']}")
+    return await load_tokens(network["name"])
```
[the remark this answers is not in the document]
ClementWalter: the check on the existence of the list belongs to the `load_tokens` functions, which is responsible for loading or raising.

kakarot_scripts/deployment/dualvm_token_deployments.py
```diff
@@ -0,0 +1,188 @@
...
+
+
+async def load_tokens(network_name: str) -> List[Dict[str, Any]]:
+    file_name = network_name.replace("starknet-", "")
+    file_path = TOKEN_ADDRESSES_DIR / f"{file_name}.json"
```
ClementWalter, line 96: you can add a here

```
if not file_path.exists():
    # raise ValueError
    # logger.error
```

but you can also just let the `open` crashes

kakarot_scripts/deployment/dualvm_token_deployments.py
```diff
@@ -0,0 +1,188 @@
...
+)
+
+logging.basicConfig()
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+# %%
+
+
+async def deploy_starknet_token(owner_address: str) -> Any:
+    address = await deploy_starknet("StarknetToken", int(1e18), owner_address)
```
ClementWalter, line 37: decimals should come from the token list

kakarot_scripts/deployment/dualvm_token_deployments.py
```diff
@@ -0,0 +1,188 @@
...
+    tokens = await get_tokens(NETWORK)
+
+    for token in tokens:
+        token_name = token["name"]
+        if token_name not in evm_deployments:
+            await deploy_new_token(token_name, token, kakarot, evm_deployments)
+        else:
+            await verify_and_update_existing_token(
+                token_name, token, kakarot, evm_deployments
+            )
```
ClementWalter, line 74:
```suggestion
    for token in tokens:
        token_name = token["name"]
        if token_name not in evm_deployments:
            await deploy_new_token(token_name, token, kakarot, evm_deployments)
        else:
            await verify_and_update_existing_token(
                token_name, token, kakarot, evm_deployments
            )
```

why would you need to extract `token_name` from `token` to eventually give them both to the functions?

kakarot_scripts/deployment/dualvm_token_deployments.py
```diff
@@ -0,0 +1,188 @@
...
+
+    logger.info("Finished processing all tokens")
+    dump_evm_deployments(evm_deployments)
+    logger.info("Updated EVM deployments have been saved")
+    # %%
+
+
+async def get_tokens(network) -> List[Dict[str, Any]]:
```
ClementWalter, line 82:
```suggestion
async def get_tokens_list(network) -> List[Dict[str, Any]]:
```

tests/end_to_end/CairoPrecompiles/test_dual_vm_token.py
```diff
@@ -2,38 +2,24 @@
...
 from tests.utils.errors import cairo_error
 
 
 @pytest_asyncio.fixture(scope="function")
-async def starknet_token(owner):
-    address = await deploy_starknet(
-        "StarknetToken", int(1e18), owner.starknet_contract.address
-    )
-    return get_contract_starknet("StarknetToken", address=address)
+async def starknet_token(dual_vm_token):
```
ClementWalter, line 13: scope="session"

tests/end_to_end/CairoPrecompiles/test_dual_vm_token.py
```diff
@@ -2,38 +2,24 @@
...
-    )
-
-    await invoke(
-        "kakarot",
-        "set_authorized_cairo_precompile_caller",
-        int(dual_vm_token.address, 16),
-        True,
-    )
-    return dual_vm_token
+async def dual_vm_token():
```
ClementWalter, line 19: scope="sessions

tests/end_to_end/CairoPrecompiles/test_dual_vm_token.py
ClementWalter, line 1: please define fixture in order

tests/end_to_end/conftest.py
```diff
@@ -86,6 +86,7 @@ async def _factory(amount=0):
         if balance < tx_cost:
             continue
 
+        breakpoint()
```
obatirou, line 89: should not be here ?

cairo_zero/openzeppelin/ERC20.cairo
```diff
@@ -41,6 +39,13 @@ func totalSupply{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr
     return (totalSupply=totalSupply);
 }
 
+@view
+func total_supply{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr}() -> (
```
obatirou, line 43: Why modifying this file ?
I suppose this is to support the dualVMToken ?
It seems too me the end-to-end tests are using the starknet_token in Cairo folder.
Moreover it corresponds to a specific OZ version v0.6.0 hence any token in this version not having those modification will not be supported by the dualVMToken
Unless I miss something

cairo_zero/openzeppelin/ERC20.cairo
ClementWalter, line 1: Let’s just remove this file and use a Cairo ERC20

kakarot_scripts/deployment/dualvm_token_deployments.py
enitrat: > decimals should come from the token list

There's no notion of decimals in the tokenlist. Plus, we're only deploying these tokens for testing purposes on dev networks. Thus I don't consider keeping 18 default decimals problematic

cairo_zero/openzeppelin/ERC20.cairo
enitrat: This reflects better the new token standard that has dual casing entrypoints

cairo_zero/openzeppelin/ERC20.cairo
enitrat: I think it's used as the default native token for EF-Tests

cairo_zero/openzeppelin/ERC20.cairo
enitrat: Also - we want the `ERC20.cairo` used in tests & ef-tests to have these snake_case entrypoints to reflect real-world behavior.

kakarot_scripts/constants.py
ClementWalter, line 1: could rename `starknet-sepolia-staging` into `staging` only

kakarot_scripts/deployment/dualvm_token_deployments.py
```diff
@@ -0,0 +1,206 @@
...
+
+
+async def deploy_dualvm_tokens() -> None:
+    # %% Deploy DualVM Tokens
+    kakarot = get_starknet_deployments()["kakarot"]
+    kakarot_native_token = (
+        await call_contract("kakarot", "get_native_token")
+    ).native_token_address
+    evm_deployments = get_evm_deployments()
+    tokens = get_tokens_list(NETWORK)
```
ClementWalter, line 72:
```suggestion
    tokens = get_tokens_list()
```

solidity_contracts/src/Kakarot/Coinbase.sol
```diff
@@ -1,23 +1,32 @@
...
-    /// @dev DualVmToken.balanceOf(this) is the same as address(this).balance
+    function receiveEther() external payable {}
+
+    /// @notice Withdraws the native token collected by the contract to an address
+    /// @dev Uses CairoLib to make a StarknetCall to transfer this contract's balance to a starknet address.
     /// @param toStarknetAddress The Starknet address to withdraw to
     function withdraw(uint256 toStarknetAddress) external onlyOwner {
         uint256 balance = address(this).balance;
-        kakarotEth.transfer(toStarknetAddress, balance);
+        uint128 balanceLow = uint128(balance);
```
ClementWalter, line 24: I think that someone reported that solidity is now doing safe casting, meaning that one needs to add `% 2**128`
```suggestion
        uint128 balanceLow = uint128(balance % 2**128);
```

solidity_contracts/src/Kakarot/Coinbase.sol
```diff
@@ -1,23 +1,32 @@
...
-    /// @dev The EVM address of the DualVmToken for Kakarot ETH.
-    DualVmToken public immutable kakarotEth;
+    using CairoLib for uint256;
+
+    uint256 public immutable nativeTokenStarknetAddress;
 
     /// Constructor sets the owner of the contract
-    constructor(address _kakarotEth) Ownable(msg.sender) {
-        kakarotEth = DualVmToken(_kakarotEth);
+    constructor(uint256 _nativeTokenStarknetAddress) Ownable(msg.sender) {
```
obatirou, line 13: The owner will be the déployer of the contract 
Do we want to keep this ?
Will this be the kkrt-labs multi sig or only the déployer of contracts (which I assume are different)

tests/end_to_end/CairoPrecompiles/test_dual_vm_token.py
```diff
@@ -36,6 +44,29 @@ async def dual_vm_token(kakarot, starknet_token, owner):
     return dual_vm_token
 
 
+@pytest_asyncio.fixture(scope="function", autouse=True)
+async def fund_owner(owner, starknet_token, max_fee):
```
ClementWalter, line 48: I dont understand the need for this fixture

tests/end_to_end/CairoPrecompiles/test_dual_vm_token.py
ClementWalter, line 1: tests should use the existing deployed tokens now that they are part of the main deployment flow

solidity_contracts/src/Kakarot/Coinbase.sol
```diff
@@ -1,23 +1,32 @@
...
+        nativeTokenStarknetAddress = _nativeTokenStarknetAddress;
     }
 
-    /// @notice Withdraws ETH from the contract to a Starknet address
-    /// @dev DualVmToken.balanceOf(this) is the same as address(this).balance
+    function receiveEther() external payable {}
```
ClementWalter, line 17: use regular fallback function

solidity_contracts/src/Kakarot/Coinbase.sol
obatirou: After a second look, 
The issue is invalid 
```solidity 
pragma solidity 0.8.27;

import "forge-std/Test.sol";
import "forge-std/console.sol";

contract Ok is Test {
    function test_conversion() public {
        uint256 a = type(uint256).max;
        uint128 b = uint128(a);
        console.log(a);
        console.log(b);
    }
}
```

Logs:
  115792089237316195423570985008687907853269984665640564039457584007913129639935
  340282366920938463463374607431768211455
  
 It compiles and output the value truncaned 
 
 My bad for wrongly doing the verification

tests/end_to_end/CairoPrecompiles/test_dual_vm_token.py
enitrat: > tests should use the existing deployed tokens now that they are part of the main deployment flow

existing tokens lack specific instances required for tests, like total_supply > 2**256

solidity_contracts/src/Kakarot/Coinbase.sol
enitrat: > use regular fallback function

I don't want to spend more time trying to make our web3.py wrapper work with the `receive` special function. I can but at this point i'm out of ideas on how to make it work

tests/end_to_end/CairoPrecompiles/test_dual_vm_token.py
enitrat: this one is required if we use the deployed DualVMTokens in the tests. Can be removed if I keep using test-tokens for the tests instead of "official" tokens.

solidity_contracts/src/Kakarot/Coinbase.sol
enitrat: ok i made it work

kakarot_scripts/deployment/dualvm_token_deployments.py
```diff
@@ -0,0 +1,190 @@
...
+    evm_deployments[token_name] = {
+        "address": int(contract.address, 16),
+        "starknet_address": contract.starknet_address,
+    }
+    logger.info(
+        f"Deployed new DualVMToken for {token_name} at address {contract.address}"
+    )
+
+
+async def check_dualvm_token_deployment(
```
obatirou, line 117:
```suggestion
async def is_dualvm_token_deployed(
```

obatirou: More explicit naming suggested it will return a bool

tests/end_to_end/Kakarot/test_coinbase.py
```diff
@@ -48,6 +41,18 @@ async def test_should_revert_when_not_owner(self, coinbase, other):
             with evm_error(error):
                 await coinbase.withdraw(0xDEAD, caller_eoa=other.starknet_contract)
 
+    class TestReceive:
+        async def test_should_receive_ether(self, coinbase: Web3Contract, owner):
+            amount = 0.001
```
obatirou, line 46: Maybe convert this to a fuzzing test with hypothesis

kakarot_scripts/deployment/dualvm_token_deployments.py
```diff
@@ -0,0 +1,190 @@
...
+    evm_deployments = get_evm_deployments()
+    tokens = get_tokens_list()
+    for token in tokens:
+        if int(token["l2_token_address"], 16) == kakarot_native_token:
+            logger.info(f"Skipping {token['name']} as it is the native token")
+            continue
+        if token["name"] not in evm_deployments:
+            await deploy_new_token(token, kakarot, evm_deployments)
+        else:
+            is_deployed = await check_dualvm_token_deployment(token, evm_deployments)
```
obatirou, line 52:
```suggestion
            is_deployed = await is_dualvm_token_deployed(token, evm_deployments)
```

cairo_zero/openzeppelin/ERC20.cairo
obatirou: I see 
So native tokens that are wrapped by dualVMToken will be like this one ?
As for now dualVMToken has been though of to wrap stargate tokens / standard ones.

This needs to be documented somewhere if this changes to ensure what is compatibles with dualVMToken

tests/end_to_end/Kakarot/test_coinbase.py
enitrat: just checking whether it's possible to send ether basically

cairo_zero/openzeppelin/ERC20.cairo
enitrat: > So native tokens that are wrapped by dualVMToken will be like this one ?

"like this one" as in have both snake_case / camelCase entrypoints - yes
they're a specific version with some more entrypoints related to the bridge logic

> This needs to be documented somewhere if this changes to ensure what is compatibles with dualVMToken

it's as simple as: all tokens adhering to the ERC20 Starknet Token specification are supported

enitrat: more on this here https://docs.openzeppelin.com/contracts-cairo/0.8.0/erc20#:~:text=The%20ERC20%20token%20standard%20is,20%20in%20Cairo%20for%20Starknet.

enitrat:
I had to use explicit `address: int` types in `get_contract` because passing a hexstring, that was not 40-bytes padded would crash the program. It seems that `to_checksum_address` from the library we use does not support hexstrings that are not 40 bytes.

kakarot_scripts/utils/kakarot.py
```diff
@@ -399,21 +406,26 @@ def _get_matching_logs_for_event(event_abi, log_receipts) -> List[dict]:
...
         gas_limit = kwargs.pop("gas_limit", TRANSACTION_GAS_LIMIT)
         value = kwargs.pop("value", 0)
         caller_eoa_ = kwargs.pop("caller_eoa", caller_eoa)
         max_fee = kwargs.pop("max_fee", None)
-        calldata = self.get_function_by_signature(fun)(
-            *args, **kwargs
-        )._encode_transaction_data()
 
-        if abi["stateMutability"] in ["pure", "view"]:
+        if fun is not None:
```
enitrat, line 419: enables sending txs to contracts that are not function invokations, through `contract.w3.eth.send_transaction`

kakarot_scripts/deployment/dualvm_token_deployments.py
ClementWalter, line 1: you don't need to register_lazy because basically deployments only appears for katana. So you'd better `remove_lazy_accout` at the beginning and simplify the flow

kakarot_scripts/deployment/dualvm_token_deployments.py
```diff
@@ -0,0 +1,163 @@
...
+                "address": int(contract.address, 16),
+                "starknet_address": contract.starknet_address,
+            }
+
+    await execute_calls()
+    dump_evm_deployments(evm_deployments)
+    remove_lazy_account(account)
+
+    # Check deployments
+    for token in tokens_to_deploy:
```
ClementWalter, line 128: check for all tokens, not only the one deployed

cairo_zero/openzeppelin/ERC20.cairo
obatirou: OK I see 
So only  ERC20 Starknet token are supported for DualVmTokens for now as they have snake_case and they need to always send back true for all operations. Any changes in futur (or current version) that do not comply with will make them incompatible.

nit: the comment at the beginning of the file should be updated

kakarot_scripts/deployment/dualvm_token_deployments.py
```diff
@@ -0,0 +1,167 @@
...
+async def deploy_dualvm_tokens() -> None:
+    # %% Deploy DualVM Tokens
+
+    # The lazy execution must be done before we check the deployments succeeded, as the l2 contracts
+    # need to be deployed first
+    account = await get_starknet_account()
+
+    # Remove lazy account before we check the deployments succeeded, as the l2 contracts
+    # need to be deployed sequentially.
+    remove_lazy_account(account.address)
```
ClementWalter, line 40:
```suggestion
    # Remove lazy execution as the l2 contracts need to be deployed sequentially.
    account = await get_starknet_account()    
    remove_lazy_account(account.address)
```

but I don't understand why this is true actually, why we need to have the Starknet contract deployed to deploy the corresponding dualvm token

tests/end_to_end/conftest.py
```diff
@@ -86,10 +86,17 @@ async def _factory(amount=0):
...
+            gas_limit=gas_limit,
+            gas_price=gas_price,
+        )
+
+        # Withdraw the funds to the deployer
+        await coinbase.functions["withdraw(uint256)"](
+            toStarknetAddress=deployer.address,
             gas_limit=gas_limit,
             gas_price=gas_price,
         )
```
ClementWalter, line 102: do this once for all outside of the loop

kakarot_scripts/deployment/dualvm_token_deployments.py
enitrat: > but I don't understand why this is true actually, why we need to have the Starknet contract deployed to deploy the corresponding dualvm token

we don't need them to deploy, we need them to to check deployments

if you do

```
        assert await token_contract.name() == token["name"]
```

if the underlying token is not deployed it's going to fail.

we discussed this and you said that it actually didn't matter because the only environment we were deploying starknet tokens in is katana.

ClementWalter (approved):

PR #1501, opened 2024-10-11
Raise if argent status is not TX_ACCEPTED_L2

ClementWalter:
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot/1501)

### DIFF
--- a/kakarot_scripts/deployment/kakarot_deployment.py
+++ b/kakarot_scripts/deployment/kakarot_deployment.py
@@ -69,6 +69,9 @@ async def deploy_or_upgrade_kakarot(owner):
             DEFAULT_GAS_PRICE,
             address=starknet_deployments["kakarot"],
         )
+        # Temporarily set the coinbase to the default EVM deployer so that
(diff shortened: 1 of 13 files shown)


kakarot_scripts/deployment/kakarot_deployment.py
```diff
@@ -68,6 +69,12 @@ async def deploy_or_upgrade_kakarot(owner):
             DEFAULT_GAS_PRICE,
             address=starknet_deployments["kakarot"],
         )
+        await invoke(
```
enitrat, line 72: so coinbase is the default EVM address set in the .env ?

ClementWalter: initially, so that the EVM deployer pays for itself, until coinbase is deployed

obatirou (dismissed):

kakarot_scripts/deployment/kakarot_deployment.py
ClementWalter: added more context about this

obatirou (approved):

PR #1502, opened 2024-10-11
dev: setup script

### LINKED ISSUE #1500: dev: supercharge `make setup`
cc @enitrat for scoping

=> supercharge `make setup` with the following commands:
- `which scarb` to check asdf scarb installation is in path 
- do `foundryup` 
- asdf install

enitrat:
Closes #1500

- - -
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot/1502)

### DIFF
--- a/kakarot_scripts/setup/setup.py
+++ b/kakarot_scripts/setup/setup.py
@@ -1,0 +1,118 @@
+#!/usr/bin/env python3
+
+import os
+import shutil
+import subprocess
+import sys
+
+
+def run_command(command, error_message):
+    try:
+        subprocess.run(command, check=True, shell=True)
+    except subprocess.CalledProcessError:
+        print(f"Error: {error_message}")
+        sys.exit(1)
+
+
+def is_command_available(command):
+    return shutil.which(command) is not None
+
+
+def install_dependency(name, install_command, check_command):
+    if is_command_available(check_command):
+        print(f"{name} is already installed.")
+    else:
+        print(f"Installing {name}...")
+        run_command(install_command, f"Failed to install {name}")
+
+
+def get_shell_config_file():
+    shell = os.environ.get("SHELL", "").split("/")[-1]
+    home = os.path.expanduser("~")
+
+    config_files = {
+        "bash": [".bashrc", ".bash_profile"],
+        "zsh": [".zshrc"],
+        "fish": [".config/fish/config.fish"],
+    }
+
+    for file in config_files.get(shell, []):
+        full_path = os.path.join(home, file)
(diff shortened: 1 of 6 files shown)


enitrat:
@Eikix can you try `make setup` with this and check all runs fine now?

kakarot_scripts/setup/setup.py
```diff
@@ -0,0 +1,164 @@
...
+        "katana",
+        version=KATANA_VERSION,
+    )
+
+
+def setup_local() -> None:
+    if sys.version_info < PYTHON_MIN_VERSION:
+        raise SetupError(
+            f"Python {PYTHON_MIN_VERSION[0]}.{PYTHON_MIN_VERSION[1]} or higher is required."
+        )
```
ClementWalter, line 89: install version with `uv`. This script should be able to run with even the lowest versions

kakarot_scripts/setup/setup.py
```diff
@@ -0,0 +1,164 @@
...
+                f"echo '. $HOME/.asdf/asdf.sh' >> {shell_config}",
+                "Failed to source asdf environment",
+            )
+            print("Please restart your terminal to use asdf.")
+        else:
+            print("Please add asdf to your shell configuration manually.")
+    else:
+        print("asdf is already installed.")
+
+    if is_command_available("asdf"):
```
ClementWalter, line 131:
```suggestion
```

kakarot_scripts/setup/setup.py
ClementWalter, line 1: you don't unstall uv here?

ClementWalter (commented):
I'm not against a simple script for doing this though more ad hoc tools exist. Just bear in mind then that
- the python version requirement for the script should be as low as possible so that almost anyone can run it
- only python should be required to run it and all the tools added therein (missing uv currently if I'm not mistaken)

kakarot_scripts/setup/setup.py
enitrat: uv is needed to run this script. so no

enitrat:
i consider that uv needs to be manually installed to run this script. perhaps its a mistake? but because it's a python script, it needs python to run anyway 🤔

enitrat:
modified the makefile to use raw `python` and installed uv in the script.

kakarot_scripts/setup/setup.py
ClementWalter, line 1: use a logger instead of print

ClementWalter (approved):
logging could be a bit nicer, and submodule updated but tested and it's already nice

PR #1503, opened 2024-10-12
feat: MulticallCairo precompile

### LINKED ISSUE #1464: feat: multicall to cairo precompile
## Feature Request
Create a solidity multicall contract that receive an array of (contract address, selector, calldata) and delegateCallCairo to the the precompile. This will allow to craft directly in the frontend starknet transaction with starknet sdk leading to a generic integration of every starknet contract

Steps: 
* create the solidity multicall contract 
* Add test:
   * craft a starknet transaction with starknet sdk: a simple one would to query the balance of several starknet tokens
    * craft a multicalll transaction to the multicall contract and execute it

enitrat:
Time spent on this PR:

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Resolves #1464

## What is the new behavior?

- Implements the `MulticallCairo` precompile - at address `0x75003`, that receives an flattened array of cairo calls and executes them sequentially.
- Permisionless precompile: anyone can call it, unlike `0x75001`
- ⚠️ **For security purposes this precompile CANNOT be called with DELEGATECALL or CALLCODE** ⚠️ 
- Ignores return data of the Cairo calls
  - Highlights the requirement of a `SinglecallCairo` precompile that would be **permissionless**, and also have **delegatecall deactivated** ?


cairo_zero/kakarot/precompiles/kakarot_precompiles.cairo
```diff
@@ -139,3 +152,161 @@ namespace KakarotPrecompiles {
...
+        Helpers.felt_array_to_bytes32_array(retdata_len, retdata, output);
+        return (output_len, output, CAIRO_PRECOMPILE_GAS, FALSE);
+    }
+
+    // @notice Checks if the selector is a valid EVM selector for a Cairo call.
+    // @dev Currently, two selectors are supported:
+    // call_contract(uint256,uint256,uint256[]) - for "regular" CallCairo Precompile
+    // call_contract(bytes4, uint256,uint256,uint256[]) - for MulticallCairo Precompile
+    // @param selector The selector to check.
+    // @returns TRUE if the selector is valid, FALSE otherwise.
```
enitrat, line 299: It's unclear to me what the _actual_ selector should be. 

The easy way is to take each individual starknet call and create a `bytes memory encodedCall = abi.encode(contractAddress, functionSelector, data);`

and at the end, take

```
        bytes4 selector = bytes4(keccak256("call_contract(bytes4,uint256,uint256,uint256[])"));
        bytes memory callData = abi.encodePacked(selector, bytes4(n_calls));
        callData = concatenateBytes(callData, encodedCall);
```

But it's probably not the most easy way to use this in solidity.



In python (or JS, any frontend that wants to leverage this) on the other hand it's probably the most simple approach:

```
            # 1st starknet call to perform in the batch
            starknet_selector = get_selector_from_name("set_counter")
            data = list(int_to_uint256(prev_count * 2 + 100))
            encoded_starknet_call_1 = encode(
                ["uint256", "uint256", "uint256[]"],
                [starknet_address, starknet_selector, data],
            )

            # 2nd starknet call to perform in the batch
            starknet_selector = get_selector_from_name("inc")
            data = []
            encoded_starknet_call_2 = encode(
                ["uint256", "uint256", "uint256[]"],
                [starknet_address, starknet_selector, data],
            )

            encoded_starknet_calls = encoded_starknet_call_1 + encoded_starknet_call_2
            calls_per_batch = 2

            tx_data = (
                evm_call_selector.hex()
                + f"{calls_per_batch:08x}"
                + encoded_starknet_calls.hex()
            )
  ```
  
  just sign this tx with metamask

File: cairo_zero/kakarot/precompiles/kakarot_precompiles.cairo [hunk 2]
```diff
@@ -78,44 +72,63 @@ namespace KakarotPrecompiles {
...
+        syscall_ptr: felt*,
+        pedersen_ptr: HashBuiltin*,
+        range_check_ptr,
+        bitwise_ptr: BitwiseBuiltin*,
+    }(input_len: felt, input: felt*, caller_address: felt) -> (
+        output_len: felt, output: felt*, gas_used: felt, reverted: felt
+    ) {
+        alloc_locals;
+
+        // Input must be at least 8 bytes long.
```
obatirou, line 101: should it accept multi call with 0 calls ?

cairo_zero/kakarot/precompiles/kakarot_precompiles.cairo
```diff
@@ -78,44 +72,63 @@ namespace KakarotPrecompiles {
...
+            return (
+                revert_reason_len, revert_reason, CAIRO_PRECOMPILE_GAS, Errors.EXCEPTIONAL_HALT
+            );
         }
 
-        let (revert_reason_len, revert_reason) = Errors.invalidCairoSelector();
-        return (revert_reason_len, revert_reason, CAIRO_PRECOMPILE_GAS, TRUE);
+        // Input is formatted as:
+        // [call_selector: bytes4][number_of_calls: bytes4]
+        // [starknet_address_1: bytes32][starknet_selector_1:bytes32][data_offset_1: bytes32][data_len_1: bytes32][data_1: bytes[]]...
```
obatirou, line 112: Same, I suggest to put this above the first if at the beginning

cairo_zero/kakarot/precompiles/kakarot_precompiles.cairo
```diff
@@ -139,3 +152,161 @@ namespace KakarotPrecompiles {
...
+        output_len: felt, output: felt*, reverted: felt
+    ) {
+        alloc_locals;
+
+        if (calls_len == 0) {
+            let (output) = alloc();
+            return (0, output, FALSE);
+        }
+
+        // Ensure that the current remaining calls_len  >= MIN_EVM_ENCODED_STARKNET_CALL_BYTES
```
obatirou, line 178:
```suggestion
        // Ensure that the current remaining calls_len >= MIN_EVM_ENCODED_STARKNET_CALL_BYTES
```

cairo_zero/kakarot/precompiles/kakarot_precompiles.cairo
```diff
@@ -139,3 +152,161 @@ namespace KakarotPrecompiles {
...
+            let (revert_reason_len, revert_reason) = Errors.outOfBoundsRead();
+            return (revert_reason_len, revert_reason, Errors.EXCEPTIONAL_HALT);
+        }
+
+        // Input is formatted as an array of:
+        // [starknet_address: bytes32][starknet_selector:bytes32][data_offset: bytes32][data_len: bytes32][data: bytes[]]]
+
+        // Load address and cairo selector called
+        // Unsafe to assume that the 32 bytes in input do not overflow a felt - responsibility of the caller to format its input correctly
+        let to_starknet_address = Helpers.bytes32_to_felt(calls);
```
obatirou, line 191: transform to Uint256 and compare to Prime to ensure this does not overflow ?

cairo_zero/kakarot/precompiles/kakarot_precompiles.cairo
```diff
@@ -139,3 +152,161 @@ namespace KakarotPrecompiles {
...
+        let starknet_selector_ptr = calls + 32;
+        let starknet_selector = Helpers.bytes32_to_felt(starknet_selector_ptr);
+
+        let data_offset_ptr = calls + 64;
+        let data_offset = Helpers.bytes32_to_felt(data_offset_ptr);
+        let data_len_ptr = calls + data_offset;
+
+        // Load input data by packing all
+        // If the input data is larger than the size of a felt, it will wrap around the felt size.
+        let data_words_len = Helpers.bytes32_to_felt(data_len_ptr);
```
obatirou, line 202: as per zellic fix https://github.com/kkrt-labs/kakarot/pull/1387/files, the overflowing should be checked

cairo_zero/kakarot/precompiles/kakarot_precompiles.cairo
```diff
@@ -139,3 +152,161 @@ namespace KakarotPrecompiles {
...
+    // call_contract(bytes4, uint256,uint256,uint256[]) - for MulticallCairo Precompile
+    // @param selector The selector to check.
+    // @returns TRUE if the selector is valid, FALSE otherwise.
+    func is_evm_selector_valid(selector: felt) -> felt {
+        // call_contract(uint256,uint256,uint256[])
+        if (selector == 0xb3eb2c1b) {
+            return TRUE;
+        }
+
+        // call_contract(bytes4, uint256,uint256,uint256[])
```
obatirou, line 306:
```suggestion
        // call_contract(bytes4,uint256,uint256,uint256[])
```

cairo_zero/kakarot/precompiles/kakarot_precompiles.cairo
```diff
@@ -78,44 +72,63 @@ namespace KakarotPrecompiles {
...
+        pedersen_ptr: HashBuiltin*,
+        range_check_ptr,
+        bitwise_ptr: BitwiseBuiltin*,
+    }(input_len: felt, input: felt*, caller_address: felt) -> (
+        output_len: felt, output: felt*, gas_used: felt, reverted: felt
+    ) {
+        alloc_locals;
+
+        // Input must be at least 8 bytes long.
+        let is_input_invalid = is_nn(8 - input_len);
```
obatirou, line 102: Should it accept multi call with 0 calls ?

cairo_zero/kakarot/precompiles/kakarot_precompiles.cairo
ClementWalter, line 112: isn't this the cairo0 formatting? cairo calls are 
- `to`
- `selector`
- `calldata_len`
- `calldata*`

let's just adopt this convention

cairo_zero/kakarot/precompiles/kakarot_precompiles.cairo
```diff
@@ -78,44 +72,63 @@ namespace KakarotPrecompiles {
...
+            let (revert_reason_len, revert_reason) = Errors.outOfBoundsRead();
+            return (
+                revert_reason_len, revert_reason, CAIRO_PRECOMPILE_GAS, Errors.EXCEPTIONAL_HALT
+            );
         }
 
-        let (revert_reason_len, revert_reason) = Errors.invalidCairoSelector();
-        return (revert_reason_len, revert_reason, CAIRO_PRECOMPILE_GAS, TRUE);
+        // Input is formatted as:
+        // [call_selector: bytes4][number_of_calls: bytes4]
```
ClementWalter, line 111: what is this `call_selector`?

enitrat: EVM call selector

cairo_zero/kakarot/precompiles/precompiles_helpers.cairo
```diff
@@ -33,4 +38,76 @@ namespace PrecompilesHelpers {
...
+
+        // Ensure that calls to CAIRO_BATCH_CALL_PRECOMPILE are not made through a delegatecall / callcode.
+        if (precompile_address == Constants.CAIRO_BATCH_CALL_PRECOMPILE) {
+            let is_not_delegatecall = Helpers.is_zero(message_address - precompile_address);
+            tempvar authorized = authorized * is_not_delegatecall;
+        } else {
+            tempvar authorized = authorized;
+        }
+        return authorized;
+    }
```
ClementWalter, line 112:
```suggestion
    func is_call_authorized{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr}(
        precompile_address: felt,
        caller_code_address: felt,
        caller_address: felt,
        message_address: felt,
    ) -> felt {
        alloc_locals;
        let precompile_requires_whitelist = requires_whitelist(precompile_address);

        // Ensure that calls to precompiles that require a whitelist are properly authorized.
        if (precompile_requires_whitelist == TRUE) {
            return is_caller_whitelisted(caller_code_address);
        }
        
        // Ensure that calls to CAIRO_BATCH_CALL_PRECOMPILE are not made through a delegatecall / callcode.
        return Helpers.is_zero(message_address - precompile_address);
    }
```

solidity_contracts/src/CairoPrecompiles/MulticallCairoCounterCaller.sol
```diff
@@ -0,0 +1,209 @@
...
+    function incrementCairoCounterCallcode() external {
+        cairoCounter.callcodeCairo("inc");
+    }
+}
+
+library MulticallCairoLib {
+    /// @dev The Batch Cairo precompile contract's address.
+    address constant BATCH_CAIRO_PRECOMPILE_ADDRESS = 0x0000000000000000000000000000000000075003;
+
+    function concatenateBytes(bytes memory a, bytes memory b) public pure returns (bytes memory) {
```
obatirou, line 60: After tinkering with [chisel](https://book.getfoundry.sh/chisel/),  you could replace this by `abi.encodePacked(a,b)` or `a.concat(b)`

cairo_zero/kakarot/precompiles/precompiles_helpers.cairo
ClementWalter: not sure to follow this comment

> // Ensure that calls to CAIRO_BATCH_CALL_PRECOMPILE are not made through a delegatecall / callcode.

when a solidity contract makes a CALL to 0x75003, then `message_address == precompile_address` but we want to allow only for external calls (`depth = 0`)

solidity_contracts/src/CairoPrecompiles/MulticallCairoCounterCaller.sol
```diff
@@ -0,0 +1,209 @@
...
+        returns (bytes memory)
+    {
+        uint256 functionSelector = uint256(keccak256(bytes(functionName))) % 2 ** 250;
+        return callCairo(contractAddress, functionSelector, data);
+    }
+
+    function callCairo(uint256 contractAddress, string memory functionName) internal returns (bytes memory) {
+        uint256[] memory data = new uint256[](0);
+        return callCairo(contractAddress, functionName, data);
+    }
```
obatirou, line 128: are those necessary ?
Should it be more restrictive ?
As it increase the code size hence attack surface 

Same for delegate / callcode

solidity_contracts/src/CairoPrecompiles/MulticallCairoCounterCaller.sol
```diff
@@ -0,0 +1,209 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.0 <0.9.0;
+
+using MulticallCairoLib for uint256;
```
obatirou, line 4: Should put it in the contract as the lib is defined in the same file

solidity_contracts/src/CairoPrecompiles/MulticallCairoCounterCaller.sol
```diff
@@ -0,0 +1,209 @@
...
+        return delegatecallCairo(contractAddress, functionSelector, data);
+    }
+
+    function delegatecallCairo(uint256 contractAddress, string memory functionName) internal returns (bytes memory) {
+        uint256[] memory data = new uint256[](0);
+        uint256 functionSelector = uint256(keccak256(bytes(functionName))) % 2 ** 250;
+        return delegatecallCairo(contractAddress, functionSelector, data);
+    }
+
+    function callcodeCairo(uint256 contractAddress, uint256 functionSelector, uint256[] memory data)
```
obatirou, line 167: Should callcode be authorized ?
It seems not necessary as it is advised to not use it, I suppose it is only for testing ?

tests/end_to_end/CairoPrecompiles/test_batch_cairo_precompile.py
```diff
@@ -0,0 +1,171 @@
...
+    )
+    return caller_contract
+
+
+@pytest.mark.asyncio(scope="module")
+@pytest.mark.CairoPrecompiles
+class TestCairoPrecompiles:
+    class TestCounterPrecompiles:
+        @given(
+            calls_per_batch=st.integers(min_value=0, max_value=100),
```
ClementWalter, line 41:
```suggestion
            calls_per_batch=st.integers(min_value=0, max_value=100)
```

tests/end_to_end/CairoPrecompiles/test_batch_cairo_precompile.py
```diff
@@ -0,0 +1,171 @@
...
+            ]
+
+            # Starknet call to perform in the batch
+            starknet_address = cairo_counter.address
+            starknet_selector = get_selector_from_name("inc")
+            data = []
+            encoded_starknet_call = encode(
+                ["uint256", "uint256", "uint256[]"],
+                [starknet_address, starknet_selector, data],
+            )
```
ClementWalter, line 60: use higher level functions to be closer to the real target usage

```
call = cairo_counter.functions["inc"].prepare()
```

cairo_zero/kakarot/precompiles/precompiles_helpers.cairo
enitrat: Why restricting to external calls?

tests/end_to_end/CairoPrecompiles/test_batch_cairo_precompile.py
```diff
@@ -0,0 +1,171 @@
...
+
+            tx_data = (
+                evm_call_selector.hex()
+                + f"{calls_per_batch:08x}"
+                + encoded_starknet_call.hex() * calls_per_batch
+            )
```
ClementWalter, line 66:
```
calls = [call] * calls_per_batch
tx_data = account._prepare_invoke_v3(calls)
```

tests/end_to_end/CairoPrecompiles/test_batch_cairo_precompile.py
```diff
@@ -0,0 +1,171 @@
...
+            await eth_send_transaction(
+                to=f"0x{0x75003:040x}",
+                gas=21000
+                + 20000
+                * calls_per_batch,  # Gas is 21k base + 10k per call + calldata cost
+                data=tx_data,
+                value=0,
+                caller_eoa=owner.starknet_contract,
```
ClementWalter, line 74: does this change something to the test? if not, leave default value

tests/end_to_end/CairoPrecompiles/test_batch_cairo_precompile.py
```diff
@@ -0,0 +1,171 @@
...
+            )
+
+            new_count = (await cairo_counter.functions["get"].call()).count
+            expected_increment = calls_per_batch
+            assert (
+                new_count == prev_count + expected_increment
+            ), f"Expected count to increase by {expected_increment}, but it increased by {new_count - prev_count}"
```
ClementWalter, line 81:
```suggestion
            assert new_count == prev_count + calls_per_batch
```

cairo_zero/kakarot/precompiles/kakarot_precompiles.cairo
```diff
@@ -139,3 +152,161 @@ namespace KakarotPrecompiles {
...
+
+        let output_len = retdata_len * 32;
+        Helpers.felt_array_to_bytes32_array(retdata_len, retdata, output);
+        return (output_len, output, CAIRO_PRECOMPILE_GAS, FALSE);
+    }
+
+    // @notice Checks if the selector is a valid EVM selector for a Cairo call.
+    // @dev Currently, two selectors are supported:
+    // call_contract(uint256,uint256,uint256[]) - for "regular" CallCairo Precompile
+    // call_contract(bytes4, uint256,uint256,uint256[]) - for MulticallCairo Precompile
```
obatirou, line 297: why `bytes4` as first argument for multicall ?
From my understanding, this is a "virtual selector": it matches the param types for the cairo call. So for the Multicall shouldn't it be `call_contract(uint256,uint256,uint256,uint256[])` ?

solidity_contracts/src/CairoPrecompiles/MulticallCairoCounterCaller.sol
enitrat: Yes, this entire file is just for testing purposes, as indicated in line 105

solidity_contracts/src/CairoPrecompiles/MulticallCairoCounterCaller.sol
enitrat: it's a copy/paste of the CairoLib, for testing purposes only.

solidity_contracts/src/CairoPrecompiles/MulticallCairoCounterCaller.sol
enitrat: bytes.concat(a,b) is the right syntax, I couldn't get it to work, thanks

cairo_zero/kakarot/precompiles/kakarot_precompiles.cairo
enitrat: bytes4 is the amount of calls performed, encoded over 4 bytes (u32)

the expected format is[number_of_calls, call_1, call_2, ... call_n]. Still open to discussions on what the format should be

tests/end_to_end/CairoPrecompiles/test_batch_cairo_precompile.py
enitrat: > calls = [call] * calls_per_batch
> tx_data = account._prepare_invoke_v3(calls)


This is not an EVM call?

cairo_zero/kakarot/precompiles/kakarot_precompiles.cairo
obatirou: So it is to limit the number of calls ?
Why not use uint256 for `number_of_calls` ?
I think it would make more sense from a solidity perspective

cairo_zero/kakarot/precompiles/precompiles_helpers.cairo
enitrat: This should be a generic Multicall contract that has delegatecall/callcode disabled.

cairo_zero/kakarot/precompiles/kakarot_precompiles.cairo
enitrat: this is the EVM encoding of a cairo call

cairo_zero/kakarot/precompiles/kakarot_precompiles.cairo
enitrat: bytes32_to_uint256 -> ensure not > PRIME -> reconstruct felt from (low, high)? 

idk if we should put guard rails around this. @ClementWalter thoughts?

cairo_zero/kakarot/precompiles/kakarot_precompiles.cairo
enitrat: I thought it would be better to have a smaller input considering that we never this will never overflow 4 bytes

solidity_contracts/src/CairoPrecompiles/MulticallCairoCounterCaller.sol
obatirou: maybe this should also apply to the cairoLib ?

cairo_zero/kakarot/precompiles/kakarot_precompiles.cairo
obatirou: I see but you force the dev to format his input. Also not sure sure gas wise the impact of the formating

enitrat: ok, then we can use a u256 but check only the last 4 bytes as done for data_len

cairo_zero/kakarot/precompiles/kakarot_precompiles.cairo
enitrat: done

cairo_zero/kakarot/precompiles/kakarot_precompiles.cairo
enitrat: renamed

cairo_zero/kakarot/precompiles/precompiles_helpers.cairo
enitrat: I will keep my version as it's 'safer' for very little overhead. there could be in the future a combination of whitelist but can't delegatecall.

cairo_zero/kakarot/precompiles/kakarot_precompiles.cairo
```diff
@@ -46,76 +49,87 @@ namespace KakarotPrecompiles {
...
+            calldata_len,
+            calldata,
+            skip_returndata=FALSE,
+        );
+    }
 
-        let data_offset_ptr = args_ptr + 64;
-        let data_offset = Helpers.bytes32_to_felt(data_offset_ptr);
-        let data_len_ptr = args_ptr + data_offset;
+    // @notice Executes a batch of call to cairo contracts.
```
ClementWalter, line 82:
```suggestion
    // @notice Executes a batch of calls to cairo contracts.
```

cairo_zero/kakarot/precompiles/kakarot_precompiles.cairo
```diff
@@ -46,76 +49,87 @@ namespace KakarotPrecompiles {
...
-            }
-            let syscall_ptr = cast([ap - 4], felt*);
-            let pedersen_ptr = cast([ap - 3], HashBuiltin*);
-            let range_check_ptr = [ap - 2];
-            let caller_starknet_address = [ap - 1];
-
-            let (retdata_len, retdata, success) = IAccount.execute_starknet_call(
-                caller_starknet_address, to_starknet_address, starknet_selector, data_len, data
+        // Input must be at least 8 bytes long.
+        let is_input_invalid = is_nn((EVM_SELECTOR_BYTES + NUMBER_OF_CALLS_BYTES) - input_len);
```
ClementWalter, line 101:
```suggestion
        let is_input_invalid = is_nn((EVM_SELECTOR_BYTES + NUMBER_OF_CALLS_BYTES) - input_len - 1);
```

cairo_zero/kakarot/precompiles/kakarot_precompiles.cairo
```diff
@@ -139,3 +153,193 @@ namespace KakarotPrecompiles {
...
+        alloc_locals;
+
+        let to_address = Helpers.bytes32_to_felt(evm_encoded_call_ptr);
+
+        let selector_ptr = evm_encoded_call_ptr + 32;
+        let selector = Helpers.bytes32_to_felt(selector_ptr);
+
+        let data_offset_ptr = evm_encoded_call_ptr + 64;
+        let data_offset = Helpers.bytes32_to_felt(data_offset_ptr);
+        let data_len_ptr = evm_encoded_call_ptr + data_offset;
```
ClementWalter, line 325: you should check that `call_ptr + data_offset < call_len`

cairo_zero/kakarot/precompiles/kakarot_precompiles.cairo
```diff
@@ -139,3 +153,193 @@ namespace KakarotPrecompiles {
...
+
+        // TODO: don't throw cairo vm errors, just fail the call.
+        with_attr error_message("Invalid data length") {
+            assert invalid_data_words_len = 0;
+        }
+
+        let data_words_len = Helpers.bytes4_to_felt(data_len_ptr + 28);
+        let data_bytes_len = data_words_len * 32;
+        let data_ptr = data_len_ptr + 32;
+        let (calldata_len, calldata) = Helpers.load_256_bits_array(data_bytes_len, data_ptr);
```
ClementWalter, line 339: assert that there is enough data in `data_ptr`

solidity_contracts/src/CairoPrecompiles/MulticallCairoCounterCaller.sol
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+    /// @notice Calls the Cairo contract to increment its internal counter in a batch of multiple calls
+    function incrementCairoCounterBatch(uint32 n_calls) external {
+        uint256[][] memory calls = new uint256[][](n_calls);
+        for (uint32 i = 0; i < n_calls; i++) {
+            calls[i] = new uint256[](3);
+            calls[i][0] = cairoCounter;
+            calls[i][1] = FUNCTION_SELECTOR_INC;
+            calls[i][2] = 0; // Length of the empty calldata array
```
ClementWalter, line 34: can't we create a solidity `struct` representing a cairo call (and consequently have this `0` automatically put by the compiler)

cairo_zero/kakarot/precompiles/kakarot_precompiles.cairo
```diff
@@ -139,3 +153,193 @@ namespace KakarotPrecompiles {
...
+        return (output_len, output, CAIRO_PRECOMPILE_GAS, FALSE);
+    }
+
+    // @notice Checks if the selector is a valid EVM selector for a Cairo call.
+    // @dev Currently, two selectors are supported:
+    // call_contract(uint256,uint256,uint256[]) - for "regular" CallCairo Precompile
+    // call_contract(uint256,uint256,uint256,uint256[]) - for MulticallCairo Precompile
+    // @param selector The selector to check.
+    // @returns TRUE if the selector is valid, FALSE otherwise.
+    func is_evm_selector_valid(selector: felt) -> felt {
```
ClementWalter, line 283: `evm_selector` is a legacy from the time when we did both `contract` and `library` calls at the same precompile address, it's no longer required

tests/end_to_end/CairoPrecompiles/test_batch_cairo_precompile.py
```diff
@@ -0,0 +1,166 @@
...
+            self, cairo_counter, owner, calls_per_batch
+        ):
+            prev_count = (await cairo_counter.functions["get"].call()).count
+
+            # Starknet call to perform in the batch
+            call = cairo_counter.functions["inc"].prepare_call()
+            encoded_starknet_call = encode(
+                ["uint256", "uint256", "uint256[]"],
+                [call.to_addr, call.selector, call.calldata],
+            )
```
ClementWalter, line 54:
```
# standard starknet multicall processing
call = cairo_counter.functions["inc"].prepare_call()
calldata = account._prepare_invoke([call] * calls_per_batch).calldata
# split each felt into 32 bytes values
data = flatten(list[int.to_bytes(felt, 32, "big")) for felt in evm_data])
await eth_send_transaction(to=0x75003, data=data, ...)
```

cairo_zero/kakarot/precompiles/kakarot_precompiles.cairo
```diff
@@ -46,76 +48,81 @@ namespace KakarotPrecompiles {
     ) {
         alloc_locals;
 
-        // Input must be at least 4 + 3*32 bytes long.
-        let is_input_invalid = is_nn(99 - input_len);
+        // Input must be at least MIN_EVM_ENCODED_STARKNET_CALL_BYTES bytes long to be valid.
+        let is_call_invalid = is_nn(MIN_EVM_ENCODED_STARKNET_CALL_BYTES - (input_len + 1));
```
ClementWalter, line 52: maybe add a comment about the fact that because of gas cost, `input_len < 2**128`

cairo_zero/kakarot/precompiles/kakarot_precompiles.cairo
```diff
@@ -46,76 +48,81 @@ namespace KakarotPrecompiles {
...
+            );
+        }
+
+        return Internals.execute_cairo_call(
+            caller_address, to_address, selector, calldata_len, calldata, skip_returndata=FALSE
+        );
+    }
+
+    // @notice Executes a batch of calls to cairo contracts.
+    // @dev Cannot be called with CALLCODE / DELEGATECALL - _should_ be checked upstream.
```
ClementWalter, line 77:
```suggestion
    // @dev Cannot be called with CALLCODE / DELEGATECALL - _must_ be checked upstream.
```

tests/end_to_end/CairoPrecompiles/test_batch_cairo_precompile.py
```diff
@@ -0,0 +1,147 @@
...
+)[:4]
+
+
+@pytest_asyncio.fixture(scope="module")
+async def cairo_counter(max_fee, deployer):
+    cairo_counter = get_contract("Counter", provider=deployer)
+
+    yield cairo_counter
+
+    tx = await cairo_counter.functions["set_counter"].invoke_v1(0, max_fee=max_fee)
```
ClementWalter, line 25: it's better to use our wrapper `call` and `invoke` as starknet py is not so stable, and so it's easier when we need to update/patch

tests/end_to_end/CairoPrecompiles/test_batch_cairo_precompile.py
```diff
@@ -0,0 +1,147 @@
...
+    await wait_for_transaction(tx.hash)
+
+
+@pytest_asyncio.fixture(scope="module")
+async def multicall_cairo_counter_caller(owner, cairo_counter):
+    caller_contract = await deploy(
+        "CairoPrecompiles",
+        "MulticallCairoCounterCaller",
+        cairo_counter.address,
+        caller_eoa=owner.starknet_contract,
```
ClementWalter, line 35: remove if `msg.sender` is not used in `constructor`

tests/end_to_end/CairoPrecompiles/test_batch_cairo_precompile.py
```diff
@@ -0,0 +1,147 @@
...
+        @given(calls_per_batch=st.integers(min_value=0, max_value=100))
+        @settings(max_examples=5)
+        async def test_should_increase_counter_in_batches(
+            self, cairo_counter, owner, calls_per_batch
+        ):
+            prev_count = (await cairo_counter.functions["get"].call()).count
+
+            calls = [
+                (cairo_counter.functions["inc"], []) for _ in range(calls_per_batch)
+            ]
```
ClementWalter, line 68:
```suggestion
            calls = [
                cairo_counter.functions["inc"].prepare_call() for _ in range(calls_per_batch)
            ]
```

tests/end_to_end/CairoPrecompiles/test_batch_cairo_precompile.py
```diff
@@ -0,0 +1,147 @@
...
+        ["uint256", "uint256", "uint256[]"],
+        [call.to_addr, call.selector, call.calldata],
+    )
+
+
+def prepare_transaction_data(calls: List[Tuple[Any, str, List[Any]]]) -> str:
+    encoded_calls = b"".join(
+        encode_starknet_call(entrypoint.prepare_call(*calldata))
+        for entrypoint, calldata in calls
+    )
```
ClementWalter, line 51:
```suggestion
    encoded_calls = b"".join(
        encode_starknet_call(call)
        for call in calls
    )
```

tests/end_to_end/CairoPrecompiles/test_batch_cairo_precompile.py
```diff
@@ -0,0 +1,147 @@
...
+    caller_contract = await deploy(
+        "CairoPrecompiles",
+        "MulticallCairoCounterCaller",
+        cairo_counter.address,
+        caller_eoa=owner.starknet_contract,
+    )
+    return caller_contract
+
+
+def encode_starknet_call(call) -> bytes:
```
ClementWalter, line 40: looks like it's only used once so I'd inline it directly in the `prepare_transaction_data`

tests/end_to_end/CairoPrecompiles/test_batch_cairo_precompile.py
```diff
@@ -0,0 +1,147 @@
...
+                to=f"0x{0x75003:040x}",
+                gas=21000 + 20000 * calls_per_batch,
+                data=tx_data,
+                value=0,
+            )
+
+            new_count = (await cairo_counter.functions["get"].call()).count
+            assert (
+                new_count == prev_count + calls_per_batch
+            ), f"Expected count to increase by {calls_per_batch}, but it increased by {new_count - prev_count}"
```
ClementWalter, line 81:
```suggestion
            assert new_count == prev_count + calls_per_batch
```

ClementWalter: assertion error message in test don't really make sense as the test runner will print the assert error with values

tests/end_to_end/CairoPrecompiles/test_batch_cairo_precompile.py
enitrat: call and invoke are good for contracts available in artifacts, not so much when passing an address because the calldata  is not parsed.

tests/end_to_end/CairoPrecompiles/test_batch_cairo_precompile.py
enitrat: not sure I follow this. why don't you prepare the call?

tests/end_to_end/CairoPrecompiles/test_batch_cairo_precompile.py
ClementWalter: if `get_contract` gives you the right contract without providing `address=`, then it means that `call` and `invoke` will target the right contract by default

enitrat: ah yes ok i got it now!

ClementWalter (approved):

PR #1514, opened 2024-10-21
bump rust to 1.82

tcoratger:
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot/1514)


enitrat:
@tcoratger im not sure it works fine with scarb

tcoratger:
@enitrat I see nothing in the CI, anyway this is not blocker but should be since to have since the next release of reth will be based on this rust version and it is always better to have same rust version with our rpc.

enitrat (approved):

PR #1554, opened 2024-10-29
Update BenchmarkCairoCalls.sol

famouswizard:
Снимок экрана 2024-10-29 в 14 33 35

The parameter `n_felt_input` and variable `i` are now of type `uint256` to avoid type mismatch errors when working with arrays in Solidity.

Time spent on this PR:

5 mins

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Resolves #<Issue number>

## What is the new behavior?

-
-
-

- - -
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot/1554)

### DIFF
--- a/solidity_contracts/src/CairoPrecompiles/BenchmarkCairoCalls.sol
+++ b/solidity_contracts/src/CairoPrecompiles/BenchmarkCairoCalls.sol
@@ -21,16 +21,16 @@ contract BenchmarkCairoCalls {
 
     function empty() external {}
 
-    function callCairoWithFeltInputs(uint32 n_felt_input) external {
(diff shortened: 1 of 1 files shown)


enitrat:
what is the issue?

famouswizard:
> The parameter `n_felt_input` and variable `i` are now of type `uint256` to avoid type mismatch errors when working with arrays in Solidity.

I described the problem and made changes.

enitrat:
I don't understand why changing to u256 when we always fit a smaller type

ClementWalter:
there is no need to go higher than u32 for this input. At the bytecode level it's the same. I don't understand the issue.

PR #1555, opened 2024-10-29
Update setup.py

famouswizard:
Undefined `logger` in `install_dependency` function: The `logger` object isn’t defined globally at the time `install_dependency` is called, so logging statements will cause a `NameError`. Moving the logger setup above this function should fix this.

Unused `version` parameter in `run_command:` The `version` argument is passed to `run_command` but isn’t used in that function, which can cause unexpected behavior.

### Summary of Changes

Moved the logging configuration setup to the beginning of the script, ensuring `logger` is defined before it is used in `install_dependency`.

Removed the `version` parameter from the `run_command` call in i`nstall_dependency`, as it was unnecessary.

Time spent on this PR:

19 mins

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Resolves #<Issue number>

## What is the new behavior?

-
-
-

- - -
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot/1555)

### DIFF
--- a/kakarot_scripts/setup/setup.py
+++ b/kakarot_scripts/setup/setup.py
@@ -16,10 +16,13 @@ SHELL_CONFIG_FILES = {
     "fish": [".config/fish/config.fish"],
 }
 
+# Set up logging before using logger
+logging.basicConfig(level=logging.INFO, format="%(message)s")
+logger = logging.getLogger(__name__)
+
 
 class SetupError(Exception):
     """Custom exception for setup errors."""
-
     pass
 
 
@@ -112,8 +115,7 @@ def setup_local() -> None:
         logger.info("Installing asdf...")
         run_command(
             f"git clone https://github.com/asdf-vm/asdf.git ~/.asdf --branch {ASDF_VERSION}",
-            "Failed to install asdf",
-            version=ASDF_VERSION,
+            "Failed to install asdf"
         )
         shell_config = get_shell_config_file()
         if shell_config:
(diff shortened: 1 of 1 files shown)


kakarot_scripts/setup/setup.py
```diff
@@ -112,8 +115,7 @@ def setup_local() -> None:
         logger.info("Installing asdf...")
         run_command(
             f"git clone https://github.com/asdf-vm/asdf.git ~/.asdf --branch {ASDF_VERSION}",
-            "Failed to install asdf",
-            version=ASDF_VERSION,
```
enitrat, line 116: we should set the version to `v0.14.1`

enitrat:
thank you! can you add ASDF_VERSION ?

famouswizard:
> thank you! can you add ASDF_VERSION ?

Good catch! To incorporate `ASDF_VERSION` during the installation or update of asdf, the script needs to use this version variable properly when installing `asdf`. Here’s the modified part of the `setup_local` function, incorporating `ASDF_VERSION` where relevant.

I made changes to my fork.

enitrat:
Nothing changed

famouswizard:
> Nothing changed

Check my fork

enitrat:
you have to push to the same branch as the PR

ClementWalter:
@famouswizard do you plan to fix your PR or should I close it?

PR #1566, opened 2024-11-04
[KGA-82] [KGA-80] fix: codehash commitment

enitrat:
Fixes the codehash assigned to an account upon deployment as per https://github.com/code-423n4/2024-09-kakarot-findings/issues/10

- In any case, the codehash of an account is keccak(code_bytes)
- IF that account has no code, nonce nor balance, then `extcodehash` opcode returns 0.

As such, upon deploying the starknet contract corresponding to an account, we should _always_ set the codehash to `self.code_hash`

- - -
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot/1566)

### DIFF
--- a/cairo_zero/backend/starknet.cairo
+++ b/cairo_zero/backend/starknet.cairo
@@ -177,6 +177,10 @@ namespace Internals {
         if (starknet_account_exists == 0) {
             // Deploy account
             Starknet.deploy(self.address.evm);
+            // Commit the code hash upon deployment
+            // of a new account, in all cases.
+            // Retrieved in `fetch_or_create` in the next transaction.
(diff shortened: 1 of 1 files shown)


ClementWalter (changes requested):
remove the `set_code_hash` call in `if (self.created != FALSE) {` then

enitrat:
> remove the set_code_hash call in if (self.created != FALSE) { then

I think you should always set the codehash when writing the bytecode. Perhaps there are instances where the starknet contract is already deployed and the code is re-written: re-deployed selfdestructed contracts.

ClementWalter (approved):

PR #1567, opened 2024-11-04
[KGA-29] [KGA-136] fix: cases with Cairo precompiles and delegatecalls

### LINKED ISSUE #1562: bug: Kakarot precompiles can be abused by malicious contracts by delegate-calling whitelisted contracts
## Bug Report
Malicious contracts can exploit users to make arbitrary calls to whitelisted contracts on their behalf via delegatecall
https://github.com/code-423n4/2024-09-kakarot-findings/issues/38

## Fix to implement
Make sure that DualVmToken, L2KakarotMessaging, as well as any other future contract using Kakarot precompiles, make extensive use of [noDelegateCall modifiers](https://github.com/Uniswap/v3-core/blob/main/contracts/NoDelegateCall.sol).

enitrat:
Fixes the issue raised during the audit about eventual honeypots that could lure a user and manipulate starknet tokens without explicit approvals

1. Remove dependence on the `code_address` for whitelisting, which were reported dangerous.
2. Disable ability of having nesting a delegatecall to 0x75001 at protocol level. (no longer possible to do User A -call-> contract C -delegatecall->DualVmToken -delegatecall->0x75001 as a byproduct of 1.
3. Disable ability of calling 0x75002 with a DELEGATECALL at protocol level
4. Added noDelegateCall modifier to L2KakarotMessaging and DualVMToken for extra security
4. Added associated tests under `Security/`

https://github.com/code-423n4/2024-09-kakarot-findings/issues/38

Closes #1562

- - -
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot/1567)


solidity_contracts/src/Security/NoDelegateCall.sol
```diff
@@ -0,0 +1,28 @@
+// SPDX-License-Identifier: BUSL-1.1
```
obatirou, line 1: uniswapV3 should be imported as a dependency instead of copy paste

obatirou: It is not possible to import it from the uniswapV3 repo as the version is fixed to 0.8.12 
https://github.com/Uniswap/v3-core/blob/0.8/contracts/NoDelegateCall.sol

So let leave it this way

tests/end_to_end/CairoPrecompiles/test_call_cairo_precompile.py
```diff
@@ -101,9 +101,7 @@ async def test_should_fail_when_called_with_delegatecall(
         async def test_should_fail_when_called_with_callcode(
             self, cairo_counter_caller
         ):
-            with cairo_error(
-                "EVM tx reverted, reverting SN tx because of previous calls to cairo precompiles"
-            ):
+            with evm_error("CairoLib: call_contract failed with"):
```
obatirou, line 104: This cannot stay like this. I do not understand the entire error 
```
b'\x08\xc3y\xa0\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00 \x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00%CairoLib: call_contract failed with: \x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00'
```

@enitrat if you have inputs

obatirou: It is the error from the contract encoded

obatirou:
All relevant end to end tests passes but there is problems with nonce and relayer cc @ClementWalter

tests/end_to_end/Security/test_dual_vm_token_hack.py
```diff
@@ -0,0 +1,166 @@
...
+
+@pytest_asyncio.fixture(
+    scope="package",
+    params=[
+        ("CairoPrecompiles", "DualVmToken"),
+        ("Security", "DualVmTokenWithoutModifier"),
+    ],
+    ids=["Modifier", "NoModifier"],
+)
+async def dual_vm_token(request, kakarot, starknet_token, owner):
```
enitrat, line 37: params on fixture makes every function that relies on `dual_vm_token` run twice, one with each version?

cairo_zero/kakarot/precompiles/precompiles_helpers.cairo
```diff
@@ -87,9 +87,18 @@ namespace PrecompilesHelpers {
...
         if (precompile_requires_whitelist == TRUE) {
-            let is_whitelisted = is_caller_whitelisted(caller_address);
+            // If the call is not a DELEGATECALL / CALLCODE, the actual caller is the caller_address
+            // Otherwise, the actual caller is the message_address.
+            if (is_not_delegatecall != FALSE) {
+                tempvar actual_caller_address = caller_address;
+            } else {
+                tempvar actual_caller_address = message_address;
+            }
+            let is_whitelisted = is_caller_whitelisted(actual_caller_address);
```
enitrat, line 101: can you explain the reasoning behind this? im not sure I get it

tests/end_to_end/Security/test_dual_vm_token_hack.py
obatirou: Yes we want to check that the fix on solidity is working (the nodelegateCallModifier) but also without it that the cairo fix is working.
No sure yet this is the right approach and might need to rethink how to test this.

cairo_zero/kakarot/precompiles/precompiles_helpers.cairo
obatirou: The idea of the cairo fix is to whitelist on the code that is calling the precompile. In this case the dualVmToken.
In case of a normal call: the address of the dualVMtoken is the caller_address (the address that called the address of execution)
`tempvar caller_address = evm.message.caller;`

In case of delegatecall, the latter is changed. The `evm.message.caller` is now the one that called the dualVmTokens that calls the precompiles.
And the address of the dualVMtokens is `evm.message.address.evm`

From `exec_precompile`
```
@param caller_address The address of the caller of the precompile. Delegatecall rules apply.
@param message_address The address being executed in the current message.
```
In fact delegate rules applies also for the `message_address`

solidity_contracts/src/CairoPrecompiles/DualVmToken.sol
```diff
@@ -1,16 +1,21 @@
...
+import {NoDelegateCall} from "../Security/NoDelegateCall.sol";
 
 /// @notice EVM adapter into a Cairo ERC20 token
 /// @dev This implementation is highly experimental
 ///      It relies on CairoLib to perform Cairo precompile calls
 ///      Events are emitted in this contract but also in the Starknet token contract
+/// @dev External functions are  to prevent a user making an EVM call to a malicious contract,
+/// with any calldata, that would be able to directly control on their behalf any quantity of any one of the ERC20
+/// tokens held by the victim's account contract, with the sole condition that the ERC20 has an
+/// authorized DualVmToken wrapper.
```
obatirou, line 15: I do not understand what you meant @enitrat, could you rephrase please ?

enitrat, line 15:
```suggestion
/// @dev External functions are `NoDelegateCall` to prevent a user making an EVM call to a malicious contract,
/// with any calldata, that would be able to directly control on their behalf any quantity of any one of the ERC20
/// tokens held by the victim's account contract, with the sole condition that the ERC20 has an
/// authorized DualVmToken wrapper.
```

enitrat: it was missing a word.

enitrat: it was missing a word

solidity_contracts/src/Security/DualVmTokenHack.sol
```diff
@@ -0,0 +1,80 @@
+// SPDX-License-Identifier: MIT
```
obatirou, line 1: The naming of the folder is confusing 
It should be changed

solidity_contracts/src/CairoPrecompiles/DualVmToken.sol
```diff
@@ -194,7 +200,7 @@ contract DualVmToken {
                                 CONSTRUCTOR
      //////////////////////////////////////////////////////////////*/
 
-    constructor(uint256 _kakarot, uint256 _starknetToken) {
+    constructor(uint256 _kakarot, uint256 _starknetToken) NoDelegateCall() {
```
ClementWalter, line 203:
```suggestion
    constructor(uint256 _kakarot, uint256 _starknetToken) {
```

ClementWalter (approved):
ok for me, I was a bit surprised in the beginning in the `tests/end_to_end/NoDelegateCall/test_dual_vm_token_hack.py` test that the test was failing with and without the modifier, but it's because we also did the modif in cairo to remediate the issue

PR #1579, opened 2024-11-06
[KGA-3] [KGA-122] fix: SSJ pow

enitrat:
Fixes `pow` in SSJ

- - -
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot/1579)

### DIFF
--- a/cairo/kakarot-ssj/crates/utils/src/math.cairo
+++ b/cairo/kakarot-ssj/crates/utils/src/math.cairo
@@ -38,6 +38,9 @@ impl ExponentiationImpl<
 > of Exponentiation<T> {
     fn pow(self: T, mut exponent: T) -> T {
         let zero = Zero::zero();
+        if exponent.is_zero() {
+            return One::one();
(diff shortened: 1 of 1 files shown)


cairo/kakarot-ssj/crates/utils/src/math.cairo
```diff
@@ -367,6 +370,7 @@ mod tests {
 
     #[test]
     fn test_wrapping_pow() {
+        assert(0_u256.wrapping_pow(0) == 1, '0^0 should be 1');
```
obatirou, line 373: Why not using the assert! macro ?

obatirou (approved):

cairo/kakarot-ssj/crates/utils/src/math.cairo
enitrat: No reason I did as the otherlines in the test. It could be moved to use assert ea

PR #1584, opened 2024-11-06
[KGA-70] fix: handle_l1_message gas price set to 0

obatirou:
Time spent on this PR:

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Resolves #<Issue number>

## What is the new behavior?

Fees are payed on L1 in the starknet messaging contract.
Users should not pay a second time.

- - -
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot/1584)

### DIFF
--- a/cairo_zero/kakarot/constants.cairo
+++ b/cairo_zero/kakarot/constants.cairo
@@ -39,7 +39,7 @@ namespace Constants {
     const FELT252_PRIME_LOW = 0x1;
 
     // Infinite gas handle_l1_message
-    const INFINITE_GAS = 2100000000;
+    const INFINITE_GAS = 2 ** 64 - 1;
 }
 
 // See model.Opcode:
--- a/cairo_zero/kakarot/constants.cairo
+++ b/cairo_zero/kakarot/constants.cairo
@@ -37,6 +37,9 @@ namespace Constants {
     // FIELD PRIME
     const FELT252_PRIME_HIGH = 0x8000000000000110000000000000000;
     const FELT252_PRIME_LOW = 0x1;
+
+    // Infinite gas handle_l1_message
+    const INFINITE_GAS = 2100000000;
 }
 
 // See model.Opcode:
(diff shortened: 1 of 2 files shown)


cairo_zero/kakarot/library.cairo
```diff
@@ -410,7 +410,7 @@ namespace Kakarot {
         let (access_list) = alloc();
 
         return eth_call(
-            0, l1_sender, to, 2100000000, 1, value_u256, data_len, data, 0, access_list
+            0, l1_sender, to, 2100000000, 0, value_u256, data_len, data, 0, access_list
```
ClementWalter, line 413: wondering if we should create a contant called `INFINITE_GAS` or something explicit meaning : we don't want this call to fail because oog, this tx is paid for in the messaging flow by starknet and should not revert. Price is 0 so no refund will occur

obatirou: I agree 
Done with [bcc4d01](https://github.com/kkrt-labs/kakarot/pull/1584/commits/bcc4d01268ef4f7a2e0ee58f74ed571700913a4d)

Added a docstring on the function to explain why those values are choosen

cairo_zero/kakarot/constants.cairo
```diff
@@ -37,6 +37,9 @@ namespace Constants {
     // FIELD PRIME
     const FELT252_PRIME_HIGH = 0x8000000000000110000000000000000;
     const FELT252_PRIME_LOW = 0x1;
+
+    // Infinite gas handle_l1_message
+    const INFINITE_GAS = 2100000000;
```
ClementWalter, line 42: let's try `P - 1`

enitrat: let's use `2**64 - 1` instead, as in most clients gas is a u64 field

obatirou: Done

ClementWalter (approved):

PR #1590, opened 2024-11-07
cw/fix tests e2e staging

ClementWalter:
- **Some initial fixes**
- **Update deployment folders, works on katana**
- **Fix DualVMToken for legacy sn ERC20**
- **Fix declaration for StarknetToken**
- **Remove useless fixtures**
- **Refacto relayers**

- - -
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot/1590)


deployments/staging/kakarot_deployments.json
```diff
@@ -0,0 +1,58 @@
+{
```
Eikix, line 1: are these the real tokens on sepolia?

kakarot_scripts/utils/starknet.py
```diff
@@ -387,18 +400,43 @@ async def deploy_starknet_account(class_hash=None, private_key=None, amount=1):
...
+            address=address,
+            class_hash=class_hash,
+            salt=salt,
+            key_pair=key_pair,
+            client=RPC_CLIENT,
+            constructor_calldata=constructor_calldata,
+            max_fee=_max_fee,
+        )
+    except Exception as e:
+        logger.info(f"❌ Error deploying account: {e}, waiting 5s before retrying")
```
Eikix, line 429: dumb question but why don't we logger.warn when something wrong happens in general?

Eikix (commented):
is the way one runs e2e tests affected? should we add something in the readme about deploying a relayer fleet with a specific private key in env?

deployments/staging/kakarot_deployments.json
ClementWalter: not with the same kakarot, so they are "invisible"

ClementWalter:
image
latest run on staging:
- all errors are for ercXX test, error in argent multisig api
- failures are related to updated tests not matching the current deployed kakarot, mutlsig REVERTED raising instead of returning status, state not updated. I'll just investigate the messaging to make sure, otherwise all green to me

.env.example
```diff
@@ -1,64 +1,48 @@
...
-STARKNET_SEPOLIA_PRIVATE_KEY=
+# Starknet accounts
+STARKNET_DEVNET_ACCOUNT_ADDRESS=
+STARKNET_DEVNET_PRIVATE_KEY=
+# This default value is Anvil first account private key
+STARKNET_DEVNET_EVM_PRIVATE_KEY=0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80
 
 KATANA_ACCOUNT_ADDRESS=0xb3ff441a68610b30fd5e2abbf3a1548eb6ba6f3559f2862bf2dc757e5828ca
 KATANA_PRIVATE_KEY=0x2bbf4f9fd0bbb2e60b0316c1fe0b76cf7a4d0198bd493ced9b8df2a3a24d68a
+# This default value is Anvil first account private key
```
enitrat, line 17:
```suggestion
# This default value is Katana first account private key
```

kakarot_scripts/deployment/dualvm_token_deployments.py
```diff
@@ -134,12 +137,13 @@ async def deploy_dualvm_tokens() -> None:
             "CairoPrecompiles", "DualVmToken", evm_deployments[token["name"]]["address"]
         )
         assert await token_contract.starknetToken() == l2_token_address
-        assert await token_contract.name() == token["name"]
-        assert await token_contract.symbol() == token["symbol"]
+        assert (await token_contract.name()).lstrip("\x00") == token["name"]
```
enitrat, line 140: create issue to strip zeroes for zeroes?

kakarot_scripts/utils/kakarot.py
```diff
@@ -55,10 +57,74 @@
...
+                _get_starknet_contract("account_contract", address=account.address)
+                .functions["get_nonce"]
+                .call(block_number="pending")
+            )
+        ).nonce
+
+    retries = 10
+    while network_nonce != _nonces[account.address] and retries > 0:
+        logger.info(
+            f"⏳ Waiting for network nonce {network_nonce} to be {_nonces[account.address]}"
```
enitrat, line 99: c'est pour solve quoi exactement ça ? attendre que les tx soient "minées"?

kakarot_scripts/utils/kakarot.py
```diff
@@ -451,7 +514,9 @@ async def _wrapper(self, *args, **kwargs):
                 payload["data"] = list(payload["data"])
                 payload["origin"] = int(payload["from"], 16)
                 del payload["from"]
-                result = await kakarot_contract.functions["eth_call"].call(**payload)
+                result = await kakarot_contract.functions["eth_call"].call(
+                    **payload, block_number="pending"
```
enitrat, line 518: should pending not be the default argument?

tests/end_to_end/CairoPrecompiles/test_call_cairo_precompile.py
```diff
@@ -105,17 +106,20 @@ async def test_should_fail_when_called_with_callcode(
             ):
                 await cairo_counter_caller.incrementCairoCounterCallcode()
 
+    @pytest.mark.skipif(
```
enitrat, line 109: q: why not xfail here? so that once done, this fails and we don't forget it

tests/end_to_end/L1L2Messaging/test_messaging.py
```diff
@@ -114,38 +154,28 @@ async def test_should_fail_unauthorized_message_sender(
...
+        current_l2_block = await RPC_CLIENT.get_block_number()
         message_app_l1.increaseL2AppCounterFromCounterPartOnly(
             message_app_l2.address, value=increment_value
         )
-        time.sleep(4)
+        await wait_for_l2_messaging(current_l2_block, message_app_l2.address)
         msg_counter_after = await message_app_l2.receivedMessagesCounter()
         assert msg_counter_after == msg_counter_before + increment_value
-
-    async def test_should_fail_unauthorized_message_sender(
```
enitrat, line 137: removing this one?

enitrat (commented):
end-to-end tests could be pulling the state from SN Sepolia when starting Katana so that we run everything like on a sepolia deployment?

.env.example
ClementWalter: no it's anvil first account private key (EVM account)

kakarot_scripts/deployment/dualvm_token_deployments.py
ClementWalter: ?

kakarot_scripts/utils/kakarot.py
ClementWalter: because the state got from the RPC node might not be up to date with the latest tx, I was facing a lot of nonce mismatch so I decided to take this conservative approach here: waiting for retrieved nonce to be the one expected, or wait for 1s

kakarot_scripts/utils/kakarot.py
ClementWalter: you mean keep latest as default?

tests/end_to_end/CairoPrecompiles/test_call_cairo_precompile.py
ClementWalter: skipif because we skip some test for some networks (e.g. if the network is not up to date, or some other conditions like for messaging)

tests/end_to_end/L1L2Messaging/test_messaging.py
ClementWalter: added as unit test instead to avoid updating kakarot real values in staging

kakarot_scripts/deployment/dualvm_token_deployments.py
enitrat: to strip zeroes for the name* sorry

or we don't care if the result is 0x00000000name?

kakarot_scripts/utils/kakarot.py
enitrat: no i meant I thought the default argument (so block_number=None) resolves to "pending" but this actually might not be the case.

enitrat (approved):

kakarot_scripts/deployment/dualvm_token_deployments.py
ClementWalter: you mean to remove this `lstrip` ? So an issue to update the sol files? I think that we don't care for now, do we?

kakarot_scripts/utils/kakarot.py
ClementWalter: actually this change looks useless since the default is

```
def _get_raw_block_identifier(
    block_hash: Optional[Union[Hash, Tag]] = None,
    block_number: Optional[Union[int, Tag]] = None,
) -> Union[dict, Hash, Tag, None]:
    if block_hash is not None and block_number is not None:
        raise ValueError(
            "Arguments block_hash and block_number are mutually exclusive."
        )

    if block_hash in ("latest", "pending") or block_number in ("latest", "pending"):
        return block_hash or block_number

    if block_hash is not None:
        return {"block_hash": _to_rpc_felt(block_hash)}

    if block_number is not None:
        return {"block_number": block_number}

    return "pending"
```

kakarot_scripts/deployment/dualvm_token_deployments.py
enitrat: i'm just worried something like STRK ends up showing like 000000000000000000STRK in the wallets. I actually don't know what it looks like until we can import it and see

PR #1606, opened 2024-11-13
[KGA-34] feat: update base fee starting from next block only

enitrat:
Updates the base fee mechanism.

In the EVM there can only be one base fee for a given block.
Thus, we use an index to manage two different base fees:
     - The base fee to use for the current block (index: 'current_block')
     - The base fee to use for the next block (index: 'next_block')
     
The startegy is:
- when getting the base fee, check if you can get the one from the 'next_block' (if block_number > next_fee_start_block_number)
- if yes, then update the value in `current_block` and return the new value
- otherwise, return the value from `current_block`     

When setting the base_fee, it will be applicable starting from `block_number + 1`.

Note: for tests we have to manually patch both `addr` and `addr+1` as we'restoring two values.

- - -
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot/1606)

### DIFF
--- a/cairo_zero/backend/starknet.cairo
+++ b/cairo_zero/backend/starknet.cairo
@@ -7,7 +7,7 @@ from starkware.cairo.common.bool import FALSE, TRUE
 from starkware.cairo.common.cairo_builtins import HashBuiltin
 from starkware.cairo.common.dict_access import DictAccess
 from starkware.cairo.common.uint256 import Uint256
-from starkware.cairo.common.math_cmp import is_nn
+from starkware.cairo.common.math_cmp import is_nn, is_le
 from starkware.cairo.common.memset import memset
 from starkware.starknet.common.syscalls import (
     emit_event,
@@ -23,6 +23,7 @@ from kakarot.precompiles.precompiles_helpers import PrecompilesHelpers
 from kakarot.constants import Constants
 from kakarot.interfaces.interfaces import IERC20, IAccount
 from kakarot.model import model
+from utils.utils import Helpers
 from kakarot.state import State
 from kakarot.storages import (
     Kakarot_native_token_address,
@@ -110,7 +111,7 @@ namespace Starknet {
         let (block_number) = get_block_number();
         let (block_timestamp) = get_block_timestamp();
         let (coinbase) = Kakarot_coinbase.read();
-        let (base_fee) = Kakarot_base_fee.read();
+        let (base_fee) = get_base_fee();
         let (block_gas_limit) = Kakarot_block_gas_limit.read();
         let (prev_randao) = Kakarot_prev_randao.read();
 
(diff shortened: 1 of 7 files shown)


cairo_zero/kakarot/library.cairo
```diff
@@ -160,21 +160,38 @@ namespace Kakarot {
...
+        let next_base_fee = res_next_block[0];
+        let starting_block = res_next_block[1];
+        Kakarot_base_fee.write('next_block', (base_fee, block_number + 1));
+
+        let is_next_fee_ready = is_le(starting_block, block_number);
+        if (is_next_fee_ready == FALSE) {
+            return ();
+        }
+
+        Kakarot_base_fee.write('current_block', (base_fee, starting_block));
```
ClementWalter, line 182:
```suggestion
```

you cannot set base fee directly, otherwise the whole PR is somehow pointless as you can submit a set_base_fee with prvious block numbers and have it performed directly.

cairo_zero/backend/starknet.cairo
```diff
@@ -128,6 +129,37 @@ namespace Starknet {
...
+        let is_already_updated = Helpers.is_zero(
+            current_start_block_number - next_start_block_number
+        );
+
+        if (is_already_updated == FALSE and is_next_fee_ready != FALSE) {
+            Kakarot_base_fee.write('current_block', (next_base_fee, next_start_block_number));
+            return (next_base_fee,);
+        }
+
+        return (current_base_fee,);
```
ClementWalter, line 161:
```suggestion
        let (next_block_base_fee) = Kakarot_base_fee.read('next_block');
        let next_base_fee = res_next_block[0];
        let next_start_block_number = res_next_block[1];
        let (block_number) = get_block_number();
        let is_next_fee_ready = is_le(next_start_block_number, block_number);
                
        if (next_start_block_number != 0 and is_next_fee_read != FALSE) {
            // update current_block storage and return next_block value
            Kakarot_base_fee.write('current_block', (next_base_fee, block_number));
            Kakarot_base_fee.write('next_block', (0, 0));
            return next_base_fee;
        }
        
        let (res_current_block) = Kakarot_base_fee.read('current_block');
        return res_current_block[0];
```

cairo_zero/kakarot/library.cairo
enitrat: > you can submit a set_base_fee with prvious block numbers

wdym? `set_base_fee` is always going to use `block_number + 1`. What this does is simply write the value of `next_block` in `current_block` if the value in `next_block` is already "active"; but no one has called `get_base_fee` in between.

cairo_zero/tests/src/kakarot/test_kakarot.py
```diff
@@ -142,11 +142,168 @@ def test_should_assert_only_owner(self, cairo_run):
...
+            ),
+            value=2,
+        )
+        @SyscallHandler.patch(
+            get_storage_var_address(
+                "Kakarot_base_fee", int.from_bytes(b"next_block", "big")
+            )
+            + 1,
+            value=0x101,
+        )
```
ClementWalter, line 237: I'd rather update the patch decorator to handle a list like

```
@patch("Kakarot_base_fee", "current_block", [1, 0x100])
```

and update here

```
cls.patches[selector_if_storage] = value
```

to something like

```
if isinstance(value, Iterable):
    for i, v in enumerate(value):
        cls.patches[selector_if_storage + i] = v
else:
    cls.patches[selector_if_storage] = value
```

cairo_zero/kakarot/library.cairo
ClementWalter: so maybe you wanted to do 

```
Kakarot_base_fee.write('current_block', (next_base_fee, starting_block));
```

but this just shouldn't be here. Why would you add in a `set_base_fee` an logic regarding `get_base_fee`?

enitrat: > Why would you add in a set_base_fee a logic regarding get_base_fee?

Because if you do

block=n-1: `current_base_fee=z`
block=n: `set_base_fee(x)`; `current_base_fee=z`
block=n+1: set_base_fee(y); `current_base_fee=z` (because no one has called `get_base_fee` yet

you would expect that at block=`n+1`, `current_base_fee=x`, but because you overwrote the `next_base_fee`, the value was lost and `current_base_fee` was not updated

cairo_zero/tests/src/kakarot/test_kakarot.py
enitrat: yeah agreed will get on that

ClementWalter (approved):

PR #1616, opened 2024-11-18
[KGA-26] [KGA-17] feat: add checks on Starknet return values for DualVMToken

enitrat:
https://github.com/code-423n4/2024-09-kakarot-findings/issues/51
https://github.com/code-423n4/2024-09-kakarot-findings/issues/42

Checks the return value for DualVMToken. If the SN call returned false, panicking the EVM call will make the starknet tx revert as a side effect.


kakarot_scripts/compile_kakarot.py
```diff
@@ -22,8 +31,27 @@ def main():
     # %% Compile
     logger.info(f"ℹ️  Compiling contracts for network {NETWORK['name']}")
     initial_time = datetime.now()
+
+    # Split contracts into Cairo 0 and Cairo 1 to avoid
+    # re-compiling the same package multiple times.
+    cairo0_contracts = []
+    cairo1_packages = set()
```
obatirou, line 38: Why a list for one and a set for the other ?

enitrat: you compile cairo0 contracts individually but you compile cairo1 contracts per package. Thus you only want packages to be represented once, even if a package contains multiple contracts.

obatirou (approved):

solidity_contracts/src/CairoPrecompiles/DualVmToken.sol
```diff
@@ -285,7 +286,8 @@ contract DualVmToken is NoDelegateCall {
         transferCallData[1] = uint256(amountLow);
         transferCallData[2] = uint256(amountHigh);
 
-        starknetToken.delegatecallCairo("transfer", transferCallData);
+        bool success = abi.decode(starknetToken.delegatecallCairo("transfer", transferCallData), (bool));
+        require(success, "Transfer failed");
```
obatirou, line 290: use solidity errors instead of require statement to keep the same way of handling errors

obatirou (approved):

ClementWalter (approved):

PR #1620, opened 2024-11-19
fix: fixed foundry version to a working one

enitrat:
Latest foundry nightly is bugged with WETH9. Fix the version to the latest working one.

### DIFF
--- a/.github/workflows/cairo-zero-ci.yml
+++ b/.github/workflows/cairo-zero-ci.yml
@@ -95,7 +95,7 @@ jobs:
       - name: Install Foundry
         uses: foundry-rs/foundry-toolchain@v1
         with:
-          version: nightly
+          version: nightly-a79dfaed6fc6f88cda5f314a25d1b484d9d8c051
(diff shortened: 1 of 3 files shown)


.github/workflows/cairo-zero-ci.yml
```diff
@@ -95,7 +95,7 @@ jobs:
       - name: Install Foundry
         uses: foundry-rs/foundry-toolchain@v1
         with:
-          version: nightly
+          version: nightly-a79dfaed6fc6f88cda5f314a25d1b484d9d8c051
```
obatirou, line 98: This is only a temporary fix 
Nightly are erased when new ones are pushed

enitrat: we'll update back when that will be fixed

obatirou (approved):

obatirou:
As said in comments 
This is only a temporary fix waiting on https://github.com/foundry-rs/foundry/issues/9349 to be resolved

PR #1644, opened 2024-11-27
fix: do not store code hash for EOAs

obatirou:
Time spent on this PR:

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Resolves #<Issue number>

## What is the new behavior?

Code hash for eoa cannot be stored as they can change. Put the logic of 0 or empty code hash in function.

### DIFF
--- a/cairo_zero/kakarot/account.cairo
+++ b/cairo_zero/kakarot/account.cairo
@@ -123,10 +123,7 @@ namespace Account {
             tempvar address = new model.Address(starknet=starknet_address, evm=evm_address);
             let balance = fetch_balance(address);
             assert balance_ptr = new Uint256(balance.low, balance.high);
-            // empty code hash see https://eips.ethereum.org/EIPS/eip-1052
-            tempvar code_hash_ptr = new Uint256(
-                low=Constants.EMPTY_CODE_HASH_LOW, high=Constants.EMPTY_CODE_HASH_HIGH
-            );
+            tempvar code_hash_ptr = new Uint256(0, 0);
             let account = Account.init(
                 address=address,
                 code_len=0,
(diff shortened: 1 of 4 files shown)


cairo_zero/kakarot/account.cairo
```diff
@@ -123,10 +123,7 @@ namespace Account {
             tempvar address = new model.Address(starknet=starknet_address, evm=evm_address);
             let balance = fetch_balance(address);
             assert balance_ptr = new Uint256(balance.low, balance.high);
-            // empty code hash see https://eips.ethereum.org/EIPS/eip-1052
-            tempvar code_hash_ptr = new Uint256(
-                low=Constants.EMPTY_CODE_HASH_LOW, high=Constants.EMPTY_CODE_HASH_HIGH
-            );
+            tempvar code_hash_ptr = new Uint256(0, 0);
```
ClementWalter, line 126:
```suggestion
```

ClementWalter (changes requested):
fix empty pointer

ClementWalter (approved):

enitrat (commented):
TLDR;
1. All accounts have a codehash. If that account is an eoa it's simply keccak256("");
2. You never want to have a disrepancy between the account's code and it's hash, thus it makes sense to write the code_hash in the same function as the account.

I think there's a small misconception between the `code_hash` of an account and  the `extcodehash` in the way we handled this finding:

- All accounts have a code hash that is `keccak256(bytecode)`. If bytecode is empty, that code hash is `keccak256("")` = EMPTY_CODE_HASH
- The finding is valid when saying that account deployed with `deploy_externally_owned_account` would have the wrong codehash

However I don't really agree with the way we handled this; I would rather have:
1. Modified `Starknet.deploy` to `set_code_hash(EMPTY_CODE_HASH)` (you're deploying a contract that doesn't have any code)
2. Replaced the `write_bytecode` of the account contract to write the code hash as well.

